PR #74, opened 2026-06-09
Add managed automation seeds

rocketman-21:
## Summary
- Add `MURPH_MANAGED_AUTOMATIONS`, a Murph-owned scheduled-automation seed list in `assistant-engine` (first seed: weekly health digest), persisted through the existing canonical automation APIs — no new scheduler, backfill script, or durable state.
- Add `applyMurphManagedAutomations`, an idempotent apply function: creates a missing seed only when a deliverable route resolves, updates only when seed content actually changed, and skips inactive automations, stale one-shots, and slugs already owned by a user automation.
- Hook managed-automation maintenance into the hosted runtime workspace-assistant phase as best-effort optional work, reporting a checkpoint only when records change and skipping entirely on fresh foreground input so it never blocks the user's turn.

## Review fixes folded in
- Seed uses a valid canonical automation id shape instead of a bespoke id.
- Explicit id-then-slug existence checks avoid relying on `upsertAutomation` slug-collision normalization (a user-owned slug is left alone).
- Create-route resolution is lazy — only when a seed may actually be created.

## Testing
- `pnpm --filter @murphai/assistant-engine test` — 1177 passed (includes new mocked unit suite + real-core registry-write regression tests)
- `pnpm --filter @murphai/assistant-runtime test` — 799 passed (includes foreground fresh-input skip assertion)
- Typecheck green for both packages; full diff-aware verification (assistant-engine, assistant-runtime, CLI, setup-cli, Cloudflare verify) passed pre-rebase, and focused suites re-ran green after fast-forwarding onto current `main`.

---
<a href="https://app.blacksmith.sh/cobuildwithus/codesmith/murph/pr/74"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783624258&installation_id=127572939&pr_number=74&repository=cobuildwithus%2Fmurph&return_to=https%3A%2F%2Fgithub.com%2Fcobuildwithus%2Fmurph%2Fpull%2F74&signature=d7fb1a5a5950436cd7b877fee94a85ea530ceb25a130a16bfbccc578699a01bc"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


rocketman-21:
Follow-up commit (53638ab83) from a thorough review pass — net **-38 lines**:

**Fixed (latent bug):** `murphManagedAutomationSeedChanged` compared `existing.summary` against `null` for a summary-less seed, while `upsertAutomation` preserves the existing summary when the field is omitted — any future seed without a summary would have flagged "changed" and rewritten the vault record on every background wake, forever. The comparison now treats a null seed summary as unmanaged, matching the write semantics.

**Deleted (speculative generality):**
- The one-shot `'at'`-schedule staleness machinery (`isStaleMurphManagedOneShotSeed` + 30-min expiry constant) — no production seed uses it; `MurphManagedAutomationSchedule` now excludes `'at'` so it can't reappear untested.
- The test-only `seeds?` injection parameter on `ApplyMurphManagedAutomationsInput`.

**Hardened:** new real-registry apply-twice idempotency test proving the persisted record round-trips (markdown body, trimming, tag dedup) without triggering a rewrite — the churn class above is now caught by an actual persistence test, not mocks.

**Documented:** update path intentionally preserves the user/runtime-owned route without re-validating deliverability (only the create path chooses a route).

Verified: assistant-engine suite 1177 passed, assistant-engine + assistant-runtime typechecks green.

rocketman-21:
364b00c10 restores one-shot ('at') seed support — the product need (one-time feature-drop notifications) is now concrete. Back in: the `at` schedule allowance, the 30-minute staleness guard (now with a comment explaining the late-wake hazard it prevents), and the `seeds` injection input, which is required to test the time-dependent staleness path. The earlier review fixes (summary-churn guard, real-registry idempotency test) are unchanged.

PR #127, opened 2026-06-11
Make codex app-server client thread-aware and bill subagent thread usage

rocketman-21:
## Why

Codex app-server auto-attaches every connection to every thread it creates — including future `spawn_agent` subagent threads (`codex-rs/app-server/src/lib.rs:1024`). Murph's codex client correlated every incoming event against a single turn id and ignored `thread_id` entirely, so:

1. The first child-thread event would fail the parent turn with `ASSISTANT_CODEX_APP_SERVER_STALE_TURN_EVENT` (retryable → retry storm).
2. Child-thread token usage would never land in `hosted_ai_usage` — invisible spend.

Codex ships no aggregate usage primitive (no usage RPC, no usage on the protocol `Turn`, no parent-side aggregation, collab results carry no usage), so consuming per-thread `thread/tokenUsage/updated` notifications is the canonical upstream pattern.

This lands the **primitives only**: codex collab/multi-agent tools remain OFF. Enabling them later is config + prompt guidance, with usage observability already in place.

## What

- **Thread-aware routing** (`assistant-codex.ts`): events whose thread id differs from the parent's are routed off the single-turn correlation path. Child tokenUsage events are buffered per thread (first/last sample, 32-thread cap with per-thread dropped count); child server requests get a JSON-RPC error response instead of failing the turn; other child events are dropped. Parent-thread strictness is unchanged by construction.
- **Per-child usage drafts** (`providers/helpers.ts`): `extractCodexSubagentUsageDrafts` reuses the existing total-delta arithmetic and flows through the existing `additionalUsages` → `recordAdditionalAssistantUsageEvents` seam (`providerRequestOrdinal ≥ 1`, shared counter with dynamic-tool drafts). No schema change; per-row `requested_model`/`served_model` carry the child's effective model from parent-thread collab spawn items, so multi-model billing (e.g. mini subagents) works day one.
- **Spawn-evidence billing gate** (review finding): warm processes are reused across threads, so a foreign thread id alone is not proof of a subagent — only threads named in a spawn item's `receiverThreadIds` produce usage rows. A stale flush from a previous thread can never mint a row; unattributed threads are counted in draft `rawUsageJson`.
- Drafts appear in **both** the success result and the turn-failure context.
- `rawUsageJson` stays metadata-only (thread id, token delta, counts) — spawn `prompt` provably never leaks (asserted in tests).

## Verification

- `pnpm --dir packages/assistant-engine typecheck` — clean
- `pnpm test:diff packages/assistant-engine` — green (incl. apps/cloudflare verify)
- 4 touched test files: 168 tests passed — unit (drafts builder, spawn-evidence gate, prompt non-leak, snake_case tolerance, thread-id extractor) + integration on the fake-process harness (tolerate+record multi-model, deny child server request, strictness regression, failure-context drafts, 33-thread cap, ordinal continuity after `generate_image`)

## Audits

- `coverage-write`: added failure-context, cap-overflow, ordinal-uniqueness, prompt-non-leak, extractor-shape coverage (all green).
- `task-finish-review`: 1 medium (spurious billing from warm-process stale events — **fixed** via spawn-evidence gate + tests), 2 low (dropped-count semantics — **fixed** with per-thread set; plan/ledger drift — **fixed**).

Plan: `agent-docs/exec-plans/completed/2026-06-10-codex-thread-aware-usage.md`

Note for mergers: a separate active lane ("Turn-profile command label quoted-shell unwrap") touches other functions in `helpers.ts`; diffs are in different functions but expect adjacent-line merges.

---
<a href="https://app.blacksmith.sh/cobuildwithus/codesmith/murph/pr/127"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783743144&installation_id=127572939&pr_number=127&repository=cobuildwithus%2Fmurph&return_to=https%3A%2F%2Fgithub.com%2Fcobuildwithus%2Fmurph%2Fpull%2F127&signature=7f1ecdcaf32c463fb09bc194607562291050a9bec5029879a9f1a6034b47adae"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/agent-docs/exec-plans/completed/2026-06-10-codex-thread-aware-usage.md
+++ b/agent-docs/exec-plans/completed/2026-06-10-codex-thread-aware-usage.md
@@ -1,0 +1,211 @@
+# Codex thread-aware event correlation and subagent usage accounting
+
+## Goal
+
+Make Murph's codex app-server client thread-aware so that events from codex
+subagent threads (native `spawn_agent` children) neither fail the parent turn
+nor escape usage accounting. This lands the two primitives required before the
+codex collab/multi-agent feature can ever be enabled; it does NOT enable that
+feature.
+
+Success criteria:
+
+- An event stream containing interleaved child-thread events (own
+  `thread_id`/`turn_id`) completes the parent turn successfully; today the
+  first such event rejects the turn with
+  `ASSISTANT_CODEX_APP_SERVER_STALE_TURN_EVENT`.
+- Per-child-thread token usage is recorded into `hosted_ai_usage` as
+  additional usage rows on the parent turn (existing
+  `providerRequestOrdinal >= 1` seam), with child thread id carried in the
+  row's `rawUsageJson` (decision: `gateway_tags_json` is populated from
+  per-turn usage attribution, not per-draft; `rawUsageJson` is the existing
+  per-draft JSON column and needs no recording-path change) and the child
+  model attributed from parent-thread spawn items.
+- Parent-thread correlation strictness is byte-for-byte unchanged: events on
+  the parent thread with a mismatched/missing turn id still reject exactly as
+  before.
+- No codex feature flags flipped; no behavior change for today's
+  single-thread production turns other than tolerating foreign-thread events.
+
+## Why (evidence)
+
+Verified 2026-06-10 against the `../codex` checkout (2026-05-29, `27e256bc40`):
+
+- App-server auto-attaches every initialized connection as a listener for
+  every newly created thread, including spawned subagent threads
+  (`codex-rs/app-server/src/lib.rs:1024`, `in_process.rs:489`;
+  `thread_processor.rs:2378` has no source filtering). Child events WILL
+  arrive on Murph's connection.
+- Token usage notifications are thread-scoped and carry `thread_id` +
+  `turn_id` (`bespoke_event_handling.rs:1574`). Codex ships no aggregate
+  usage primitive: no usage RPC, no usage field on the protocol `Turn`
+  (`app-server-protocol/src/protocol/v2/thread_data.rs:153`), no parent-side
+  aggregation of child usage, and collab wait/close results carry no usage.
+  Consuming per-thread `thread/tokenUsage/updated` events is the canonical
+  upstream pattern.
+- Murph's client correlates every incoming event against a single
+  `expectedTurnId` and ignores `thread_id` entirely
+  (`packages/assistant-engine/src/assistant-codex.ts` —
+  `validateWarmTurnEventCorrelation`, `bindExpectedTurnId`,
+  `codexEventMethodRequiresTurnCorrelation`); mismatches call
+  `rejectOnce(...)`, failing the turn. The "one thread per connection"
+  assumption was never codex's contract.
+- Usage extraction resolves tokenUsage events for the parent turn id only
+  (`packages/assistant-engine/src/assistant/providers/helpers.ts` —
+  `extractCodexAssistantProviderUsage`,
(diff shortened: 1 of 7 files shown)


rocketman-21:
Addressed all review findings in 9f5ab11:

**Finding 1 (late child events after parent completion) — fixed.** The process now remembers its bound thread id (`noteBoundThreadId`, set at turn release). While idle, foreign-thread notifications are tolerated (never billed) and foreign-thread server requests get a `-32000` deny so the child doesn't hang; idle output for the *bound* thread still poisons, so the contamination guard is unchanged. Policy made explicit in code comments + test: `tolerates subagent thread notifications between turns without poisoning the warm process` (late tokenUsage + late server request, then warm reuse on the same process). The pre-existing `stops warm Codex when output arrives outside an active turn` test still passes — its idle event carries the bound thread id.

**Finding 2 (cap consumed by stale threads) — fixed.** The turn now tracks collab receiver thread ids live; at a full buffer, an evidenced child evicts an unattributed sample (which is never billable anyway) instead of being dropped. Test: `evicts unattributed foreign threads from a full buffer in favor of spawned children` (32 ghosts → spawned child still bills; dropped=1, unattributed=31 surfaced in rawUsageJson).

**Finding 3 (spawn-only attribution) — broadened.** Billing evidence is now ANY parent-thread `collabAgentToolCall` naming `receiverThreadIds` (spawnAgent/sendInput/wait/resume), so reused children bill too; model attribution remains spawn-only (the only item that carries it). Scope limit documented: between-turn child usage is tolerated but not billed — the supported delegation pattern is spawn+wait within a turn.

**PR scope** — resolved by rebase: the companion-app doc commits were local-main commits that leaked into the branch base; they've since landed on origin/main independently, and the branch now contains only the code commit. This also fixed `mergeable: false` (conflict with the turn-profile helpers.ts fix on main).

Verification on the new head: assistant-engine typecheck clean; 172 tests across the 4 touched files; `pnpm test:diff packages/assistant-engine` green incl. apps/cloudflare verify. CI running on 9f5ab11.

rocketman-21:
Final review patch in e129e58:

**Pre-thread/start race — fixed.** The routing gate now falls back to the process's `lastBoundThreadId` while `codexThreadId` is still null (fresh thread/start in flight on a warm process), so a late child event in that window is routed instead of failing correlation. Events for the last bound thread itself keep strict stale handling. New regression test: `routes late child events arriving before the next thread/start response resolves` (child tokenUsage lands between the thread/start request and its response; turn 2 completes on the same warm process; the unevidenced child is tolerated, not billed).

**Stale comment — fixed.** The drafts-builder header now describes the broadened collab-receiver evidence gate (spawn/sendInput/wait/resume), model attribution spawn-only.

Product contract noted and agreed: supported delegation is **spawn + wait within one parent turn** — fire-and-forget subagents are not fully billed by this primitive, and the enablement task must carry that constraint into the prompt guidance.

Verification: typecheck clean, 173 tests across touched files, `pnpm test:diff packages/assistant-engine` green incl. apps/cloudflare verify.

PR #128, opened 2026-06-11
feat(cloudflare): add hosted runner container CPU watchdog diagnostics

rocketman-21:
## Why

Production container metrics show recurring unattributed CPU pinning at ~100% of 1 vCPU on hosted runner containers — both seconds after boot and mid-lifetime on warm containers (one observed 12-minute sustained burn between jobs, instance `c224...`, Jun 11 01:54–02:06 UTC). Existing diagnostics (the per-job heartbeat) sample only memory and only during active jobs, so these burns are unattributable today.

## What

- **`apps/cloudflare/src/container-cpu-watchdog.ts`** (new, self-contained): every 20s samples `/sys/fs/cgroup/cpu.stat` + `/proc/<pid>/stat`; when interval usage ≥ 0.5 cores, emits one structured log (`lifecycleStage: "entrypoint-cpu-watchdog"`) with cgroup usage/throttle deltas, the per-pid attributed total (`attributedCpuCores` — exposes short-lived-process churn as an explicit gap), and the top-3 processes by CPU.
- **Comm redaction**: comm names pass an infrastructure allowlist; anything else (script basenames, PR_SET_NAME) logs as a stable `other:<sha256-prefix>` label, so process-controlled text never reaches persisted Workers Logs.
- **Boot visibility**: the baseline seeds immediately at start, so the first reportable interval covers the boot window — half the production mystery.
- **Liveness**: a one-time `entrypoint-cpu-watchdog-started` log reports `cgroupCpuStatAvailable` + `sampledProcessCount`, disambiguating "no burns" from "sampler broken" and answering whether the cgroup fallback is live fleet-wide.
- **Wiring** (`container-entrypoint.ts`, CLI path only): started in `startHostedContainerEntrypointCli` on the default process API — entrypoint unit tests inject sequenced `processApi` mocks the watchdog must not consume. Stopped on server close; interval is unref'd; every failure path is swallowed (diagnostics-only).

## Verification

- `pnpm test:diff` → full `apps/cloudflare verify` (typecheck + Node + Workers lanes): **84 files / 1,323 tests, green** (run after every revision)
- New focused suite: 11 tests (attribution + redaction both branches, threshold silence, cgroup fallback, started signal, pid-reuse clamp, unreadable `/proc`, tick re-entrancy, wall-clock guard, emit-throw resilience, unref)
- Known proof gap: not yet observed against real Linux `/proc`+cgroup (dev is macOS; fixtures encode the stat(5)/cgroup-v2 layouts). The started-signal log closes this within one interval of first deploy.

## Audits (repo completion workflow)

- `security-privacy-review`: 1 MEDIUM (raw process-controlled comm → persisted logs) — fixed via allowlist+digest; scoped re-review: **resolved, no new findings**
- `simplify`: removed unused config options; module boundary + structural ProcessApi subset (no `kill` capability) confirmed; cgroup fallback consciously kept (~8 lines vs silent blindness)
- `coverage-write`: +6 boundary tests, production code untouched
- `task-finish-review`: boot-window blind spot + liveness gap + attribution-gap findings accepted and fixed; sanitizer-seam test rejected (boundary proven by inspection; sanitizer preserves the payload shape, nesting depth 3 ≤ 4, valid keys)

---
<a href="https://app.blacksmith.sh/cobuildwithus/codesmith/murph/pr/128"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783745979&installation_id=127572939&pr_number=128&repository=cobuildwithus%2Fmurph&return_to=https%3A%2F%2Fgithub.com%2Fcobuildwithus%2Fmurph%2Fpull%2F128&signature=033454d374afe7be18849982e0d3d0f154d385c0b9069931e53eff62dfde5361"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/cloudflare/src/container-cpu-watchdog.ts
+++ b/apps/cloudflare/src/container-cpu-watchdog.ts
@@ -1,4 +1,4 @@
-import { createHash } from "node:crypto";
+import { createHmac } from "node:crypto";
 
 import { emitHostedExecutionStructuredLog } from "@murphai/hosted-execution";
 
@@ -47,12 +47,21 @@ export interface HostedContainerCpuWatchdogProcessApi {
 interface HostedContainerCpuWatchdogProcessSample {
   comm: string;
   cpuTicks: number;
+  // stat field 22 kept as an opaque equality token: same pid + same start time
+  // is the same process; a changed start time is a reused pid.
+  startTimeTicks: string | null;
 }
 
 interface HostedContainerCpuWatchdogSample {
   cgroupUsageUsec: number | null;
   nrThrottled: number | null;
   perPid: Map<number, HostedContainerCpuWatchdogProcessSample>;
+  // False when /proc enumeration failed or a live process could not be read,
+  // meaning a process may be missing from perPid. Attribution for pids absent
+  // from an incomplete sample is unknowable and must be suppressed, or a
+  // long-running process missed once would have its whole lifetime counted as
+  // interval CPU on the next tick.
+  procScanComplete: boolean;
   sampledAtMs: number;
   throttledUsec: number | null;
 }
@@ -66,18 +75,25 @@ interface HostedContainerCpuWatchdogSample {
 export function startHostedContainerCpuWatchdog(input: {
   processApi: HostedContainerCpuWatchdogProcessApi;
 }): () => void {
+  const fingerprintSecret = resolveCpuWatchdogFingerprintSecret();
   let previous: HostedContainerCpuWatchdogSample | null = null;
   let sampling = false;
+  let stopped = false;
   const tick = async () => {
     // Skip overlapping ticks if /proc reads stall instead of queueing them.
-    if (sampling) {
+    if (stopped || sampling) {
       return;
     }
     sampling = true;
     try {
       const current = await readHostedContainerCpuWatchdogSample(input.processApi);
+      // A sample already in flight when the server closed must not emit into
+      // a shutting-down process or advance state.
+      if (stopped) {
+        return;
+      }
       if (previous) {
-        emitHostedContainerCpuWatchdogReport({ current, previous });
+        emitHostedContainerCpuWatchdogReport({ current, fingerprintSecret, previous });
       } else {
         // One-time liveness signal: steady-state watchdog silence is otherwise
         // ambiguous between "no burns" and "sampler broken on this platform",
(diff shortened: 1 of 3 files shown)


rocketman-21:
All four review findings addressed in dc60bf1:

**1. False per-pid attribution (medium)** — implemented the starttime variant: each process sample carries stat field 22 as an opaque equality token, and each scan carries `procScanComplete` (readdir success + no non-ENOENT/ESRCH pid read failures + no parse failures). Attribution rule: same pid+starttime → counter delta; changed starttime → pid reuse, full ticks; absent from a **complete** previous scan → new process, full ticks; absent from an **incomplete** previous scan → unattributed. Fallback `cpuCoresUsed` inherits the rule, so an incomplete scan can only under-report, never false-alarm. Two regression tests pin the reviewer scenario (30k lifetime ticks ≠ 15 interval cores): cgroup-available (emits cgroup truth, `attributedCpuCores: 0`, empty top list) and fallback (full silence).

**2. Boot coverage** — watchdog now starts before `await startHostedContainerEntrypoint(...)` with try/catch stop on startup failure, exactly as suggested. Remaining gap is module evaluation only.

**3. Stop in-flight** — `stopped` flag checked at tick entry and re-checked after the sample read; in-flight samples can't emit (including the started signal) or advance state after stop. Tested via stalled-read-then-stop.

**4. Comm labels** — keyed HMAC-SHA256 (16 hex) using `HOSTED_LOG_FINGERPRINT_SECRET` ?? `HOSTED_AI_USAGE_REPORTING_SECRET` (same resolution order as the hosted context diagnostics fingerprints); without the secret, unknown comms collapse to a plain `other` bucket — no comm-derived identifier in either mode. The existing fingerprint helper lives in `assistant-engine`, so the watchdog reads the same env contract directly rather than inverting its dependency on `hosted-execution` only.

Verification: focused suite 16/16 (5 new tests), entrypoint suite 46/46, `pnpm test:diff` → full `apps/cloudflare verify` 84 files / 1,328 tests green.

PR #129, opened 2026-06-11
Probe runner liveness before clearing the write fence on transport failure

rocketman-21:
## Problem

Prod incident 2026-06-11 ~02:47–02:51 UTC: the UserRunner DO's transport call for a healthy, running accepted invocation failed (`runner.accepted_attempt_failed`, ~6×/day baseline) and `invokePreparedWithFence` treated that as invocation death, clearing the runtime write fence. The still-running container invocation became an orphan: it held the exclusive runner slot, no wake could route to it, and its idle-shutdown snapshot 401'd stale-authority. A user message arriving in that window waited ~96s for the orphan's 3-minute idle timer before a cold start could reply (2m03s total vs the normal 2–8s warm). This also violated the documented invariant that mailbox processing must not wait behind container lifecycle locks.

## Fix

One decision point changes: **verify before declaring death.**

- On invoke transport failure, probe the RunnerContainer DO (`readActiveRuntimeUserFence`, extended to return `attemptId`/`leaseGeneration`) for the exact fence identity, bounded by a 5s timeout. Still-active match → keep the fence (wakes keep routing to the live invocation; the existing not-wakeable → start-required replacement path reconciles after it exits). Missing / mismatched / probe error / timeout → clear exactly as before (fail toward today's behavior).
- Probe-first ordering is required: the committed-progress path would otherwise complete-and-clear a live invocation's fence.
- Observability: the persisted `runner.accepted_attempt_failed` row now carries `attemptId`, `leaseGeneration`, and metadata-only `redactedJson` (sanitized error fields under the parser-allowlisted `safeErrorMessage` key, `errorCodeDetail` token-constrained, plus `attemptStillActive`/`fenceCleared` flags). Probe error/timeout emits a metadata-only warn log.
- Durable doc: `agent-docs/references/hosted-runtime-protocol.md` write-fence and accepted-failure sections updated.

No new timers, no preemption machinery, no web changes — the web log route already passes the new fields through generically.

## Completion audits (all resolved)

- **security-privacy-review**: 1 medium fixed — bare `errorMessage` redactedJson key is rejected by the web parser's key gate, which would have silently dropped the row and the existing `runtime_recheck_requested` liveness signal; now `safeErrorMessage`. Fence reclaimability, confused-deputy, fail-closed, and exposure checks all verified clean.
- **coverage-write**: added committed-progress ordering, non-accepted-path, userId-mismatch, missing-probe-method, and redacted-json unit tests; hardened the harness to round-trip every captured log body through the real `parseHostedRuntimeLogRequest`.
- **deep-review** (feynman): probe timeout (F2), generation-field comparison (F3), unconfirmed-probe warn log (F6) accepted + implemented. F1 (probe can't see liveness across a RunnerContainer DO restart — in-memory active-op record) accepted as documented residual: incident evidence (96s of startup-confirmation attempts blocked on the lifecycle lock held by the still-running DO-side invoke) indicates the RPC-layer mode this change fixes. F4 (recheck-cooldown sharing) rejected as bounded-delay web-scope follow-up.
- **task-finish-review**: 1 medium fixed (`errorCodeDetail` value gating); timer hygiene, keep-fence rethrow compatibility, and deploy-skew behavior (old container DO → fence clears = today's behavior) verified clean.

## Verification

- `pnpm test:diff` (full `apps/cloudflare verify`: typecheck + Node + Workers lanes) green ×3 during development
- Post-rebase `pnpm --dir apps/cloudflare verify` green on current `origin/main`
- Focused: 59/59 in `runtime-invocation-transport-failure.test.ts` (11 cases) + `user-runner-alarm.test.ts`
- Direct scenario gap (human/post-deploy): the incident mode has no automated end-to-end reproduction; watch for `attemptStillActive: true` rows in `hosted_runtime_log` after deploy to confirm kept-fence behavior in real infra.

DEPLOYMENT CONCERNS: None blocking. Cloudflare-only change; `apps/web` and shared packages untouched, and the enriched log rows were round-trip-verified against the shared parser version web already runs. Deploy-skew is fail-safe in both directions (old RunnerContainer DO → probe mismatch → today's clear behavior). Post-deploy check: confirm `runner.accepted_attempt_failed` rows keep appearing in `hosted_runtime_log` (now with `redacted_json` populated) and that no 4xx spike shows on the `/internal/hosted-runtime/log` route.

---
<a href="https://app.blacksmith.sh/cobuildwithus/codesmith/murph/pr/129"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783779528&installation_id=127572939&pr_number=129&repository=cobuildwithus%2Fmurph&return_to=https%3A%2F%2Fgithub.com%2Fcobuildwithus%2Fmurph%2Fpull%2F129&signature=fc38eb19b68d38947abaa90ad255eb220719d83bc82402dd6083b48598f71d72"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


rocketman-21:
Addressed the review in 52839d889. Point-by-point:

**Accepted:**
- **Generation single source of truth (note 3)**: the probe now compares `active.leaseGeneration === prepared.job.request.leaseGeneration` — the exact value the container records — instead of the token alias.
- **Probe outcome diagnostics (note 4, minimal form)**: the probe returns an outcome enum (`active | inactive | mismatch | error | timeout`) and the persisted `runner.accepted_attempt_failed` row carries `attemptLivenessProbeOutcome` alongside the existing flags. One enum rather than per-field matched booleans — `inactive` vs `mismatch` is the distinction that matters for watching the documented DO-restart residual in prod. No raw IDs; round-trip-tested through the real parser.
- **Generation-only mismatch test** (same attempt, different lease generation → clears, row says `mismatch`).
- **`errorCodeDetail` gating tests** (path-shaped `.code` dropped; plain token like `ECONNRESET` kept).
- **Contract documentation (issue 1/2, the honest-naming part)**: `WorkerActiveRuntimeUserFenceResult` and the probe now document explicitly that "active" means the DO-side operation is in flight — including the pre-dispatch readiness window — and that it does not prove the runner child accepted work.

**Pushed back, with reasoning:**
- **Stage-gating the keep (issue 1's recommended fix)**: keeping the fence during the pre-dispatch window is not an overclaim bug — it's required for correctness. If transport fails while the DO-side invoke is in readiness, that invoke keeps running autonomously and will dispatch and run the invocation under the intact fence; clearing there orphans a soon-to-be-live invocation (the exact incident this PR fixes). If the DO invoke instead dies in readiness, its `finally` clears the active op and stops the warm container, so the kept fence goes stale with no slot held and the existing start-required/replacement path reclaims it within the grace window + recheck cadence. Both sub-cases are safe with "keep"; one sub-case is broken with stage-gated clear. Documented in the probe's contract comment instead.
- **Splitting `readActiveRuntimeUserFence` (issue 2)**: the egress consumer legitimately needs only user binding (attempt identity isn't available or meaningful on that fallback path), and two methods returning projections of the same in-memory record is API surface with no behavior change. Documented the dual-consumer contract on the type instead. Happy to split if a third consumer shows up with different needs.
- **Real RunnerContainer pre-dispatch lifecycle test**: this would test pre-existing `invokeHostedExecution` assignment timing rather than this diff's logic, and needs a substantial container-DO harness; the probe's decision table is fully covered at its boundary, and the keep-during-readiness rationale above is a property of the existing lifecycle code, not new behavior.

Verification after changes: apps/cloudflare verify green (typecheck + Node + Workers lanes, 1326 tests), focused suite 62/62.

PR #130, opened 2026-06-11
feat(hosted): compact oversized threads at idle stop, off the reply hot path

rocketman-21:
## Why

OpenAI Standard-tier prompt caches evict within ~45 minutes regardless of prefix size or `prompt_cache_retention` (measured exhaustively 2026-06-10; support confirmed large-entry-first eviction), so any member message after an idle gap re-pays the full thread at uncached input rates, and every tool round-trip re-pays it at cached rates all day. The 128k auto-compact ceiling (#125) bounds the worst case; this PR stops threads from *carrying* yesterday's bulk into today at all — and removes the "member messages, Murph stalls 30s compacting first" turn.

Design plan (committed with this PR, adversarially reviewed against the codebase before implementation): `agent-docs/exec-plans/completed/2026-06-11-idle-stop-thread-compaction.md`.

## What

When a runner idle-stops (TTL path only — never deploy evacuation) and the warm thread exceeds **100k tokens**, the runtime compacts it before the workspace snapshot, so the next wake resumes a ~40k thread. Four small primitives at existing seams:

- **`compactWarmCodexThread`** (assistant-engine): non-turn compaction on the warm-process module, sibling to the existing lifecycle exports. Thread size comes from retained `tokenUsage` vitals (zero extra RPCs). **Kill-don't-cancel**: any abort/timeout/failure poisons the codex child — always safe because rollouts only contain completed entries — so a pending wake can never queue behind a compact.
- **`buildAssistantMaintenanceUsageRecord`** (hosted-execution): synthetic-turn-id usage record (`automation_idle_compact`) flowing through the existing record route; `credentialSource` resolved like turns so member-key members are never allowance-debited.
- **`runHostedIdleCheckpointMaintenance`** (assistant-runtime): one gated step in the idle loop — skips on shutdown, pending work (budget-exhausted backlog or imminent projected wake), unpriced model, or sub-threshold threads; fail-open structurally; logs `workspace.checkpoint.idle_compact` with outcome/duration/usage-captured.

No cron, no timezone logic, no schedules: overnight is just the longest idle gap.

## Verification

- Engine scripted lane **9/9 against the real pinned codex binary**: off-turn compact with usage captured; **cold-resume of the compacted rollout after killing the warm process** (the production payoff path); mid-turn compact attempt skips without disturbing the live member turn; abort/provider-failure mid-compact recover cleanly; below-threshold makes zero provider calls
- Runtime unit 7/7 (shutdown/pending-work/unpriced-model skips, wake-abort + re-notify, shutdown-mid-compact abort, credential attribution, record-failure fail-open); hosted-execution 166; full `pnpm test:diff` lane over all touched owners exit 0; typechecks clean
- Audits run + resolved: pre-implementation plan stress-test (2 reviewers), security-privacy-review (member-credential attribution; SIGTERM mid-compact abort), simplify (10 findings applied), coverage-write, task-finish-review (4 findings applied incl. pending-work gate and guarded usage port)

Post-deploy verification: `idleCompactDurationMs`/`idleCompactThreadTokensBefore` in runtime logs, `automation_idle_compact` rows in `hosted_ai_usage`, and the gated `RUN_HOSTED_CODEX_AUTOCOMPACTION_E2E` live lane for the murph-shaped post-compaction floor.

DEPLOYMENT CONCERNS: ships entirely in the next Cloudflare runner deploy. No web coordination required — final review verified the maintenance usage record passes the currently deployed web parser, storage uniqueness, and allowance pricing unchanged. Watch for `activity-expired-active-operation` log events coinciding with `idle_compact` stages (none expected; activity renewal holds during the bounded 120s compact window).

---
<a href="https://app.blacksmith.sh/cobuildwithus/codesmith/murph/pr/130"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783782337&installation_id=127572939&pr_number=130&repository=cobuildwithus%2Fmurph&return_to=https%3A%2F%2Fgithub.com%2Fcobuildwithus%2Fmurph%2Fpull%2F130&signature=7ea15118456433313a563d507d712dcb192c6638e409c3112c96380a22743d54"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


rocketman-21:
Review addressed in 5fa93b389 — all six items verified as real before fixing, then fixed:

1. **sessionId identity (blocker)**: maintenance records now carry the real Murph assistant session id, resolved via a new `findAssistantSessionIdByCodexThreadId` store helper (persisted `resumeState.threadId` → session). The builder input is renamed `assistantSessionId` and the codex thread id lands in `providerRequestId` — the two identities can no longer be conflated. If no session matches the thread, **no record is written** (ambiguous identity never reaches the ledger; covered by a unit test).
2. **Wake behind usage recording (blocker)**: recording is now fire-and-forget (`void …catch`), mirroring turn usage; the 30s usage-port timeout can no longer delay a wake.
3. **Runtime fail-open**: the engine call is wrapped; a thrown exception returns `{kind:"failed", reason:"exception"}` and the checkpoint proceeds (unit-tested with a rejecting mock).
4. **Kill-don't-cancel made true**: failure paths now `void poison(...)` — the unusable flag is synchronous and SIGTERM→SIGKILL teardown completes out of band. `stop()` is memoized for concurrency safety (a racing turn's replacement stop awaits the same teardown), with the memo cleared on unsuccessful teardown to preserve retry semantics — the full 122-test adversarial process suite passes.
5. **Method-spelling drift**: one shared `isCodexThreadTokenUsageUpdatedMethod` (all five spellings) used by both the vitals reader and the turn-correlation allowlist.
6. **Skip-reason split**: `missing_model` vs `unpriced_model`.

Mergeable state was resolved separately (origin/main merged with two semantic guards for #127's subagent threads: vitals are bound-thread-only, and the compact binding routes foreign-thread messages through the same tolerance/denial path).

Verification: engine scripted 9/9 (real pinned binary), adversarial runtime suite 122/122, hosted-execution 166, runtime unit 9/9, full `test:diff` lane exit 0, typechecks clean.

rocketman-21:
Round-2 review addressed in c72a7bc83:

1. **Wake-after-abort teardown wait**: investigated the "truly detach the slot" option and rejected it on safety grounds — `getOrStartWarmCodexProcess` awaits the old stop *deliberately*, because two app-server processes on one CODEX_HOME would be concurrent writers to the same rollout JSONL. So the invariant is now stated honestly and bounded tightly, with the teardown policy split by what actually comes next:
   - `aborted` (wake/shutdown racing) → **detached** poison; a racing turn joins the memoized stop — bounded by the 3s SIGTERM ceiling, typically milliseconds, never the provider call or compact timeout
   - `timeout`/`rpc_error`/`process_exit`/exception → **awaited** teardown, because the idle checkpoint is next and the snapshot must never capture a rollout mid-teardown (a hazard the stress pass surfaced in the fully-detached version)
   - The bound is now proven by test: the abort case holds the provider request open 8s with a 30s compact timeout, then asserts the immediately following member-turn resume completes in <8s against the real binary.
2. **Accounting path fully fire-and-forget**: session resolution + record write now run in one detached `void (async () => …)().catch()` — nothing on the wake-sensitive path does telemetry I/O.
3. **Observability nit**: expected aborts log as `done`; only `timeout`/`rpc_error`/`process_exit`/`exception` hit the error-level `fail` status.

Stress pass on these changes also re-verified closure captures and sync-throw handling in the detached accounting block, and the memoized-stop retry semantics (the adversarial 122-test process suite caught a retry regression in an earlier iteration; it's pinned now).

Verification: engine 131/131 (scripted + adversarial runtime suites), runtime unit 9/9, hosted-execution 166, full `test:diff` lane exit 0, typechecks clean.

PR #110, opened 2026-06-10
fix(hosted): durable conversation consumed watermark + rollout SIGTERM checkpoint

rocketman-21:
## Incident

2026-06-10 ~2:08 AM ET: a user got duplicate assistant replies. The 2:01 AM deploy's runner-image rollout SIGTERM'd (exit 143) the active runner container before any workspace snapshot; the replacement lease restored a 9-minute-stale snapshot, replayed conversation seqs 186–188 as foreground replay with `knownInputIdCount: 0`, and re-replied to an already-answered message (re-logging an experiment session and generating a second image).

Root cause: the mailbox consume cursor and replied-state lived only in workspace-local files, durably captured solely by the idle-shutdown snapshot (~15 min cadence). Any ungraceful runner death inside that window replayed and re-replied — every runner-image deploy was a trigger.

## Fix 1 — durable consumed watermark (correctness)

- `hosted_mailbox_lane_counter.consumed_seq` (additive migration, default 0)
- Runtime acks the conversation watermark after a clean foreground pass only (zero failed replies, no pending foreground input) via new runner-allowlisted `POST /api/internal/hosted-mailbox/consume` — best-effort, monotonic max, clamped to the lane append high-water (a buggy/compromised runner cannot suppress unseen future messages)
- Mailbox fetch returns `consumedSeqByLane`; replayed items at/below the floor are re-staged as conversation context with `replyTarget: null` — context preserved, reply candidacy removed
- `packages/assistant-engine` (local runtime) untouched; all changes live in hosted layers

## Fix 2 — rollout SIGTERM checkpoint (state quality)

Container entrypoint catches SIGTERM and threads a `shutdownSignal` into the runtime idle wait: the idle window is treated as elapsed and the normal `idle_shutdown` checkpoint runs inside the termination grace period. Active foreground turns are not interrupted (Fix 1 owns correctness there).

## Compatibility

Optional port method + optional response field; either deploy order degrades to today's behavior (old web + new runner → consume 404s into the best-effort catch; new web + old runner → extra field ignored).

## Verification

- Typechecks: assistant-runtime, cloudflare-runner, hosted-web (prisma generate)
- `pnpm test:diff` across touched owners green (assistant-runtime 1214, web 2253, hosted-execution/cloudflare suites)
[...]

PR #133, opened 2026-06-11
feat(hosted): instrument consume-ack skips and durable container fatal traces

rocketman-21:
## Why

Three unattributable runner-container deaths on 2026-06-11 (00:36, 04:47, 16:53 UTC) each rolled members back to a stale snapshot; the 04:47 one produced duplicate, context-rolled-back replies on linq + email. Root-cause evidence (prod `hosted_runtime_log`, `hosted_mailbox_lane_counter`, CF workers telemetry):

1. Containers die **unrequested with exit code 1** ~2.5–3 min after the last foreground pass, before any durable checkpoint. The in-container cause is unrecoverable post-hoc: container stdout is console-only and never reaches a queryable sink.
2. The durable consumed watermark (PR #110) — built to make exactly this replay reply-safe — has **never engaged in prod**: `consumed_seq = 0` for every member/lane ever, and telemetry shows **zero** POSTs to `/api/internal/hosted-mailbox/consume` since deploy (385k rows scanned). The ack has four silent early-returns; static analysis clears three of them on real reply passes, so the tripping gate is unconfirmed — and unlogged.

This PR is **instrumentation only** (no behavior fixes): make the dead watermark gate name itself, and make the next container death attributable.

## What

**Consume-ack skip/advance logging** (`packages/assistant-runtime`, `packages/hosted-execution`)
- Every early-return in `acknowledgeHostedConversationMailboxConsumedBestEffort` now writes `mailbox.consume_ack_skipped` with a `skipReason` (`reply_failed` / `reply_outcome_missing` / `consume_port_missing` / `pending_assistant_input` / `empty_watermark`); unexpected reasons log at `warn`.
- Successful advances log `mailbox.consume_ack_advanced` with the acked seq.
- Safe hardening: a missing `mailboxPort` no longer throws (`mailboxPort?.consume`), it logs a skip.
- New event codes added to `HOSTED_RUNTIME_LOG_EVENT_CODES`.

**Durable container fatal traces** (`apps/cloudflare`)
- New `container-fatal-report.ts`: builds a redacted payload via the shared safe-error helper (`buildHostedExecutionSafeErrorDetails`; non-Error values stay out entirely) and POSTs it best-effort (1.5s timeout, all failures swallowed) to a new internal endpoint `runner-control.worker/v1/container-fatal`.
- Entrypoint reports before exiting 1 from: `uncaughtException` / `unhandledRejection` (new CLI-only process handlers preserving Node's exit-1 contract, with a hard-exit backstop), both poison paths (shell isolation, ambiguous aborted job), and startup failure.
- Egress intercept handles the path at the stateless worker layer (never forwarded to the DO — the DO may be the wedged component) and emits one sanitized, size-capped (8KB) structured worker log. Deliberately reachable without a bound user or write fence: these deaths happen outside invocations when neither exists; enum-ish fields pass a strict regex allowlist, free-form details only pass through the shared sanitizer.

**Deferred:** readiness-timeout tuning (8s budget lost a race with an 8.35s cold boot at 04:47:38, burning a generation) lives in `runner-container.ts` / `ensureContainerReady`, which is an active ledger lane — left for that lane.

## Verification

- `pnpm test:diff` exit 0 (62 files, 845 passed / 2 skipped after the run incorporating these changes).
- New tests: 5 consume-ack gate tests (skip reasons + advance + best-effort failure), 6 fatal-report module tests (payload redaction, endpoint/headers/timeout, swallow-on-failure), 6 egress-route tests (204 happy path, bound-user presence flag (privacy posture: raw user ids never persist in worker logs), hostile-field normalization + no-relay assertion, 405/413, no DO forward), extended poison test asserting the fatal report fires before the exit scheduler.
- 4 pre-existing exact-sequence assertions updated for the new log writes.
- Post-deploy: the first real reply pass emits either `consume_ack_advanced` or a `consume_ack_skipped` row naming the dead gate; the next unrequested container stop should be preceded by a `container-fatal` worker log.

DEPLOYMENT CONCERNS: the new runtime-log event codes are posted by the runner and validated by the web route parser from the same `@murphai/hosted-execution` source — deploy **web (Vercel) before Cloudflare**, mirroring PR #110's order. Until the web deploy lands, runners on the new build would have their consume-ack log entries rejected by the old parser (best-effort writes; no functional impact).

---
<a href="https://app.blacksmith.sh/cobuildwithus/codesmith/murph/pr/133"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783791487&installation_id=127572939&pr_number=133&repository=cobuildwithus%2Fmurph&return_to=https%3A%2F%2Fgithub.com%2Fcobuildwithus%2Fmurph%2Fpull%2F133&signature=ff1d57c1fb4fe25007b6b2262f5a8bc24ea182da80bb71f5c6e92e3f83da4971"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


rocketman-21:
Addressed the review (commit 660320124) — all three merge-bar items plus the hardening points:

1. **8KB silent-loss fixed**: the reporter now client-side caps the serialized payload at 6KB (headroom under the 8KB route limit); oversized sanitized details degrade to a `detailsTruncated: true` envelope that always carries `stage`/`errorCode`/`errorName`. Regression test included (32×300-char details → envelope under cap). The worker log also surfaces `detailsTruncated`.
2. **Attribution clarified**: keeping the existing privacy posture — the persisted worker record carries `userIdPresent` only (the structured-log builder nulls raw user ids by design). The bound-user header remains a correlation input, `containerIdPresent` was added to the fatal line, and real correlation comes from the adjacent DO lifecycle stop events. PR body updated to say "user-presence attribution", not user attribution.
3. **Both missing gate tests added**: `pending_assistant_input` (staged auto-reply input via `saveAssistantAutomationState` + `upsertAssistantInputEvent`) and `empty_watermark` (empty mailbox). All five skip reasons are now pinned.
4. (Review §5) The hard deadline now lives **inside** `reportHostedContainerFatalBestEffort` (Promise.race with a +500ms resolving timer), so every exit path gets the same bound even if fetch ignores its abort signal — tested with fake timers against a never-settling fetch. The entrypoint backstop remains as defense in depth.
5. (Review §4 / security audit) Chose rate-limiting + documentation over a fatal-report token: per-isolate fixed window (5 logs/min + one suppression marker) bounds forged-log noise, and the deliberately-unauthenticated trust boundary is now documented in `agent-docs/SECURITY.md` ("treat fatal lines as container-asserted diagnostics, correlate with DO lifecycle stop events"). A token would add env/config surface for a log-only sink whose worst case is noise — happy to revisit if you disagree.

Also from the security audit: process-fatal handlers now set a module-level poisoned flag so the entrypoint rejects new invocations during the ≤1.5s pre-exit report window (`/health` reports `poisoned: true`), closing the zombie-intake gap. Skip-log levels demoted to info except `consume_port_missing` (the one true wiring failure) to avoid recurring warn noise on routine background passes.

Verification: `pnpm test:diff` exit 0; 199 cloudflare tests + 850 assistant-runtime tests green.

PR #135, opened 2026-06-11
test(cli,cloudflare): import-surface contract and bundle byte-budget regression guards

rocketman-21:
> Stacked on #134 — review/merge that first; this PR's base is its branch and I'll rebase onto main after it merges.

## Why

The June 2026 vault-cli latency regression happened because the scoped lazy-loading architecture had an invariant — "a scoped invocation loads only a small module set" — with **no enforcement**. One static import (llms-normalizer → command manifest → every command module → ink/react/yoga-WASM) silently defeated it on every invocation, and the cost was only visible in the hosted container. Static imports look free in review; the graph only grows. These guards make the invariant executable and the surface ratchetable.

## Guard 1: import-surface contract (`packages/cli`)

`vault-cli-import-surface-contract.json` lists every package the scoped hot path may resolve — currently **16 packages**, with the entire regression class (`ink`, `react`, `yoga-layout`, `@murphai/assistant-*`, `@murphai/setup-cli`) conspicuously absent. The test runs three representative scoped commands (`wearables day`, `meal totals`, `list`) against the built CLI under a `node:module` resolve hook (`registerHooks`, so both ESM and CJS resolutions are observed), normalizes every resolved module to its owning package (pnpm `.pnpm` layout + workspace dirs; unknown origins fail loudly), and asserts:

- the resolved package set ⊆ the contract → **adding anything to the hot path becomes an explicit reviewed contract edit**
- per-probe resolved-module ceilings (measured 297/430/280 on 2026-06-11, +~30% headroom) → creep *within* allowed packages is bounded too
- each probe exits with `missing_vault` → proves it traversed routing → services → vault resolution before recording, so the measurement cannot go vacuous
- unused contract entries are surfaced informationally → removals are wins; the contract is the collapse ratchet

Fully deterministic — no timing assertions. **Non-vacuity proven by mutation**: re-introducing a manifest-class static import fails both the ceiling (297 → 901 modules) and the contract (12 new packages listed by name); production dist verified byte-identical after revert.

## Guard 2: bundle byte budgets (`apps/cloudflare`)

The assembly step now asserts esbuild-metafile byte budgets on the real installed artifact: total **7,052,933 B measured → 9,000,000 B budget**, entry chunk **15,569 B → 20,000 B** (baselines + date documented at the constants). Violations throw with actual-vs-budget plus the top-10 largest inputs sorted descending, so the culprit is diagnosable from the build log alone. **Mutation-proven**: budget set below reality fails the real assembly listing the heaviest inputs. (Found and pinned a real esbuild subtlety along the way: with `splitting`, dynamic-import chunks also carry `entryPoint`, so entry detection is `entryPoint` + basename, locked by a decoy-chunk test.)

Plus warn-only parity-probe duration lines in every assembly log for longitudinal trend visibility (explicitly never a hard assertion — timing budgets on shared runners flake):

```
vault-cli bundle size: total 7058928B of 9000000B budget, entry 15569B of 20000B budget
parity probe `--llms-full --format json`: unbundled 489ms, bundled 324ms
parity probe `wearables day 2026-01-01 --format json`: unbundled 168ms, bundled 129ms
```

Deliberately NOT included: CI wall-time budget tests — noisy shared runners make them flake, and a stable budget would be too loose to catch regressions of the observed size. The deterministic graph/byte guards catch the cause; prod turn profiles catch the symptom.

## Verification

- Contract test, shape guard (`verify-package-shape`), all 6 bundle tests, and both package typechecks green on the rebased base.
- Full real `runner:bundle:assemble-only` green with all guards active (log excerpt above).
- Both guards mutation-proven non-vacuous (details above).

---
<a href="https://app.blacksmith.sh/cobuildwithus/codesmith/murph/pr/135"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783798060&installation_id=127572939&pr_number=135&repository=cobuildwithus%2Fmurph&return_to=https%3A%2F%2Fgithub.com%2Fcobuildwithus%2Fmurph%2Fpull%2F135&signature=b117e78b26aad9d5df171c4cf9f2fc6b3045bcbb074ae138746ce0630313b0df"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


rocketman-21:
Added `325b3ad02` — the contract guard caught its first real cut on day one:

**`occurred-at-option.ts` statically imported `loadVault` from `@murphai/core`**, and since that module registers on every command group supporting `--occurred-at` (meal, measurement, workout, document, capture, experiment, intervention), the entire core *write* runtime + its runtime-state subtree (~120 modules) loaded on every read-only scoped invocation. The consuming function was already async; the import is now lazy inside the only call site that needs the vault.

Measured effect: `meal totals` 430 → 311 resolved modules; `@murphai/core` removed from the import-surface contract (16 → 15 packages); runtime-state down to 3 subpath modules. Probe ceiling re-measured and documented.

Remaining scoped surface (~300 modules): zod ~81 (schema-first floor), incur's eager yaml/MCP-SDK/json-schema surface ~110 (upstream candidate — incur eagerly imports yaml and the MCP server SDK on every invocation), murph domain schema packages ~85.

rocketman-21:
Addressed the review in `d0b2c6172`:

1. **Probe child lifecycle owned** — 60s SIGTERM with 5s SIGKILL escalation inside `runScopedImportSurfaceProbe`; a wedged CLI now rejects with the probe label + captured output head instead of orphaning a child under a bare test timeout.
2. **Unused contract entries now FAIL** — agreed that warn-only weakens the ratchet; an entry no probe resolves must be removed, so the contract can only ever describe the real surface.
3. **Ceiling failures name the culprits** — per-package module counts are aggregated before the ceiling assertion and the top groups are included in the failure message (e.g. `zod 81, yaml 72, ...`).
4. **Contract hygiene** — duplicate and empty/whitespace entries are rejected alongside the existing sorted-order check.
5. **incur eager surface** — tracked separately as agreed: upgrade to 0.4.8 confirmed NOT to help (same eager yaml/MCP imports, `Cli.create()` even constructs an MCP HTTP handler eagerly); an upstream lazy-loading patch against wevm/incur is in preparation.

PR #320, opened 2026-06-26
Codex/linq first contact admission

rocketman-21:
## Why
This PR hardens the Linq first-contact admission flow so the OpenAI classifier can protect onboarding from non-signup traffic without persisting raw provider details or bricking legitimate new-member onboarding during provider outages.

## User-visible Goal
Unknown one-on-one Linq first contacts that look like signup-link candidates should still receive the normal hosted onboarding signup link when the classifier is unavailable, while explicit low-confidence/unsupported/non-Murph/refusal/content-filter blocks continue to stop member creation, invite creation, replies, read receipts, wakes, and mailbox side effects.

## Invariants
- Classifier prompts, raw responses, raw provider errors, and provider response bodies are never persisted.
- Terminal allow/block decisions remain event-id keyed so replay/concurrency observe the same result.
- Classifier-unavailable states do not persist synthetic decisions; they log only bounded metadata and continue through the normal signup-link path.
- Deterministic spam, group chats, own messages, active members, explicit routes, and non-invite paths still bypass or block exactly as before.

---
<a href="https://app.blacksmith.sh/cobuildwithus/codesmith/murph/pr/320"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785077253&installation_id=127572939&pr_number=320&repository=cobuildwithus%2Fmurph&return_to=https%3A%2F%2Fgithub.com%2Fcobuildwithus%2Fmurph%2Fpull%2F320&signature=6adaef3f643fb0fa32b78fdb3872cbc664158deb7d455e2c3d7e1c1ed657cd9d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


rocketman-21:
Closing in favor of #332, which solves the same exactly-once invariant with one table, one PK, and one INSERT instead of a 1,000-line processing/consumed receipt + lease + processing-owner-token state machine.

47 fix commits over ~42 hours chasing fence/lease/proof corner cases was the signal that the abstraction here was wrong, not that another round was needed — the side-effect layer below already enforces its own dedupe (mailbox keys, invite uniqueness, member upserts, daily-state counters), so a top-level admit-once is sufficient.

This anti-pattern is now documented as a named invariant in `docs/contracts/00-invariants.md` → **Minimal Mechanism Bias** (introduced in #332), so the same complexity explosion doesn't recur.

Not deleting the branch — the commit history of the chase is the best paper trail.

PR #332, opened 2026-06-27
Admit each Linq webhook event once via single-row PK

rocketman-21:
## Summary

Replace PR #320's processing/consumed receipt + processing-owner-token + lease state machine with the minimal idempotency primitive: one table with one primary key on `event_id`, and an idempotent INSERT at the top of the Linq webhook handler. A retry collides on the PK and returns a duplicate-event no-op; the original delivery continues into side effects that already carry their own dedupe (mailbox dedupe keys, `HostedInvite` uniqueness, `HostedMember` upserts, daily-state counters).

## Why

PR #320 grew to **7,374 additions across 43 files** and **47 fix commits** chasing edge cases in a multi-state processing protocol — fences, lease cutoffs, owner tokens, stale-receipt reclaim, replay-vs-reclaim corner cases, explicit-route ownership preservation. The underlying invariant was real ("exactly-once Linq webhook handling"), but the abstraction was wrong: each round of review found another fence/lease/proof corner that needed a new helper, and the line count kept growing instead of shrinking.

Side effects below the webhook handler are already individually idempotent:

- Mailbox writes dedupe on `(channel, externalEventId)`.
- Invites are unique per `(memberId, channel)`.
- Member identity goes through phone/email upserts.
- Daily-state counters are upserts.

If each side effect is retry-safe, the only thing a top-level fence buys is one fewer OpenAI classifier call per duplicate — and that's already bounded by main's per-contact admission budget (#324). A single-row admit-once at the entry covers the cost concern with **one table, one INSERT, no states, no leases, no owner tokens**.

PR #320 will be closed in favor of this change.

## What changed

- **New model** `HostedLinqProcessedEvent` — `event_id @id`, `created_at` with index.
- **New migration** `20260627150000_linq_processed_event`.
- **handleHostedOnboardingLinqWebhook** now calls `admitHostedLinqWebhookEventOnce` immediately after acquiring prisma; on a Prisma `P2002` it returns `{ ignored: true, ok: true, reason: "duplicate-event" }` and skips all planning + side effects.
- **New regression** in `hosted-onboarding-webhook-idempotency.test.ts` asserts a duplicate `event_id` triggers no side effects (no mailbox append, no invite issue, no chat send, no read receipt, no runner nudge, no runtime signal).
- **New invariant** in `docs/contracts/00-invariants.md`: "Minimal Mechanism Bias" — three rules naming PR #320 as the anti-pattern so this complexity explosion doesn't repeat.

## Trade-offs (intentional)

- **Mid-processing crash forfeits the retry.** If the first delivery inserts the row but fails before completing side effects, a retry will see the row and no-op. This is acceptable because (a) Linq webhook retries are eventual-consistency, not exactly-once delivery, (b) the side-effect layer's own retries can still recover most user-visible state, and (c) the alternative (PR #320's processing/consumed state machine) was 1,000 lines of bookkeeping chasing this exact case and still missed corners.
- **Single PK, not composite.** We do not include any "claim version" or "processing owner" — exactly the columns whose accumulation drove PR #320's complexity.

## Validation

```
pnpm --dir ../.. exec vitest run --config apps/web/vitest.workspace.ts --no-coverage \
  apps/web/test/hosted-onboarding-linq-dispatch.test.ts \
  apps/web/test/hosted-onboarding-linq-first-contact-admission.test.ts \
  apps/web/test/hosted-onboarding-linq-route.test.ts \
  apps/web/test/hosted-onboarding-linq-webhook-auth.test.ts \
  apps/web/test/hosted-onboarding-linq-usage-reset-e2e.test.ts \
  apps/web/test/hosted-onboarding-privacy-foundation-migration.test.ts \
  apps/web/test/hosted-onboarding-webhook-idempotency.test.ts
```

→ **135/135 passing.**

## Deployment

- The web deploy must run the new Prisma migration before the new web code handles Linq webhooks — same constraint as #324.
- No Cloudflare tandem deploy required.

## References

- Closes #320.
- Invariant: `docs/contracts/00-invariants.md` → "Minimal Mechanism Bias".
- Related: #324 (per-contact admission budget — keeps cost bound on classifier calls).

---
<a href="https://app.blacksmith.sh/cobuildwithus/codesmith/murph/pr/332"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785178963&installation_id=127572939&pr_number=332&repository=cobuildwithus%2Fmurph&return_to=https%3A%2F%2Fgithub.com%2Fcobuildwithus%2Fmurph%2Fpull%2F332&signature=37b4a7907a7733f07b8256492f7ce487a3fbb147ffb88867fdacd02a3e57535e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/apps/web/prisma/migrations/20260627150000_linq_processed_event/migration.sql
+++ b/apps/web/prisma/migrations/20260627150000_linq_processed_event/migration.sql
@@ -1,0 +1,9 @@
+CREATE TABLE "hosted_linq_processed_event" (
+  "event_id" TEXT NOT NULL,
+  "created_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+
+  CONSTRAINT "hosted_linq_processed_event_pkey" PRIMARY KEY ("event_id")
+);
+
+CREATE INDEX "hosted_linq_processed_event_created_at_idx"
+  ON "hosted_linq_processed_event"("created_at");
--- a/apps/web/prisma/schema.prisma
+++ b/apps/web/prisma/schema.prisma
@@ -846,6 +846,14 @@ model HostedLinqFirstContactAdmissionDecision {
   @@map("hosted_linq_first_contact_admission_decision")
 }
 
+model HostedLinqProcessedEvent {
+  eventId   String   @id @map("event_id")
+  createdAt DateTime @default(now()) @map("created_at")
+
+  @@index([createdAt])
+  @@map("hosted_linq_processed_event")
+}
+
 model HostedLinqFirstContactAdmissionBudget {
   participantContactLookupKey String   @map("participant_contact_lookup_key")
   eventId                     String   @map("event_id")
--- a/apps/web/src/lib/hosted-onboarding/webhook-service.ts
+++ b/apps/web/src/lib/hosted-onboarding/webhook-service.ts
@@ -1,6 +1,6 @@
-import type {
+import {
   Prisma,
-  PrismaClient,
+  type PrismaClient,
 } from "@prisma/client";
 
 import { getPrisma } from "../prisma";
@@ -119,6 +119,21 @@ export async function handleHostedOnboardingLinqWebhook(input: {
     }
 
     const prisma = input.prisma ?? getPrisma();
+
+    const duplicateResponse = await admitHostedLinqWebhookEventOnce({
+      eventId: event.event_id,
+      prisma,
+    });
+    if (duplicateResponse) {
+      responseReason = duplicateResponse.reason ?? null;
+      finishHostedOnboardingTiming(timing, "completed", {
+        eventIdSuffix: toHostedOnboardingLogIdSuffix(eventId),
+        eventType,
+        responseReason,
+        signalAbortedBeforeReturn: input.signal?.aborted ?? false,
+      });
+      return duplicateResponse;
+    }
     const planTiming = startHostedOnboardingTiming(
       "hosted-onboarding.webhook.linq.plan",
       {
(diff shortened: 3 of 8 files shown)


rocketman-21:
Closing after a code review surfaced a load-bearing bug in this PR's design and a parallel discussion clarified that the existing mailbox-keyed dedupe at the planner layer already covers the duplicate-event case adequately.

**Why this PR shouldn't ship:**
- The admit-once INSERT commits OUTSIDE the planner transaction. Any post-admit failure (OpenAI timeout, container restart, drain throw) leaves the row durable while the retry sees \`duplicate-event\` and silently no-ops — the user's message is permanently dropped. That's strictly worse than no change at all.
- \`reason: "duplicate-event"\` diverges from the existing \`"duplicate-webhook-event"\` already used by the planner and route ingress, breaking telemetry buckets.
- The new wake-handoff path is removed; old plan-level duplicate detection re-nudged the runner as defense against lost wake signals (known cause of prior duplicate-reply incidents).
- The new table is Linq-only when Telegram/WhatsApp have the same shape — exactly the mechanism-per-channel proliferation the new \"Minimal Mechanism Bias\" invariant warns against.

**Why no replacement PR is needed:**
- Main's mailbox-keyed dedupe in the planner already returns \`duplicate-webhook-event\` inside a transaction, so the original concern is covered.
- Main's per-contact admission budget caps OpenAI duplicate-classifier cost at 4 per spammer — bounded.

The \"Minimal Mechanism Bias\" invariant from this PR's doc change is the one genuinely independent win and will be shipped directly to main.

The explicit-route admission gap PR 320 was also trying to close is a separate, currently-low-risk concern; fix when a real incident demands it.

PR #369, opened 2026-07-02
Group chats: typed profile-name projection + member roster on murph.group read_current

rocketman-21:
## Why this PR exists

In a group chat, Murph cannot connect a chat sender handle to a group member or to the health data that member shares, and the preferred name it learned during onboarding lives only in freeform memory text that no deterministic projector may parse. Group Murph either re-asks people their names or misattributes shared data.

Joining a group introduces you by name automatically. Group Murph addresses each participant by their preferred name and attributes shared records (e.g. sleep times) to the right person, without ever re-asking a question answered during onboarding.

How: (1) a typed canonical vault document `bank/profile.md` (`@murphai/contracts` schema, `vault-cli profile show|set-name`, deterministic `@murphai/query` reader) becomes the single home for the display name; (2) a new bounded `profile-name.v0` vault-share projection kind (single fixed recordKey) delivers it through the existing deliver path; (3) the grant is membership-implied — join-accept and group-create always grant `profile-name.v0` (launch consent asserted on every join), while health kinds stay individually selectable and `profile-name.v0` is excluded from join-page checkboxes and the `create_join_link` request contract; (4) `murph.group read_current` returns a server-derived member roster `{ memberId, handle, role, grantedVaultShareProjectionKinds }`; (5) group-chat skill guidance reads the roster.

## Invariants this PR must preserve

- `HostedVaultShare` stays the only data-sharing grant. The profile-name share is a normal grant row the server revoke machinery fully supports; its lifetime is deliberately coterminous with membership (the join page states the share plainly; the join-page revisit deselect flow governs only selectable health kinds).
- Deterministic projection only — the display name is never parsed out of freeform memory text.
- The deliver response stays a bare function of share configuration alone (`delivered` / `no-active-share`); record staleness or content can never probe finer-grained share state. The recency age bound applies to time-series kinds only. `profile-name.v0` is registered as a current-state kind (`isHostedVaultShareCurrentStateProjectionKind`): one parser-enforced fixed recordKey, no age bound (a name set long before the first join must still deliver), and a content-only delivery revision — `occurredAt` is excluded from the dedupe hash so a corrupted grantor runtime cannot mint unbounded mailbox dedupe keys by drifting timestamps on an unchanged name.
- Parser envelopes stay closed: `create_join_link` requests are closed over the individually selectable kinds; the roster keeps all kinds; no envelope widening.
- Never re-ask members for their name.

`DEPLOYMENT CONCERNS:` Deploy Cloudflare (runner) before web. The runner-side parser tolerates group summaries without `members` (old web), but pre-change runner parsers reject the widened `read_current` response via the closed-key check. Warm containers on the old bundle keep working during rollout — they just do not read rosters until recycled. Post-deploy check: `murph.group read_current` from a group runtime returns `members` and the join page still lists only selectable health kinds as checkboxes.

## Review provenance

A pre-landing Codex deep review (xhigh) of the full diff produced 6 findings; 4 accepted and fixed test-first (stale-drop of old profile names in the deliver route, schema-catalog publication, `create_join_link` parser closure, control-character contract/delivery mismatch), 2 rejected with rationale recorded in the exec plan (membership-implied revoke semantics; roster read cost bounded by real group size — roster is owner-only at create, `read_current` runs outside any transaction). A focused second Codex round on the fix delta surfaced 2 more, both fixed: content-only revision identity for current-state kinds (above), and the published JSON schema now carries the display-name pattern (non-blank, no control characters), not just min/max length.


rocketman-21:
ReviewGPT round 1 (head fa907fe8f3) triage: 1 finding, rejected.

Finding: kind-dependent recency filtering in the deliver route is a branching correctness axis / potential replay-visibility escalation if projectionKind is client-influenced.

Rejection rationale: projectionKind is not an interpretation axis over the same records — it selects both the closed per-kind data schema the records must parse against (sleep-shaped data cannot pass as profile-name.v0: fixed recordKey, {displayName}-only shape) and the (grantor, kind) share fan-out, so a stale time-series record can never be resurfaced by switching kinds. The suggested collapse (always age-filter, normalize after) would reintroduce the High bug this PR fixes (names set >60 days before the first join silently never deliver), and a delivery-policy resolver adds machinery where the per-kind policy already has one owner seam (isHostedVaultShareCurrentStateProjectionKind in the vault-share registry, with content-only revision identity closing the dedupe-minting vector). Covered by tests: old profile name delivers, future-dated drops, sleep-times stale-drop unchanged, occurredAt drift dedupes.

rocketman-21:
ReviewGPT round 2 (head f3f44904c9): no Critical/High/invariant findings. 1 complexity finding, rejected.

Finding: N+1 identity fetch in readHostedGroupMemberRoster; proposes a batched readHostedMemberIdentityBatch.

Rejection rationale (same surface as pre-landing Codex finding 4, rationale recorded in the archived exec plan): the per-member identity reads run concurrently (Promise.all), membership and grants queries are already batched, the roster is owner-only at group creation and read_current runs outside any transaction, and real group chats are an order of magnitude below the 50-200-member scenario. Adding a batch identity reader is new store surface for a speculative scale concern; if group sizes grow, the single seam to batch is readHostedGroupMemberRoster.

Loop status: two consecutive rounds with zero accepted findings.

rocketman-21:
ReviewGPT round 3 (head cfd21a0cd3, rebased + member-side backfill guidance): no Critical/High findings. 1 complexity finding, rejected.

Finding: replace the two explicit per-kind projection offers in vault-share-projection.ts with a registry-driven loop.

Rejection rationale: orchestration is already shared in offerHostedVaultShareKindBestEffort — the per-kind surface is two array entries and a test-only reader override apiece. A projection registry at N=2 is premature abstraction (repo simplicity rules); per-kind best-effort isolation, no cross-kind state, and empty-read-means-no-delivery are all already held by the existing helper and covered by tests (mixed availability included). Revisit as a mechanical collapse if/when a third projection kind lands.

Loop status: three consecutive rounds with zero accepted findings.

PR #372, opened 2026-07-03
fix(assistant): deliver progress updates in queue-only hosted auto-replies

rocketman-21:
## Why this PR exists

Hosted members saw long silent waits during slow turns (onboarding lab/supplement ingestion especially): Murph's mid-turn progress updates never arrived, and the model believed it had sent them. Live container rollouts showed GPT-5.5 emitting progress text as commentary-phase agent messages that Murph normalized correctly and then dropped, because every hosted-runner turn runs `deliveryDispatchMode: queue-only` and the progress eligibility gate treated all queue-only turns as progress-ineligible. With `progressDelivery` null, the `send_progress_update` tool was never offered either, so both progress paths were dead on hosted.

## User-visible goal

During a hosted interactive reply (a member is actively waiting), Murph's mid-turn progress updates — whether sent via the `send_progress_update` tool or emitted as GPT-5.5 commentary-phase messages — are delivered to the member's channel immediately, subject to the existing per-turn limit, spacing, and dedupe rules. Verified live on the dev stack: a Telegram progress update was delivered with a provider message id before the final reply.

## What changed

- `shouldCreateAssistantProgressDelivery`: queue-only dispatch suppresses progress only when the turn has no active audience — `automation-auto-reply` turns stay progress-eligible; cron/background turns remain suppressed. The predicate moved to `progress-constants.ts` as the single source of truth.
- `deliverAssistantProgressUpdate` now shares that predicate and fails loudly (typed `ASSISTANT_PROGRESS_DELIVERY_SUPPRESSED`) instead of returning success-shaped silent no-ops the model reported to users as "sent".
- Every progress send attempt logs a metadata-only outcome line (sent / skipped with reason / failed, plus a dispatch summary with channel and target presence booleans) — the whole surface was previously unobservable.
- New gated real-binary e2e (`assistant-codex-commentary-real-e2e.test.ts`, `MURPH_RUN_REAL_CODEX_E2E=1`) proves commentary-phase agent messages route into progress delivery and stay out of the final reply.

## Invariants to preserve

- Final-reply delivery ownership is unchanged: queue-only turns still deliver final replies exclusively through the outbox with its retry/receipt machinery.
- Progress updates remain best-effort ephemeral sends: bounded by MAX_PROGRESS_UPDATES_PER_TURN (2), 2-minute spacing, and duplicate suppression; a progress failure never fails the turn.
- Background/cron turns and notification-decision/email contexts must never emit mid-turn progress sends.
- No message text, health data, or identifiers in the new log lines — metadata only.

---
<a href="https://app.blacksmith.sh/cobuildwithus/codesmith/murph/pr/372"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785649153&installation_id=127572939&pr_number=372&repository=cobuildwithus%2Fmurph&return_to=https%3A%2F%2Fgithub.com%2Fcobuildwithus%2Fmurph%2Fpull%2F372&signature=3fe2ea1534d45dd459414d6679c49781b0f4494b8de2cd2633b63c30550ef901"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/packages/assistant-engine/src/assistant/delivery-service.ts
+++ b/packages/assistant-engine/src/assistant/delivery-service.ts
@@ -6,6 +6,7 @@ import {
   parseTelegramThreadTarget,
 } from '@murphai/messaging-ingress/telegram-webhook'
 import { VaultCliError } from '@murphai/operator-config/vault-cli-errors'
+import { shouldCreateAssistantProgressDelivery } from './progress-constants.js'
 import { markAssistantFirstContactSeen } from './first-contact.js'
 import { ASSISTANT_FIRST_CONTACT_WELCOME_MESSAGE } from './first-contact-welcome.js'
 import { createHostedDeliveryId } from './hosted-delivery-id.js'
@@ -409,12 +410,19 @@ export async function deliverAssistantProgressUpdate(input: {
   turnId: string
 }): Promise<AssistantSession> {
   // Progress updates are ephemeral current-audience sends, not final-reply
-  // delivery or commit-aware outbox decisions.
-  if (!input.input.deliverResponse) {
-    return input.session
-  }
-  if (input.input.deliveryDispatchMode === 'queue-only') {
-    return input.session
+  // delivery or commit-aware outbox decisions. The caller gates on the same
+  // shared predicate before invoking, so failing it here means the delivery
+  // context diverged mid-send; fail loudly instead of returning a
+  // success-shaped no-op that the model reports to the user as "sent".
+  if (!shouldCreateAssistantProgressDelivery(input.input)) {
+    throw new VaultCliError(
+      'ASSISTANT_PROGRESS_DELIVERY_SUPPRESSED',
+      `Progress update suppressed: delivery context is not progress-eligible ` +
+      `(turn=${input.turnId} ordinal=${input.ordinal} ` +
+      `deliverResponse=${String(input.input.deliverResponse)} ` +
+      `dispatchMode=${String(input.input.deliveryDispatchMode)} ` +
+      `turnTrigger=${String(input.input.turnTrigger)}).`,
+    )
   }
 
   const deliveryFields = resolveAssistantCurrentAudienceDeliveryFields({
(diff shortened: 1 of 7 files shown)


rocketman-21:
**ReviewGPT round 1 triage** (audit-packages/pr-372-round-1.md, thread reviewed head 155bda99e7):

**High #1 — "exception replaces best-effort no-op contract, unguarded call sites may fail turns": REJECTED with evidence.**

The failure scenario is unreachable. `deliverAssistantProgressUpdate` has exactly two production call sites (`turn-progress.ts:141` default deliver, `local-service.ts:518/526` hosted deliver callback), and both execute exclusively inside `createAssistantProgressDelivery().send()`'s try/catch, which converts any throw into the structured `{kind: 'failed'}` best-effort result the dynamic tool reports honestly ("progress update failed during best-effort delivery"). The enclosing turn never fails — that behavior is pinned by the existing deliver-throws test ("outbox unavailable" → `kind: 'failed'`).

The suggested fix (restore silent `return input.session` or catch inside the function) would reinstate the exact production bug this PR fixes: a success-shaped no-op that made the model tell members updates were "sent" while nothing was delivered. The throw is intentional defense-in-depth for future direct callers.

Added committed proof in 793c1d1e36 (test-only): direct non-eligible calls reject with `ASSISTANT_PROGRESS_DELIVERY_SUPPRESSED` without reaching channel dispatch.

**Round 1 outcome: zero accepted findings.** Per the deep-review loop, test-only proof additions do not trigger a rerun.

PR #375, opened 2026-07-03
Unify hosted member access on one resolver; fix family-blind group-chat gates

rocketman-21:
## Why

The first live iMessage group chat (2026-07-03) was silently dropped: the linq group-chat webhook
gated the sender with the family-blind `hasHostedMemberActiveAccess` (member-level
`billing_status === 'active'`) while the sender was family-sponsored (`billing_status =
'not_started'`, access derived from the account group). A follow-up architecture review found
"does this member have access" had three parallel representations (own billing status, a 4-query
family read-time derivation, and a container owner-join with a vestigial hard-set `active` on the
container member row), two exported predicates where picking the wrong one is silent, and five
family-blind gates — all on the thread-container path.

## User-visible goal

Family-plan members get exactly the same access as direct-paid members everywhere: group chats
(provisioning a group thread with Murph now works for sponsored senders — the incident scenario),
thread-route delivery and egress, runtime wakes, internal runtime APIs, device-sync settings,
pages (`/join` no longer bounces a sponsored member to checkout), and usage gating for containers
whose owner is family-sponsored. No behavior change for direct-paid members.

## What changed

One primitive: `hosted_member.billing_status` means the member's OWN Stripe relationship only
(container members are now created `not_started` instead of a frozen fake `active`), sponsorship
stays on the edges that already exist (`hosted_account_group_membership`,
`hosted_thread_container.owner_member_id`), and a single new module
`apps/web/src/lib/hosted-onboarding/member-access.ts` owns the derivation:

- `hasActiveHostedMemberAccess` (pure) / `readActiveHostedMemberAccess` (one query):
  suspended → false; thread-container member → owner access only; person → own active billing OR
  active membership in an active, unsuspended group. Depth ≤ 2 by construction (owners cannot be
  containers).
- All access gates now go through it: linq/telegram/whatsapp webhooks, linq group-chat sender +
  explicit-route gates, thread-container provisioning, thread-route egress authority,
  runtime-access, usage-allowance container branch, internal runtime routes, request/app-session
  auth, page auth, member-channel-sync, reconciliation facts, group-tool.
- The family-blind predicate is no longer reachable as an access check:
  `hasHostedMemberActiveAccess` → `hasHostedMemberOwnActiveBilling` (and the assert variant),
  used only by billing surfaces that genuinely mean "own subscription".
- Deleted: `hasHostedMemberEffectiveActiveAccess(ForMember)`,
  `assertHostedMemberEffectiveActiveAccessAllowed`, `hasActiveHostedFamilyAccess`,
  `deriveHostedEntitlement`, the `familyAccessActive` parameter threading, and the read-time
  seat re-count inside `readHostedFamilyAccessForMember` (unreachable belt-and-braces; the seat
  invariant is enforced at write time — invite issuance/acceptance assert seat fit and the
  subscription webhook fails the whole group to `unpaid` on overage).

No schema change; no data backfill. Existing container rows with the vestigial `active` are
harmless (the resolver ignores a container's own status).

## Invariants preserved

- Family entitlement contract (`agent-docs/product-specs/hosted-family-plan.md`, updated in this
  PR): sponsored access fails closed on group inactive/suspended, membership removed, seat
  overage (write-time), suspension.
- Product-critical flow preservation: sponsored members gain authorized paths; solo members and
  billing-mutation surfaces keep exact prior semantics; suspended anything still fails closed.
- Billing monotonicity untouched: the Stripe write choke points are unchanged.

## DEPLOYMENT CONCERNS

Web-only change (`apps/web`); no Cloudflare deploy required and no compatibility window: the
runner interacts through internal HTTP routes whose response shapes are unchanged. Post-deploy
check: send a group iMessage from a family-plan member — the webhook should provision a thread
container (`hosted_thread_route` row) and Murph should reply in the group; previously the planner
logged `routeStage: 'group-chat-sender-inactive'`.

---
<a href="https://app.blacksmith.sh/cobuildwithus/codesmith/murph/pr/375"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785681738&installation_id=127572939&pr_number=375&repository=cobuildwithus%2Fmurph&return_to=https%3A%2F%2Fgithub.com%2Fcobuildwithus%2Fmurph%2Fpull%2F375&signature=6b7c4eec970efc27802edd0170446dd1b73f12932cd31aaa2e12eb0f05d977fd"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


rocketman-21:
**ReviewGPT round 3 finding (sponsored authority not row-locked in `requireHostedRuntimeActiveAccessForUpdateTx`): rejected — pre-existing designed behavior, and the fix costs more than the window it closes.**

1. **Not introduced by this PR.** On `main`, the exact same for-update gate already derives family access through `hasHostedMemberEffectiveActiveAccessForMember`, which reads `hosted_account_group_membership`/`hosted_account_group` with no row lock. This PR reroutes that read through the unified resolver and extends the same read-time semantics to the container-owner limb; the seam itself is the family plan's designed read-time derivation (see `agent-docs/product-specs/hosted-family-plan.md` § Entitlement — sponsored access is derived, never materialized, and revocation takes effect on the next read).
2. **Bounded consequence.** The race window is a single in-flight transaction: an operation that was fully authorized when it began completes moments after a concurrent revocation commits. No new post-revocation operation can start — the next gate read fails closed. That is equivalent to the revocation having landed one transaction later, which the family product contract already tolerates (removal revokes sponsored access going forward; it does not claw back in-flight work).
3. **Disproportionate fix.** Locking the membership + group rows inside every runtime write would serialize all of a family's members' mailbox/runtime writes on one `hosted_account_group` row and add cross-path lock-ordering constraints against the family billing/removal transactions (which take member-row locks in their own order). That is the broad-fence shape `docs/contracts/00-invariants.md` § Minimal Mechanism Bias tells us to reject without a production-proven failure: it adds contention and deadlock surface without removing a reachable, harmful outcome.

If we later find a concrete case where completing one in-flight delivery after revocation is itself a harm (e.g. a future share kind with revocation-sensitive content), the narrow fix is a post-append recheck on that specific delivery path, not a global authority lock.

PR #378, opened 2026-07-03
Resolve replayed OAuth callback deliveries instead of failing them

rocketman-21:
## Why this PR exists

A new user connected Oura on July 3: the connection succeeded, but their browser redelivered the callback URL ~1.6s after the first navigation (Junction mobile completion-page retry / Chrome iOS navigation re-commit). The second delivery hit an already-consumed `murph_state`, got a bare unstyled 400 page ("Device connection failed") with no way back into the app, and the user reasonably concluded the connection was broken. Server logs: `OAUTH_STATE_INVALID` at 14:45:05 UTC, two deliveries of the same callback URL 1.6s apart, connection `dsc_Xx0s40BTTSPU1oX4` created successfully by the first.

Root cause is a transport-contract mismatch: the callback is a browser-navigated GET (an at-least-once channel — refresh, tab restore, back navigation, provider completion-page retries), but state consumption was exactly-once (hard delete on consume), so normal redelivery was indistinguishable from an unknown/attacker state.

- A redelivered callback quietly returns the user to where the flow was headed (`returnTo`, or `/connect` if unknown), where the UI renders connection truth from the store. No error page for a connection that already completed.
- Any other callback failure with no stored `returnTo` (e.g. state expired after the 15-minute TTL) now redirects to `/connect` with the existing friendly error notice instead of dead-ending on bare HTML.

## What changed

- `consumeOAuthState` (Prisma + local SQLite stores) marks rows `consumed_at` instead of deleting them. A consumed-but-unexpired state now consumes to a new `replayed` status carrying the stored record; expired states still report `missing` and are purged by the existing `startConnection` sweep. Concurrent consumers resolve via the `consumed_at IS NULL` count check.
- `handleConnectionCallback` maps `replayed` to a typed `OAUTH_STATE_REPLAYED` (409) DeviceSyncError carrying the sanitized `returnTo`/connect-source context — redelivery never redoes the connection work (token exchange, upsert, hooks).
- The hosted callback route redirects `OAUTH_STATE_REPLAYED` to the validated `returnTo` **without** asserting a `deviceSyncStatus` (destinations render store truth; the earlier delivery may have succeeded or failed), and gives all other no-returnTo errors a `/connect` fallback redirect.
- Additive schema change on both stores: nullable `consumed_at` column (Prisma migration + SQLite ensure-column, mirroring the `owner_id` precedent).

## Invariants to preserve

- Fail-closed stays intact for real attacks: unknown states, expired states, provider mismatch, and owner mismatch all still reject; mismatch checks run before the replay check, so replaying a state against the wrong provider/owner still reports mismatch, not a redirect.
- Replay redirects never assert a connection outcome and never leak account ids; `returnTo` goes through the same protocol/origin validation as every other callback redirect (`javascript:` etc. fall back to `/connect`).
- Redelivered callbacks must not re-run token exchange, connection upsert, or established hooks (asserted in tests via `upsertConnectionCalls`).
- Replay tolerance is bounded by the state's own 15-minute TTL; no new expiry or cleanup mechanism.

- Vercel `buildCommand` runs migrations before build, so the column exists before new code serves. The column is nullable/additive: old Prisma clients name only old columns and are unaffected.
- During the deploy overlap window (or on code rollback), old instances still delete rows on consume; a replay that lands on an old instance in that window behaves exactly as today (400 page). A state marked consumed by new code and then redelivered to an old instance would be re-consumable once (old code ignores `consumed_at`); worst case is a duplicate idempotent upsert during a minutes-long window. No tandem deploy required.
- Local daemon SQLite DBs gain the column via the boot-time ensure-column migration (verified against a legacy-schema DB).

## Verification

- `packages/device-syncd`: full vitest suite 652/652, `pnpm typecheck` clean.
- `apps/web`: `prisma-store-oauth-sessions`, `device-sync-callback-route`, `device-sync-hosted-wake`, `device-sync-connect-complete-page`, `connect-page` suites green; `pnpm typecheck` exit 0; `prisma validate` clean.
- Direct scenario proof: legacy SQLite DB (pre-column) migrated via `ensureDeviceSyncStoreSchema`, then consume → `consumed`, redeliver → `replayed`, post-expiry → `missing`.

---
<a href="https://app.blacksmith.sh/cobuildwithus/codesmith/murph/pr/378"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785685165&installation_id=127572939&pr_number=378&repository=cobuildwithus%2Fmurph&return_to=https%3A%2F%2Fgithub.com%2Fcobuildwithus%2Fmurph%2Fpull%2F378&signature=ba049c579b3a88943d32a425dcfaee2f0b6cee84780f0f9329d2308a57c135b1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


rocketman-21:
**Review round 2 triage**

**Accepted & fixed (06df8a898518):** `apps/web/test/prisma-store-oauth-connection.test.ts` was still on the delete-on-consume protocol and failed. Collapsed its four oauth-state tests into a single delegation smoke (the consume semantics are owned by `prisma-store-oauth-sessions.test.ts`) and moved the owner-mismatch case into that owning suite so the protocol is mocked in exactly one place.

**Rejected: "consumed_at is treated as a completed replay before the callback outcome exists."**

The finding is factually right about ordering (consume is marked before token exchange/upsert) but the proposed fixes add outcome-tracking state without preserving anything the current design loses:

- The replay redirect deliberately asserts **no outcome** — no `deviceSyncStatus`, no error code. The destination pages render connection truth from the store, which is the single source of truth for the user-visible fact. A duplicate delivery *cannot* know the outcome (it may still be in flight); the honest response is "return to the app," and that is what it returns.
- *Concurrent overlap:* both deliveries are navigations from the same browser; whichever commits last wins the viewport regardless of this PR. Pre-PR the losing case rendered a bare 400 "Device connection failed" for a connection that was succeeding (the incident this PR fixes). Post-PR it renders a neutral return. Server-side, exactly one delivery does the work (`upsertConnectionCalls === 1` is asserted; the store-level lost-race → `replayed` path is tested in both stores). Overlap is safe by construction: the shared identity (state row) guarantees no duplication and no loss of the durable fact.
- *Failure-after-consume:* the first delivery still carries the error redirect and surfaces the failure — unchanged. Only the *redelivered duplicate* is neutral, and the app it returns to shows not-connected with the retry surface. Pre-PR that duplicate dead-ended on a bare 400, so nothing that previously surfaced failure has been hidden.
- Persisting a per-state callback outcome (the suggested fix) moves the uncertainty instead of removing it — a crash between the connection upsert and the outcome write would make replays mis-report — and adds a write plus an outcome lifecycle to every callback to improve a transient redirect decoration. Per the repo simplicity rules, rejecting in favor of the existing one-source-of-truth derivation.

rocketman-21:
**Review round 4 triage — rejected (final): "replay handling can hide an in-flight or failed callback as a completed replay."**

This is the round-2 finding re-raised with a sharper frame ("the persisted bit means claimed/in-progress, but transports treat it as terminal"). The frame is accurate about ordering; the rejection stands, on the repo's own invariants:

- **The replay redirect asserts no terminal outcome.** It strips status/error params precisely because a duplicate delivery cannot know the outcome. The only "complete"-flavored surface is the completion page's pre-existing generic title for status-less arrivals, whose strong success claim independently verifies against the connection store before it renders. The connect page renders store truth. Nothing user-visible claims success that hasn't durably happened.
- **The durable user-visible fact is safe by construction.** The stable identity is the state token; the count-checked `consumedAt` mark guarantees exactly one delivery does the connection work (regressions in both stores: lost race → `replayed`; ingress: `upsertConnectionCalls === 1`). The connection is neither duplicated nor lost in any interleaving. Failure surfacing on the delivery that owns the work is unchanged.
- **`docs/contracts/00-invariants.md` § Minimal Mechanism Bias** directs exactly this shape: "start with a stable identity, a uniqueness constraint, and an idempotent write the losing caller can no-op on. Reach for processing-state columns … only when a concrete downstream failure cannot be expressed as a uniqueness or dedupe key." The proposed terminal-outcome column is a processing-state column whose concrete benefit is refining transient redirect decoration in two rare windows (sub-second duplicate race; crash between consume and outcome write — a window the outcome column itself would still have, one write later).
- **Operational visibility instead:** every replay logs `OAUTH_STATE_REPLAYED` via the existing callback-failure warn line. If production shows meaningful volume of replays racing in-flight or failed attempts, that data justifies revisiting with the outcome-derivation design; adding the lifecycle preemptively does not.

No accepted findings this round; head unchanged since `dac558b1d900`.

PR #380, opened 2026-07-03
Service due gated assistant wakes in-attempt; restore the idle linger

rocketman-21:
## Why this PR exists

A hosted message or assistant wake sent seconds after Murph replied could wait ~3.5 minutes: checkpoint-gated assistant work was handed back to orchestration (which parks on a fixed `recommendedRecheckAt` floor) instead of being serviced by the still-warm runtime. This keeps the fix local to the hosted runtime: after foreground work dirties runtime state, wait out the configured idle checkpoint delay, checkpoint once, run durable effects, then service a due assistant wake in-attempt from committed state. Fresh conversation input always preempts.

## Architecture: one `pendingWake`, not a wake lattice

The earlier iteration of this PR multiplexed several logically-distinct pending wakes through a single durable `(nextWakeAt, nextWakeReason)` slot using a lattice of side-flags and string keys (a 7-field projection, `projectedWakeRequiresCheckpoint`, `servicedProjectedRuntimeWakeKey`, `projectedWakeCheckpointGateFresh`, a parallel `durableCheckpointWakeAt` scalar, four foreground-pass wrappers). Nearly every review round found a different interleaving where the wrong wake won the slot or a wake was dropped/replayed across the checkpoint boundary — the accidental complexity of hand-maintaining a priority queue in scalars.

This rework collapses that to a single rule:

> The runtime tracks one in-memory value, **`pendingWake`** — the wake a checkpoint-right-now would persist. Every pre-checkpoint foreground pass runs against the committed workspace with its wake slot overlaid by `pendingWake`, unconditionally. A wake is serviced only from committed post-checkpoint state.

Deleted: the 7-field projection wake machinery, `projectedWakeRequiresCheckpoint` (incl. off the runner result), `servicedProjectedRuntimeWakeKey`, `projectedWakeCheckpointGateFresh`, `durableCheckpointWakeAt` + its reconciliation, the merge wake-lattice, and the four pass-runner wrappers (→ one `runForegroundPass`). The one genuinely irreducible case — a durable-effect wake held behind an earlier due-assistant wake ("two wakes at once") — is an explicit `{dueAssistantWake, durableWake}` holder with value-based reconciliation that can never drop the held wake.

Net production delta: **-146 lines** vs the prior head; 0 lattice concepts remain.

## Invariants preserved

1. Never checkpoint dirty hot state before the configured idle delay (default 180s) elapses since the last user/hot foreground activity; the delay resets on each hot pass. Nothing pulls it earlier; only shutdown does.
2. Fresh user conversation input is highest priority and preempts everything, including an in-flight checkpoint (`foreground_pending` / checkpoint-interrupt). A user message is never trapped behind a checkpoint.
3. A wake is serviced only from committed post-checkpoint state; a due assistant wake is serviced in-attempt after checkpoint; non-assistant due wakes are returned to orchestration.
4. A stale committed assistant wake is never shown to the assistant phase twice before checkpoint.
5. Durable post-checkpoint effects that mint a wake persist it via a follow-up checkpoint before returning.
6. No new persisted state, queues, schedulers, lifecycle owners, or protocol fields.

Runtime-internal; no protocol/route/parser contract change beyond what the prior head already shipped. The only externally-visible change is deleting the now-unused `projectedWakeRequiresCheckpoint` field from the runner result (no external consumers). Deploy order and skew tolerance are unchanged.

## Verification

- `packages/assistant-runtime` full suite: 269 tests green (entrypoint 173 + runner 96), independently re-run.
- Every behavior change from this rework is pinned by a regression test that fails on the pre-fix code.
- Deep review (gpt-5.5-pro) converged over 5 rounds: 3 real issues found → fixed → each guarded (two-wake `replaceWake` durable-wake loss; a reconcile timer override that could checkpoint hot state before the idle delay; the object-identity fragility). 2 speculative Criticals were disproven with tests and kept as permanent guards. Final round: no merge-blocking correctness issues.
- `git diff --check`; changed-file typecheck clean (0 errors in the touched files).


rocketman-21:
ReviewGPT round 1 resolution (thread: 6a47dbff-160c-83ea-8b58-1ccceb1d556a): 1 High finding, **rejected with proof** after independent Codex xhigh adjudication.

Claim: the new `!projectedWakeRequiresCheckpoint` arm can re-service a due assistant wake that already got a pre-checkpoint foreground pass ("requires-checkpoint laundering").

Why rejected:
- For the same normalized wake, `mergeHostedWorkspaceInvocationProjection`'s `preserveCheckpointGatedWake` branch keeps `projectedWakeRequiresCheckpoint=true`, so the exact shape the finding cites (the pinned `checkpoint-gates projected wakes while durable effects remain pending after an external wake` test) is NOT serviceable under the new predicate — that test passes green unchanged.
- The gate clears only when a later pass **replaces** the minted wake with a *different* plain due assistant projection — i.e. that pass's own statement of its remaining due work. Servicing that once, after the checkpoint commits its durable effects, is the incident arm's intended behavior, bounded by `servicedProjectedRuntimeWakeKey` plus the post-checkpoint exact-key/due recheck.
- The proposed "pre-checkpoint wake pass occurred" tracking bit would add lifecycle state for a speculative path already constrained by the existing merge boundary — rejected per the repo's architecture-pressure rule.

Instead of a guard, the disputed reachable shape is now **pinned as intended behavior** by a new test (`services a replaced plain due assistant wake after checkpointed durable effects`, commit f5ffb50cc396): replaced wake serviced exactly once post-effects, second checkpoint clears it, invocation returns non-due. Suite 131/131 + repo typecheck green.

Zero accepted findings remain from round 1. Note: current CI failures on this PR are inherited from main's contact-card commit 9045705499 (apps/web typecheck errors in `murph-contact-card-picker`); this branch's local full-repo typecheck is green.

rocketman-21:
ReviewGPT round 2 resolution: 1 High finding, **accepted and fixed** (commit eb2554b118).

Finding: a due assistant wake minted by a durable checkpoint effect with `requiresFollowUpCheckpoint: true` (production path: `codex-auth.updated` deferred system-mailbox records) was stranded — `forceIdleCheckpointBeforeWake` skipped the serviceability computation, so after the forced follow-up checkpoint the service branch saw a null key and returned a committed due `nextWakeAt` with no prompt orchestrator transport. Pre-existing behavior, but squarely this PR's seam.

Fix (implemented by Codex xhigh, supervised):
- The serviceability/key computation now runs on forced-checkpoint iterations too; only the dirty-window wait is skipped (`budget_exhausted` still skips servicing entirely).
- The follow-up projection marks `projectedWakeCheckpointGateFresh` true only when the selected wake strictly came from the durable effect output (ties with the carried projection stay stale/unserviceable).
- Once a pass services the effect-minted wake, the durable wake accumulator is cleared (exact key match) so the return path reports non-due instead of re-advertising the consumed wake (invariants §44).
- New regression: effect returns due assistant wake + `requiresFollowUpCheckpoint` → first checkpoint → effect → forced follow-up checkpoint → wake serviced in the same invocation → non-due return. One existing test's hardcoded 2026-04-27 effect-wake timestamp had gone stale-due in real time and was moved to an explicit future offset, preserving its original future-wake-returns intent.

Verification: entrypoint suite 132/132, `pnpm test:diff packages/assistant-runtime` + downstream cloudflare verify green, repo typecheck green. Round 3 fired on this head.

rocketman-21:
ReviewGPT round 3 resolution: 1 High finding, **accepted and fixed** (commit e2db357e5a0d).

Finding: the round-2 fix left a tie-case gap in the `requiresFollowUpCheckpoint` branch — when a due assistant wake already in the accumulated projection won or tied the durable effect's follow-up wake, the branch rewrote it to a stale gate, discarding the serviceability computed pre-checkpoint. The forced follow-up checkpoint then returned that due wake instead of servicing it (same cliff, realistic shape: due system-mailbox/pending-input wake + codex-auth.updated deferred record in one pass).

Fix (Codex xhigh, supervised): the follow-up branch now keeps the selected wake serviceable when its key equals the iteration's already-computed `checkpointBlockedProjectedWakeKey` (`followUpCheckpointWakeRemainsServiceable = cameFromDurableEffect || key-held`). The rule is now uniform: serviceability computed before a checkpoint survives that checkpoint for the same surviving wake key. Stale carried outbox/device-sync flavors are unaffected — they never hold a non-null key at that point. No new state.

Regression added: due plain assistant wake + effect returning the same due wake with `requiresFollowUpCheckpoint` → both checkpoints → serviced in the same invocation → non-due return.

Verification: suite 133/133, repo typecheck green. Round 4 fired on this head.

PR #381, opened 2026-07-03
Make Linq reply delivery exact-item based

rocketman-21:
## Why

The old checkpoint-deferred mailbox consume path left accepted Linq replies replayable until runtime checkpoint. ReviewGPT also found that runner-supplied lane high-water and proofless accepted delivery outcomes could either skip the wrong work or record provider success without durable exact consume evidence.

This PR keeps the smallest durable shape: web records the exact accepted mailbox item set on the existing Linq delivery row via child rows, status/reconciliation derive from that same exact evidence, and web no longer directly wakes Cloudflare runtime state. Temporal remains the wake owner.

This is PR 1 of `agent-docs/exec-plans/active/2026-07-03-mailbox-wake-collapse.md`.

## What Ships

**Exact accepted-item suppression.** Accepted Linq delivery rows persist the exact current-inbound mailbox item id and same-reply answered mailbox item ids after web validates the proof against mailbox item payloads, target chat, user, expiry, and routed account authority. Mailbox fetch returns bounded `consumedItems`, so accepted replies behind lower pending gaps restore as context-only instead of replyable.

**Runtime lag uses exact accepted rows.** Status, reconciliation, and quota-notice pending-item selection derive the effective conversation work floor from contiguous exact accepted rows with immutable `acceptedAt` proof, without mutating the stored lane `consumed_seq` and without depending on later mutable provider receipt status.

**Exact reply proof boundary.** Runtime same-target Linq context fallback may still authorize route/provider egress, but it is not answer-consume authority. Outgoing replies with a `replyToMessageId` need matching `currentInbound` proof before provider dispatch; otherwise consuming sends fail before provider acceptance instead of recording the wrong later inbound as answered.

**Proofless accepted outcome hard-cut.** Runtime outcomes say whether exact consume proof is required. Web records proofless accepted outcomes only when they explicitly declare `consumeRequired: false`; missing proof without that non-consuming marker fails closed instead of recording accepted success with no consume authority.

**Prepared retry authority.** Runtime persists the full `currentInbound` proof and answered mailbox item ids on the existing prepared Linq outbox intent authority, next to route/service authority, so post-checkpoint prepared retries restore the same exact consume proof before provider send. Empty answered-item arrays are omitted from persisted outbox JSON.

**Consume authority split from idempotency.** Hosted delivery idempotency inputs stay available for deterministic keys, including synthetic cron/notification occurrence ids, but only `consumeMailboxItemIds` feeds Linq delivery consume authority. Scheduled Linq reminders can record provider acceptance without sending synthetic ids as answered mailbox items.

**Replay, receipt, and deletion repairs.** Accepted replay repair keys off durable `acceptedAt`; exact consumed reads require accepted proof and ignore later parent status flips. Failed-only and skipped receipts do not suppress the inbound item; late failed receipts after accepted proof do not revoke already-validated consume authority; delivered receipts can restore the same accepted row as consume evidence; accepted-answer child rows cascade through mailbox item deletion for account deletion/privacy cleanup.

**Deleted unsafe authority.** Runtime-to-web delivery outcomes no longer use `answeredCoverage`; web ignores legacy callers that still send it, drops delivery-row lane/seq consume columns, and removes the helper that advanced `consumed_seq` from delivery outcomes. The checkpoint-deferred consume-ack path, `/hosted-mailbox/consume` route, optional mailbox-port consume method, dead `isHostedMailboxLaneCheckpointed`, and unsafe direct Cloudflare wake path are gone.

## User-Visible Goal

Prevent duplicate or stale Linq replies by making accepted final reply delivery durably suppress the exact answered mailbox item set, including accepted-behind-gap, already-staged consumed replay, multi-input same-reply, recovered-provider-thread, delivered-receipt replay, prepared-retry, scheduled-reminder, late-provider-failed receipt, same-target fallback, and account-deletion cases, without adding a second wake authority or a second lane high-water owner.

## Invariants To Preserve

- Temporal remains the sole owner of orchestration/wake state; web may signal Temporal but must not directly wake Cloudflare runtime state via OIDC.
- Delivery outcomes can suppress only web-validated exact mailbox items; they must not mutate stored contiguous lane progress.
- Status/reconciliation may derive an effective floor only from contiguous accepted exact rows and must not advance across gaps.
- Accepted-item consume authority is keyed to immutable accepted proof, not mutable provider receipt status; late failed receipts must not revoke already-validated coverage.
- Same-target Linq context fallback is route/egress authority only; answer-consume proof for replies must match the outgoing `replyToMessageId` before provider dispatch.
- Proofless accepted delivery outcomes must explicitly declare `consumeRequired: false`; otherwise web fails closed rather than recording success with no consume authority.
- Hosted delivery idempotency inputs are not consume authority; only `consumeMailboxItemIds` may become Linq answered mailbox item ids.
- Prepared Linq retries must restore persisted current-inbound and answered-item proof before provider send.
- Accepted replays may add missing exact item rows only after web validates current-inbound and answered-item proof against stored mailbox payloads and route authority.
- Failed-only and skipped outcomes must not create exact consumed item facts.
- Progress delivery is not terminal answer evidence and must not create exact consumed item facts.
- No broad managers, queues, or lifecycle machinery were added; the durable shape is one child table attached to existing delivery/mailbox item rows plus one optional prepared-intent proof.

## DEPLOYMENT CONCERNS

Apply the additive web database migration first or with the cutover. Exact-only accepted Linq consume authority is not a web-first compatible protocol change: drain/stop old Cloudflare runners before the new web delivery route handles terminal Linq outcomes, because the new route rejects proofless accepted outcomes unless they explicitly declare `consumeRequired: false`. Then deploy the web route and Cloudflare worker/runner bundle as one hard cutover for Linq delivery outcome recording.

After the new runner has written non-empty Linq outbox intents with exact consume proof, do not roll back to a runner build that predates that schema unless pending outbox is drained or scrubbed first. Post-deploy checks: accepted Linq delivery stores answered mailbox item child rows; mailbox fetch returns `consumedItems` for accepted-proof exact rows even after a late failed receipt; failed-only rows do not suppress; status/reconciliation clear lag for contiguous exact accepted answers; proofless non-consuming reminders record acceptance; proofless consuming/old-runner accepted outcomes fail closed; mismatched current-inbound proof or route authority stores no item rows; same-target fallback cannot become answer-consume proof for a different `replyToMessageId`; progress Linq delivery stores no item rows; Cloudflare rejects OIDC direct ensure; Temporal signed ensure still wakes processing.


rocketman-21:
Closing in favor of a minimal rewrite tracked under agent-docs/exec-plans/active/2026-07-03-mailbox-wake-collapse.md.

This branch drifted over ~50 review rounds into a Linq-specific consume subsystem (join table + `currentInbound` proof + `consumeRequired` + derived-floor window SQL + replay/receipt repair), and the direct-wake latency win it was built to enable was reverted mid-PR, so it now costs a lot and ships no latency benefit.

Replacement design: one composable change, a nullable `hosted_mailbox_item.consumed_at` stamped at reply-accept and read at fetch to restage answered messages context-only (channel-agnostic, per-item so no cross-channel over-claim), plus the re-landed direct Linq wake for the latency benefit, plus the planned deletions (checkpoint consume-ack block, /hosted-mailbox/consume route, mailbox-port consume method, dead isHostedMailboxLaneCheckpointed) and the PR 2 patch-retirement + wake-field collapse.

Branch kept (not deleted) so the clean deletions and the direct-wake code can be cherry-picked.

PR #389, opened 2026-07-06
fix(assistant-engine): suppress summaries for stale backfilled device activity

rocketman-21:
## Problem

Device-activity automations (the morning sleep summary, workout check-ins) fire on **every** sleep/workout record as it arrives. When a WHOOP-via-Junction backfill lands after the device was offline for a day or more, the user gets pinged about days-old activity they don't care about. Reported case: 48h offline, then two sleep summaries on reconnect (the previous night plus last night).

## Fix

One uniform rule for the whole category: drop any device-activity record whose activity **ended more than 24h before the sync** before it becomes a candidate, so the coverage cursor never advances onto it and it stays suppressed permanently (not just this pass).

- Anchored on `endAt ?? occurredAt`: sleep uses its real `endAt` (it runs for hours, so its start can look stale while it's genuinely last night); workouts have no `endAt` and are short, so their start time is a fine anchor.
- **Fails open**: a missing or unparseable timestamp keeps the notification. We'd rather occasionally send than wrongly suppress a real summary.
- 24h window is generous enough that last night's sleep and this morning's workout always pass even if the phone syncs late in the day, while anything from an earlier day is dropped.

## Scope

Fixes the stale/backfill case only. The "woke for an hour and went back to bed" fragment (two fresh records same morning) is intentionally left alone, since that needs a settle/finality mechanism we chose not to build.

## Tests

Two staleness tests (sleep + workout), each paired against an existing fresh-scheduling test with the same selectors, so the only delta is the old timestamp producing zero scheduled jobs. The workout test uses a fresh `recordedAt` but old `occurredAt` to prove the guard measures the event's real time, not import lag. 21/21 pass, typecheck clean.

---
<a href="https://app.blacksmith.sh/cobuildwithus/codesmith/murph/pr/389"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785943711&installation_id=127572939&pr_number=389&repository=cobuildwithus%2Fmurph&return_to=https%3A%2F%2Fgithub.com%2Fcobuildwithus%2Fmurph%2Fpull%2F389&signature=41fbc925255ba360729c1be10d2af97a4caac25d690f597d79f8f8698c53229a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


rocketman-21:
Closing in favor of a clean replacement PR. This branch (`fix/suppress-stale-device-activity-summaries`) collided with another agent's work in a shared checkout and picked up unrelated commits (linq receipts, contracts docs, workflow docs) via a merge. The device-activity staleness change now lives on its own uniquely-named branch.

PR #397, opened 2026-07-06
perf(hosted-runner): defer conversation connectors off the cold boot path

rocketman-21:
## Why

Cold-start `nodeStartupMs` on the hosted runner runs 1.8–4.2s. That metric spans **process start → HTTP `listen`**, so every module evaluated at import time is paid before the container can accept a request. Prod bundle inspection showed the entry chunk was **3.43 MB** and carried **69 `grammy`/`node-fetch` references** — the whole hosted-conversation connector graph (`grammy → node-fetch → whatwg-url`) was being evaluated at boot.

The runner never calls these provider clients directly (message egress goes through the Worker), so they were dead weight on the boot path — pulled in only via the `@murphai/inboxd` barrel and a Telegram driver value-import that was really only used for types.

## What changed

- **Lazy-load `events/conversation.ts`** behind a cached dynamic `import()` in `mailbox-conversation-import.ts`, so the conversation connector runtime loads on the first conversation import rather than at boot.
- **Narrow `@murphai/inboxd` root imports** to the `@murphai/inboxd/runtime` subpath (idle-maintenance, pending-input-index, events/conversation) and make the **Telegram attachment driver an `import type`** — dropping the static `grammy` edge entirely.
- **Add a bundle-assembly guard** that fails the build if provider-connector inputs re-enter the *static* boot closure. It traverses static imports only (skips `dynamic-import` edges), so the lazy chunk is correctly allowed.

## Result

| | Before | After |
|---|---:|---:|
| Entry chunk (`container-entrypoint.js`) | 3.43 MB | **2.29 MB** (−33%) |
| `grammy`/`node-fetch` refs in entry chunk | 69 | **0** |
| Total bundle | 9.11 MB | 8.20 MB |

Entry budget tightened to 2.9 MB; the total ceiling is **held at its prior 9.3 MB** (this shrinks the bundle, so the regression guard is not loosened).

## Verification

- `pnpm --dir apps/cloudflare runner:bundle` — passes (guard + budgets).
- Runner-bundle Vitest suite (incl. new static-closure guard + lazy-allowed tests) — green.
- Focused assistant-runtime tests (conversation-event, idle-maintenance) — green.
- Typechecks: assistant-runtime, inboxd, cloudflare, root — all pass. `verify:package-boundary:prepared` — pass.

Behavior is unchanged: the conversation import path is exercised by the assistant-runtime tests and goes through the new cached lazy loader; grammy is fully tree-shaken because the runner never used it at runtime.

## Deploy

Hosted-runner-only; no web/protocol tandem deploy. Existing warm containers keep the old bundle until normal Cloudflare container rollout replaces them, so the cold-start win appears as new containers take over. Use `container_rollout=immediate` if the win is wanted right away.

---
<a href="https://app.blacksmith.sh/cobuildwithus/codesmith/murph/pr/397"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785951217&installation_id=127572939&pr_number=397&repository=cobuildwithus%2Fmurph&return_to=https%3A%2F%2Fgithub.com%2Fcobuildwithus%2Fmurph%2Fpull%2F397&signature=be197dff8aadab1c06519daf2f87bcf915ba161140ab5f7306beafa028cfcb1b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


rocketman-21:

Ran the PR deep-review loop (local Codex CLI `gpt-5.5`, xhigh, review-only — ReviewGPT is retired for this loop as of 2026-07-06) against the pushed head to zero accepted findings.

- **Round 1** (head `6c65465`): 1 finding received, **1 accepted** — the lazy `loadHostedConversationEventsModule` cached a *rejected* import promise (`??=`), which would poison all later conversation imports until restart. Fixed with a retryable memoized-promise pattern (clears cache on rejection, guarded against clobbering a concurrent retry; success stays memoized) + a focused regression test (`hosted-runtime-mailbox-conversation-loader.test.ts`, verified failing pre-fix).
- **Round 2** (head `9177d1b`): **zero findings** — reviewer verified the fix is correct under concurrent interleavings and the test is non-tautological. Loop terminated.

### CI
PR-relevant checks are green: Build hosted runner bundle ✓, Release package coverage (assistant) ✓, Release build/typecheck ✓, Telegram + idle checkpoint E2E ✓, all Linq E2E ✓.

Three red checks — `Release package coverage (cli)`, `Junction wearable direct-resource replay E2E`, and the `Release checks` rollup — are **pre-existing failures on `origin/main`** (verified red on main HEAD `4b656d9`); this PR touches no cli/junction/wearable files, so they are not introduced here.

PR #404, opened 2026-07-06
perf(hosted): instrument + cut the web->Cloudflare direct-wake control hop

rocketman-21:
## Why

The direct "ensure wake" the web webhook fires at the Cloudflare worker after a Linq inbound takes a consistent **~580–820ms** (p50 ~750ms) from Temporal-signal-accepted to CF-route-received — the single biggest addressable warm-path latency between a user's message and Murph starting to type. That figure is a cross-clock span and cannot be attributed from code alone: candidates are web-side OIDC token acquisition, cold connection setup, and CF-side OIDC/JWKS verification (which runs before the existing route timestamp because the route authorizes before dispatch).

## What

One PR that both attempts the win and measures which sub-phase remains:

**Instrumentation (same-clock spans only):**
- New optional orchestration leaves: `tokenAcquireStartedAtEpochMs` / `tokenAcquiredAtEpochMs`, `directEnsureRequestStartedAtEpochMs` / `directEnsureResponseReceivedAtEpochMs` (web clock), `runtimeControlAuthStartedAtEpochMs` / `runtimeControlAuthFinishedAtEpochMs` (CF clock).
- The shared control client stamps token/RTT timings only when the caller passes an `onTiming` callback; web threads them to CF via internal headers; CF records auth start/finish around the existing route authorization for `runtime-ensure-processing` only.
- The webhook wake path persists the web-clock RTT breakdown best-effort via `.finally()` — never on the provider-response path. Cross-clock epochs are stored only as raw leaves; the web-clock request→response RTT is the skew-free ground truth for the hop.

**The two likely fixes:**
- **Module-scope, expiry-aware Vercel OIDC token cache** (60s refresh margin) in the web auth adapter: concurrent refreshes dedupe on one in-flight promise (rejections are not cached), unparseable-expiry tokens fall back to per-call, an expired token can never be served. Same token source and audience — no auth semantics change. Vercel OIDC is deployment-scoped, so cross-request reuse in a warm instance is legitimate.
- **Keep-alive connection reuse** via a web-only undici `Agent` injected through the client's existing `fetchImpl` seam. The shared `cloudflare-hosted-control` package stays transport-agnostic and worker-safe (it is imported by the CF worker itself; undici is an `apps/web` dependency only).

## Invariants preserved

- The direct wake stays best-effort fire-and-forget; nothing here can throw into or block the webhook response. Temporal remains the durable at-least-once backstop, untouched.
- All new leaves are optional/additive: web and Cloudflare can deploy in either order (missing leaf = undefined, never an error). Tandem deploy recommended to get the full split immediately.
- Auth trust boundaries unchanged (same audience/verification); only still-valid credentials and connections are reused.

## Verification

- Typechecks green: hosted-execution, cloudflare-hosted-control, apps/cloudflare, apps/web.
- Tests green: runtime-control parser (23), control client (29), CF routes (86), web token/control/webhook/latency-store focused (28) — including negative parser cases for the new leaves and token-cache expiry/refresh/dedup unit tests.
- `deps:guard`, `deps:ignored-builds`, `git diff --check` pass. New web `undici` pinned to patched 7.28.0.

## After deploy

The next prod trace splits the ~750ms into token-acquire (web clock), request RTT (web clock), and CF auth verification (CF clock). If the token cache + keep-alive landed the win, RTT drops directly; whatever remains is attributed for the follow-up.

---
<a href="https://app.blacksmith.sh/cobuildwithus/codesmith/murph/pr/404"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785961649&installation_id=127572939&pr_number=404&repository=cobuildwithus%2Fmurph&return_to=https%3A%2F%2Fgithub.com%2Fcobuildwithus%2Fmurph%2Fpull%2F404&signature=bc44885a2923c465f6fc4c825ef72faf5adfcf77ae84c5ab46d29467dd48ca3c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/agent-docs/exec-plans/completed/2026-07-06-control-hop-latency.md
+++ b/agent-docs/exec-plans/completed/2026-07-06-control-hop-latency.md
@@ -1,0 +1,59 @@
+Goal (incl. success criteria):
+- Split the measured web-to-Cloudflare direct ensure wake latency into same-clock sub-timings.
+- Ship the two low-complexity warm-path optimizations: web-side process OIDC token reuse and web-only keep-alive connection reuse.
+- Success means new optional orchestration leaves round-trip through shared contracts, Cloudflare records auth timing, web records request/token timing, tests cover parser/persistence/cache behavior, and the direct wake remains best-effort.
+
+Constraints/Assumptions:
+- Keep Cloudflare-hosted-control transport-agnostic; no Node-only imports in packages used by the Worker bundle.
+- New timing leaves are additive and optional for web/Cloudflare deploy skew.
+- Compute only same-clock spans downstream; store cross-clock epochs only as raw leaves.
+- Token reuse must preserve audience/auth semantics and never serve expired or near-expired tokens.
+- No runner bundle rebuild.
+
+Key decisions:
+- Inject a web-local undici Agent through the existing fetch implementation seam rather than changing the shared client.
+- Use module-scoped web OIDC token cache keyed by audience/config identity with expiry safety margin.
+- Thread web timing leaves as internal headers on the direct ensure request; Cloudflare merges them into orchestration diagnostics with auth start/finish leaves.
+
+State:
+- Implementation and scoped verification complete; ready to close with a scoped commit.
+
+Done:
+- Required routing, architecture, security, reliability, and verification docs read.
+- Worktree confirmed clean on the requested branch.
+- Added optional orchestration timing leaves across hosted-execution shape, parser, leaf key list, Cloudflare response assembly, and web latency-store persistence.
+- Added direct ensure client timing callback/header emission without adding Node-only imports to the shared Cloudflare control package.
+- Added web-only keep-alive fetch dispatcher injection in `apps/web`.
+- Added module-scoped expiry-aware Vercel OIDC token cache in `apps/web`.
+- Added/extended focused tests for token cache, runtime-control parser round-trip, hosted latency persistence, Cloudflare ensure route diagnostics, and shared control client headers/callback.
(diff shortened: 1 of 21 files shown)


rocketman-21:

Round 1 (head `95dfa5961d`, Codex `gpt-5.5` xhigh): **zero accepted findings**.

One Low was reported and **rejected with reasons**: "provider returns (but refuses to cache) a token already inside the 60s expiry margin" — this is identical to pre-PR per-call behavior (not a regression), failing closed would strictly worsen the best-effort fast path while Temporal remains the durable backstop, and re-minting would return the same near-expiry token. All other angles clean: rejected refresh promises evicted, undici confined to apps/web (shared client stays worker-safe), timing headers scoped to the ensure route, CF auth timing additive/route-scoped, hostile header values parse defensively, latency store merges rather than clobbers runner-written phase-breakdown fields.

PR #413, opened 2026-07-06
feat(hosted-ops): founder-gated Linq home-line rehome primitive

rocketman-21:
## Why

We now run two configured Linq lines and need to migrate a member (starting with the founder) from one line to another — and, longer term, fold members off a degraded or banned line automatically. The prod DB role is read-only and routing phone columns are encrypted app-side, so line migration must be a deployed, ops-gated action. This PR ships the minimal reusable rehome primitive that a future health-driven failover loop can call.

## User-visible goal

A founder-gated ops action (GET/POST `/api/ops/linq-rehome`) that moves one hosted member's Linq home line onto an explicitly selected assignable line. After rehome the member's next text to the new line binds the new home chat; texting the old line gets the existing redirect-to-home reply. Murph-initiated sends stay fail-closed until the member texts the new line (last-inbound intentionally resets), preserving the anti-cold-contact posture.

## Invariants preserved

- Rehome target must be in the assignable pool (configured, enabled, healthy) and under active-member / daily-assignment capacity — checked via the existing `chooseHostedLinqHomeLine` under the existing assignment advisory lock.
- Single write path: `upsertHostedMemberHomeLinqRecipientPhoneTx({ clearPending: true })`; no new routing write logic, no webhook/egress/routing-policy changes, no schema change.
- 28-day recipient-reply egress guard is not weakened: last-inbound is not carried across lines.
- Ops responses expose phone hints and lookup keys only, never decrypted numbers.
- Known accepted residual (documented in the plan doc): stale `hosted_thread_route` rows for the old 1:1 chat may let an in-flight session egress on the old thread until sessions roll over; thread-route cleanup belongs to future banned-line failover work.

---
<a href="https://app.blacksmith.sh/cobuildwithus/codesmith/murph/pr/413"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785965976&installation_id=127572939&pr_number=413&repository=cobuildwithus%2Fmurph&return_to=https%3A%2F%2Fgithub.com%2Fcobuildwithus%2Fmurph%2Fpull%2F413&signature=142467319df0e2dd422317327c0892af5067902c271075c861a21d3d972b7758"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/agent-docs/exec-plans/completed/2026-07-06-linq-line-rehome.md
+++ b/agent-docs/exec-plans/completed/2026-07-06-linq-line-rehome.md
@@ -1,0 +1,86 @@
+# Ops primitive to rehome hosted member Linq home line
+
+Status: completed
+Created: 2026-07-06
+Updated: 2026-07-06
+
+## Goal
+
+- Ship a founder-gated hosted ops primitive that can move one active hosted member's Linq home-line authority to an explicitly selected assignable home line.
+- Preserve the existing bare-authority semantics: clear bound chat and pending route state, reset `linqLastInboundAt`, and require the member to text the new line before Murph-initiated sends resume.
+
+## Success criteria
+
+- `GET /api/ops/linq-rehome?memberId=...` returns only member routing summary, line hints, lookup keys, and capacity counts; it never returns decrypted phone numbers.
+- `POST /api/ops/linq-rehome` rejects unauthorized operators, unknown/suspended members, non-assignable targets, already-targeted members, and capacity-exhausted targets with typed hosted onboarding errors.
+- Successful rehome is one database transaction guarded by the existing Linq home assignment lock and writes through `upsertHostedMemberHomeLinqRecipientPhoneTx` with `clearPending: true`.
+- Tests cover the requested success, pending-clear, none-authority, target rejection, already-target, capacity, member validation, auth/body validation, and raw-number exposure cases.
+
+## Scope
+
+- In scope:
+  - `apps/web/src/lib/hosted-ops/linq-line-rehome.ts`
+  - `apps/web/app/api/ops/linq-rehome/route.ts`
+  - Focused hosted-web tests for the new primitive and route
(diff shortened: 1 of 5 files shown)


rocketman-21:
Codex deep-review loop summary (gate for merge):

- Round 1 (head 3357e2509d): 3 findings, all addressed in 6ddada6449 — phone-identity precondition (HOSTED_LINQ_REHOME_MEMBER_PHONE_REQUIRED), lookup-key-drift-safe already-on-target guard, metadata-only operator log (durable audit table rejected: new state owner for one founder-gated action fails the architecture pressure check).
- Round 2 (head 6ddada6449): 1 finding, rejected. R2-F-01 claimed pending-route freshness on the target line surviving rehome weakens the 28-day guard. Rejected because the promoted `pendingLinqLastInboundAt` records genuine recent member inbound on the *target* line itself, so the anti-cold-contact invariant (recipient replied to this line within the window) holds truthfully; cross-line freshness carrying still never occurs (promotion requires the pending lookup key to match the target line). Suppressing promotion would add a caller-specific mode to the shared write primitive to enforce a stricter-than-invariant reading of the plan doc's common-case wording.

Zero accepted findings at head 6ddada6449.

PR #417, opened 2026-07-07
perf(hosted): ack direct ensure before runtime wake + widen keep-alive window

rocketman-21:
## Why

Post-#404 instrumentation (15 prod traces) split the direct-ensure hop into two costs with different owners:
- **Response leg 0.1–10s**: the ensure route awaits the user-runner DO before responding. Idle DOs are evicted after 70–140s (CF docs), so a post-lull ensure pays DO reconstruction + container wake while the fire-and-forget caller holds the connection. Worse: observed 10.4s RTTs flirt with the web client's AbortSignal.timeout, and per CF docs a client disconnect can **cancel in-flight work not held by `ctx.waitUntil`** — i.e. a slow cold ensure's DO wake can be killed mid-flight today.
- **Arrival leg**: 12–255ms on a reused connection vs ~600–850ms expired; reuse tracked the 60s undici window exactly, and real conversation gaps (76s, 3m47s) missed it.

## What

1. **Ack-early on `runtime-ensure-processing`**, scoped by the existing auth-kind discriminator (no new headers/flags/state): `vercel-oidc` callers (the web fire-and-forget direct wake — verified the only production caller on this credential, ignores the body) get `202 {accepted:true}` after auth+parse, with the identical DO ensure running in `ctx.waitUntil` (30s budget vs 10s observed worst case; rejections logged via the existing structured-log helper). `web-callback-signature` callers (the Temporal worker activity, which parses the synchronous result; the hosted-local harness likewise) keep today's behavior byte-identical.
2. **Keep-alive window 60s→300s** (max 600s): converts the ~700ms expired-connection arrivals for lull-length gaps; stays under Cloudflare's ~400s edge idle timeout with margin. Stale-socket reuse races remain tolerated by best-effort + Temporal backstop.

## Invariants

- Direct wake stays best-effort; Temporal backstop untouched (its activity path never sees the ack).
- Deploy order free: request shape unchanged; new web client accepts both the old full-result and the new ack; old web client during skew logs a benign parse warn while the wake still executes (waitUntil runs regardless).
- The waitUntil path makes the DO wake **survive client timeouts** — strictly fewer lost direct wakes than today.

## Verification

- Typechecks: apps/cloudflare, apps/web, cloudflare-hosted-control — green.
- CF route tests (87, incl. new: oidc→202-ack with DO still invoked via captured waitUntil; callback-signature unchanged; waitUntil rejection logged), client tests (30, incl. ack + old-shape + timing), webhook wake tests (10), hosted-local wake helper tests (4) — all green.

---
<a href="https://app.blacksmith.sh/cobuildwithus/codesmith/murph/pr/417"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785976226&installation_id=127572939&pr_number=417&repository=cobuildwithus%2Fmurph&return_to=https%3A%2F%2Fgithub.com%2Fcobuildwithus%2Fmurph%2Fpull%2F417&signature=f21139158a54914310f18cb6795d9357cd0768c610677d45fac4003487361c4c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/agent-docs/exec-plans/completed/2026-07-06-hosted-ensure-ack-early.md
+++ b/agent-docs/exec-plans/completed/2026-07-06-hosted-ensure-ack-early.md
@@ -1,0 +1,104 @@
+# Hosted Ensure Ack-Early
+
+## Goal
+
+Make the web direct `ensure-processing` wake acknowledge immediately after auth
+and request parsing, while keeping Temporal callback-signed ensure requests
+synchronous and response-compatible.
+
+Success criteria:
+
+- Vercel OIDC direct wake receives `202 { accepted: true }`.
+- The existing Durable Object `ensureRuntimeProcessingForUser` call still runs
+  identically for the direct wake, but under Worker `ctx.waitUntil`.
+- Callback-signed Temporal ensure requests return the existing full result
+  synchronously.
+- The hosted-control client tolerates both old full-result responses and the new
+  direct-wake ack, and still emits timing callbacks.
+- Web direct-control keep-alive windows are widened to cover the observed
+  1-5 minute conversation gaps without adding new state or routing.
+
+## Context
+
+Measured post-PR-404 traces showed reused direct-ensure connection arrivals at
+12-255 ms versus roughly 600-850 ms after the current 60 second undici idle
+window expires. The long tail was on the response leg because the Worker route
+awaited Durable Object wake/reconstruction before returning even though the web
+caller is fire-and-forget and ignores the response body.
+
+Cloudflare Worker `ctx.waitUntil` is the minimal platform primitive for this:
+it keeps post-response work alive after the response or client disconnect, with
+the documented post-response extension window covering the observed worst case.
+
+## Constraints
+
+- No new headers, flags, environment variables, routes, durable state, or
+  production harness bypasses.
+- Dispatch only on the existing presented auth kind:
+  `vercel-oidc` versus `web-callback-signature`.
+- Do not touch `ensureRuntimeProcessingForUser` or its diagnostics payload.
+- WaitUntil-path failures must use the existing structured logging helper and
+  must not become unhandled promise rejections.
+- Temporal backstop behavior and callback-signed response shape stay unchanged.
+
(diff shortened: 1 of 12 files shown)


rocketman-21:

Round 1 (head `6b66d4c93b`, Codex `gpt-5.5` xhigh): **zero accepted findings** — loop terminated at the stop condition.

One Low **deferred with reasons** (test-only, explicitly not production-faithful): `test/workers/worker-entry.ts:265` drops the Worker `ctx` when delegating, so a *future* `SELF.fetch` vitest-pool-workers test hitting the OIDC ensure route would reproduce a missing-ctx failure. Production (`worker/index.ts:26`) and hosted-local (`hosted-local-test-index.ts:37`) entrypoints forward `ctx` correctly, so nothing ships broken. Fix belongs alongside whichever future test first exercises that wrapper path.

Verified clean by the round: auth discrimination derives from the verified credential class (signature dispatch never falls through to OIDC); the 202 ack sits after auth + bound-user check + body limit/parse + timeout-header validation; both paths share one DO-call helper (identical payload); client rejects non-2xx before ack parsing; old-web/new-CF skew stays inside the best-effort catch; `waitUntil`'s 30s post-response cap is strictly no worse than today's client-timeout cancelation; CF docs confirm the 400s client keep-alive window matching the 300s undici sizing.

CI: 18/18 green on the reviewed head. Merge-ready.

PR #423, opened 2026-07-07
perf(hosted-runtime): notify the active turn as soon as a conversation input is staged

rocketman-21:
## Why

When a hosted user's message arrives while their runner is mid-turn, the foreground mailbox import stages the assistant input quickly, but the running turn was only notified after the entire import call returned — including best-effort local-inbox projection (sidecar init, sync SQLite, canonical locks, parser drain, attachment download), measured at ~19s in prod on 2026-07-06. In the traced incident the input was staged 2.6s after the turn began, the notify landed 1.3s before the turn ended, the fold-in was missed, and the user waited 77s for a reply. This violates User Reply Primacy: projection is enrichment and must not sit between an inbound user message and the reply path.

## User-visible goal

A message that arrives while Murph is mid-reply gets folded into the running turn (or answered immediately after it) instead of waiting out projection plus a full extra pass — cutting worst-case mid-turn reply latency by tens of seconds.

## What changed

- New shared `notifyAssistantActiveTurnInputAvailableForInputIds` in assistant-engine's active-turn-input-controller: one owner for per-input conversation-ref derivation, dedupe, and best-effort notify.
- `importHostedConversationMailboxItem` calls it immediately after non-replay staging, before projection (signal passed through; failures swallowed best-effort).
- workspace-runner's post-import notify collapsed to a thin delegation to the same helper (net deletion); it stays in place as the idempotent catch-all.
- Regression tests: notify-fires-before-projection ordering against a real registered active-turn controller with production Linq conversation-key derivation; durably-consumed replays do not early-notify; a throwing notify leaves the import outcome unchanged.

## Invariants to preserve

- User Reply Primacy and Foreground mailbox priority (docs/contracts/00-invariants.md): staging, not projection, gates turn admission; projection stays inline and unchanged (parser-retry blocked outcome, projection status recording untouched).
- Consume authority: admission and `answeredMailboxItemIds` derive from the staged input event and its hosted mailbox item mapping; delivery-time `consumedAt` plus durably-consumed replay skip make the early notify safe even if the import later reports parser-retry.
- Early notify is strictly best-effort: it must never fail, delay, or change a mailbox import outcome.

## Verification

- 54 conversation-import tests, 241 workspace entrypoint/runner tests, 113 active-turn journal + local-service runtime tests, assistant-runtime + assistant-engine typechecks — all green.
- Codex deep-review (xhigh): zero Critical/High/Medium, verdict LAND; one Low (extra end-to-end consume-authority regression) documented in the archived plan as considered-not-actioned.

DEPLOYMENT CONCERNS: none for web/Cloudflare API compatibility — runner-bundle behavior only. Prod effect requires a Cloudflare hosted runner deploy; warm containers keep old notify timing until replaced. Post-deploy check: staged→fold-in gap on `hosted_ingress_latency_trace` for mid-turn arrivals.

---
<a href="https://app.blacksmith.sh/cobuildwithus/codesmith/murph/pr/423"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785979548&installation_id=127572939&pr_number=423&repository=cobuildwithus%2Fmurph&return_to=https%3A%2F%2Fgithub.com%2Fcobuildwithus%2Fmurph%2Fpull%2F423&signature=3cff345fb9040ec15c2c835466544e6f9a3926bcae98c238e3e3c428eaed2643"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


rocketman-21:
Note for the record: this branch was cut from local main at f1835136b8 ("Isolate foreground conversation mailbox imports" — the foreground system-failure isolation lane's completion commit), which had not been pushed to origin/main. This PR therefore carries that commit plus the staged-input notify change, and has now been updated with a merge from origin/main (clean merge; PR #417 touched the DO/entrypoint side, no file overlap; 295 tests re-run green on the merged head). Landing with a merge commit rather than a squash so f1835136b8 keeps its identity on main.

PR #585, opened 2026-07-12
Enable affirmative group reaction replies

rocketman-21:
## Why

PR #560 makes Linq group reactions durable, bounded context, but intentionally
does not wake Murph. That is correct for ordinary reactions; it leaves one
conversational gap when a member likes Murph's own exact question or offer as
their answer.

## User-visible goal

If Murph says, “I have your supplements in checkout. Want me to order?”, a
member's supported affirmative reaction to that exact message should mean yes:
continue the offered action and do not ask for the same confirmation again. A
like on an ordinary statement must not make Murph act or reply by itself.

## What changes

- Reuse the existing `conversation.message` mailbox lane for supported
  affirmative adds to a canonically Murph-authored target, carrying the exact
  reacted-to content and source reference into the assistant turn.
- Keep removals, participant-targeted reactions, and non-affirmative reactions
  on #560's silent deferred-context path. The privacy-safe hashed provider-event
  key remains the mailbox dedupe boundary.
- Re-hand off an already admitted actionable duplicate without reappending it
  or trusting a stale mailbox checkpoint.
- Resolve the active route, current non-self reactor, exact target/chat/part,
  and Murph authorship before either the join owner or generic staging can act.
- Keep the existing join-offer row as the sole owner of react-to-join authority.
  Its effect derives from accepted input, normalized offer intent, and the
  authorized conversation scope, so recovery reuses one provider idempotency
  key even when Codex emits a new tool call. A later accepted input remains
  distinct even when its rendered text is identical.
- Carry that opaque effect in the callback-signed query while leaving the JSON
  body compatible with old web. New web requires it before group or offer
  mutation after the bounded rollout gate is activated.
- Keep group/link mutation and effect preparation in one existing transaction.
  Exact target binding uses a bounded two-row read, compare-and-set, and exact
  ownership reread; ambiguity retries instead of guessing.
- Treat every provider dispatch that may have started without a durable bound
  outcome as retryable through the same offer and provider key. This includes
  network/timeouts, a successful response without a usable message ID, and an
  unreadable bind result. Only a proven nonretryable rejection revokes the exact
  pending offer.
- Delete the rendered-content fallback, canonical-offer alias machinery, and
  join-URL ownership heuristic. No new table, queue, scheduler, or lifecycle
  manager is added.

## Invariants

- Signed ingress, active route/member access, canonical participant identity,
  and exact target/chat/part agreement remain mandatory. Murph authorship is
  additionally required for the wakeable affirmative exception and join-offer
  handling.
- Reaction intent is limited to the exact yes/no question or offered action. It
  cannot bypass authorization, payment, consent, confirmation, or other
  irreversible-effect safeguards.
- Reacted-to content stays bounded inside the encrypted mailbox. Temporal and
  direct-ensure state remain pointer-only.
- Provider event ID remains the durable reaction dedupe boundary through its
  privacy-safe lookup key. Every react-to-join target remains exclusively owned
  by the join flow, including pending, rejected, revoked, or ineligible outcomes.
- Removals stay context-only and do not independently wake or cancel an action.
- The PR does not claim provider-native outbound reply threading; the pinned
  Linq send contract does not serialize a native reply anchor.

## Verification

- Final explicit 47-file `pnpm test:diff` passed against the current stacked
  base: all 10 affected package typecheck/test lanes, web and Cloudflare
  verification, package boundaries, lint, dev smoke, and the Next production
  build.
- Web: 382 files passed, 1 skipped; 4,371 tests passed, 135 skipped.
- Cloudflare: 96 files and 1,736 tests passed.
- After the final clean base-only update: reaction context 23 tests, assistant
  automation 150 tests, and hosted runtime 361 tests passed.
- `git diff --check`, added-line privacy scans, current-base ancestry, and
  mergeability checks passed.
- Fresh coverage-write and security/privacy audits are clear. Coverage added
  exact proofs for ambiguous pending ownership, bind-vs-revoke, and staging
  during the provider-read window. Security/privacy found zero Medium-or-higher
  findings and corrected the shared Linq retry classifier to match the pinned
  SDK's `x-should-retry`, 408, 409, 429, and 5xx behavior.

## ReviewGPT disposition

- Invalid preflight attempts (unknown model confirmation, missing ZIP, browser
  failure, and one Round 10 network error) were discarded and did not count or
  cause code changes.
- Round 1 found removal/cancellation ambiguity and incomplete durable join-offer
  ownership. Cancellation machinery was rejected as outside the requested
  like-means-yes scope. Durable join targets were fixed to remain exclusively
  owned after rejection, revocation, or ineligibility. The fix added only a
  narrow owner-bound result/read and no cancellation store or workflow.
- Round 2 repeated withdrawal, identified an unsupported native-anchor claim,
  and found provider-send/database-bind ownership gaps. Withdrawal was rejected;
  the overclaim was removed and ownership gap accepted. A narrow recovery
  fallback/cleanup path was added, then replaced by Round 3; no SDK threading
  layer was added.
- Round 3 found URL-derived ownership could orphan lost-ack offers and capture
  ordinary same-link messages. Accepted: delete the heuristic and use the
  existing offer owner with exact thread/full-digest correlation. This added
  necessary pending-owner persistence, but no scheduler or second workflow;
  later rounds simplified it further.
- Round 4 found admission-before-authority, rollout-shape, long-digest, and
  sender/reaction races. All were accepted using canonical admission, one
  full-text digest, rollout compatibility, and compare-and-set ownership.
  Temporary fallback/alias complexity was removed in later rounds.
- Round 5 repeated intentional removal behavior and found identical pending
  owners could be ambiguous. Removal withdrawal was rejected; ownership was
  accepted. The initial alias-based fix added schema complexity, later deleted
  in Round 8.
- Round 6 was a concurrent/stale-head pass that repeated the intentional removal
  behavior and an ownership case already being corrected. No independent change
  or complexity was added.
- Round 7 found group/link authority could commit before a changed or revoked
  effect rejected. Accepted: authority mutation and effect preparation now share
  one existing transaction. This was ordering-only.
- Round 8's valid retry found tool-call-derived effects could duplicate after
  response loss while the legacy content fallback could suppress a later valid
  identical offer. Accepted: derive effects from accepted-input identity and
  normalized intent, carry them in the signed query, and delete fallback/alias
  state. Complexity decreased. Coverage also added bounded `take: 2` ambiguity
  detection without new state.
- Round 9 found four issues: deploy skew, pending bind incorrectly acknowledged
  as sent, template-whitespace identity drift, and stale security/reliability
  docs. All were accepted. The fix adds one temporary activation flag and one
  retryable branch, centralizes template canonicalization, and updates docs. Only
  the rollout flag adds temporary operational complexity.
- Round 10's valid retry found post-dispatch ambiguity was returned as success,
  potentially stranding the pending owner. Accepted: ambiguous send/bind results
  now raise one retryable 503 with the same offer/key; only a proven nonretryable
  rejection revokes the exact pending offer. No schema, queue, scheduler, or
  manager was added. The follow-up security audit aligned the shared Linq retry
  classifier without adding state.
- Ten valid rounds reached the repository hard cap. Because Round 10 contained
  an accepted finding, policy forbids an eleventh round and this draft cannot be
  represented as ReviewGPT merge-ready. No Round 11 was started.

The stack remains based on #560 (`agent/group-reaction-context`).

## Deployment

This PR stacks on #560 and must not deploy before its base. First explicitly
set and verify `HOSTED_LINQ_GROUP_REACTION_CONTEXT_ENABLED=0`, then apply the
expand-safe migration. Deploy web with
`HOSTED_GROUP_JOIN_OFFER_EFFECT_REQUIRED=0`, then deploy and immediately drain
the effect-aware Cloudflare/runner bundle. After proving the managed-bundle
fingerprint and that no old runner remains, set the effect-required gate to
exactly `1`, then enable the reaction-context gate. Before activation either
side may roll back independently; afterward, unset the effect gate before
rolling back the runner or keep the effect-aware runner and requiring web as
the rollback floor.


rocketman-21:
Superseded by #655, rebuilt directly on merged #644/current main. The old #560-stacked patch and its exhausted ReviewGPT history remain closed for reference.

PR #603, opened 2026-07-13
fix(hosted): bind Telegram signup outreach to bot authority

rocketman-21:
## Why this PR exists

Telegram-only signup lost its immediate welcome when provider identity and delivery authority were correctly separated. Later review also found that replies could be denied when an authenticated inbound target differed from the preferred proactive route, and that partial business/scoped deliveries could not be safely cleaned up.

## User-visible goal and UX

A Telegram member proceeds directly to checkout or activation and receives the activation welcome only after the configured bot proves it can address that account. Replies stay anchored to the authenticated inbound conversation even when a richer proactive route remains preferred. If a multi-part Telegram reply fails, already-sent parts are removed only when the Worker can prove the exact member, bot, target, and message IDs.

## Invariants

- Telegram identity is never promoted to delivery authority by itself.
- Proactive delivery requires the current stored route; reply delivery may use an exact retained authenticated inbound target/message under the member lock.
- Cloudflare proves direct write authority with Worker-owned credentials, exact request-target equality, runtime fencing, and the configured bot identity.
- Destructive cleanup requires one Worker-HMAC proof per exact member/bot/target/message tuple; internal target/proof headers never reach Telegram.
- Bot identity constrains delivery but remains excluded from semantic conversation hashing.
- Bot changes, identity changes, failed proof, route mismatch, and final-sender bot mismatch fail closed.
- Non-idempotent sends remain terminally ambiguous when delivery or rollback cannot be confirmed.
- The existing routing, mailbox, outbox, and member-lock owners remain authoritative; no queue, reconciliation service, or new durable state owner is introduced.

## Verification

- Focused final remediation: Web authorization 11/11; Cloudflare egress/platform 354/354; assistant-engine channel/helper/outbox 127/127; hosted callbacks 166/166; operator runtime helpers 35/35; hosted parsers 43/43.
- Full production runner build passed under TypeScript 7.0.2. Exact bundle measurement: 1,460,739-byte entry, 7,095,019-byte static closure, and 8,738,459-byte total under the unchanged 9,300,000-byte ceiling. Ratchet tests passed 28/28.
- Web, Cloudflare, contracts, operator-config, assistant-engine, assistant-runtime, and reverse-dependency typechecks passed. Full Web lint passed with zero errors and 11 unrelated existing warnings.
- Complete Cloudflare Node/Workers and Web verification passed before the final base merges; merge-sensitive focused tests and Web typecheck passed again on latest `main`.
- Required coverage/write re-audit: zero actionable findings. Required security/privacy audit: zero medium-or-higher findings.
- ReviewGPT round 12 accepted and fixed two High findings. Round 13 runs on the exact pushed head concurrently with CI.
- Remaining external proof gap: no live Privy-to-real-Telegram replay; deterministic credential, route, runtime, cleanup, and provider boundaries are covered.

## Added/deleted lines

| Area | Added | Deleted |
| --- | ---: | ---: |
| Source | 2,065 | 142 |
| Tests | 3,645 | 220 |
| Docs | 582 | 3 |
| Config/tooling | 9 | 5 |
| Generated/other | 0 | 0 |

1. Deploy compatible Vercel/web readers and the signed Telegram authorization callback with `HOSTED_TELEGRAM_BOT_BOUND_TARGET_PRODUCER_ENABLED=0`, then wait for prior functions to drain.
2. Deploy Cloudflare Worker/runner with `container_rollout=immediate`; prove the new runner fingerprint, current-route delivery, retained-inbound reply, business partial cleanup, and mismatched-proof rejection.
3. Enable the bot-bound target producer in a second Web deploy.

For rollback, disable the producer first. Once proof-bearing cleanup metadata or bot-bound routes/outbox work can exist, the compatible Web reader and proof-aware Worker/runner are rollback floors; forward-fix rather than rolling either plane below them.


rocketman-21:
Closing in favor of a clean replacement from current main. The original requirement is small; this branch accumulated disproportionate cross-plane Telegram authority and rollout machinery during review. The replacement will default to deletion, reuse existing routing/delivery owners, and add complexity only where the direct signup-welcome path proves it necessary.

PR #606, opened 2026-07-13
Fix merged automation, Family billing, and group actor regressions

rocketman-21:
## Why

Merged billing, automation, and actor-routing changes left coupled regressions: Family payment cleanup could lose durable provider ownership, direct and Family billing could race into dual authority, account deletion could miss provider objects, actor paging could cross the wrong conversation participant or configured account, and legacy Linq route repair could either over-authorize or consume queued work.

## User-visible goal

Keep Family upgrades, retries, and account deletion exact and recoverable without duplicate billing; keep automation replies attached to the intended actor and authenticated account/conversation; and preserve queued device activity and fresh foreground input while legacy routes are repaired safely.

## How it works

- Stores separate encrypted, event-owned Family compensation and Pulse cleanup receipts, then retries each exact provider obligation without recomputing mutable eligibility. Family retains exact-invoice refund ownership; Pulse owns cancellation only.
- Serializes direct-versus-Family payment authority under the existing member/group locks, converts an owner's existing direct subscription atomically, and settles any persisted Family Checkout before conversion.
- Persists exact Checkout sessions, expires or cancels provider state before deletion, and bounds pre-migration completed-session discovery to the July 2026 compatibility window. Deletion fails closed when that legacy Stripe scan cannot run.
- Revalidates direct Checkout eligibility, Family authority, and canonical subscription state under the final member lock after Stripe creates the session. A rejected Family session is expired and its exact attempt/session is retired by compare-and-set, so retries get a fresh idempotency key while newer attempts survive.
- Carries the notifier's exact newly-staged input IDs into active-turn admission. Foreground selection folds the already-materialized same-route/same-actor prefix, stops at the first different group actor, and never scans unrelated pending backlog.
- Requires exact configured-account identity, keeps preference causality anchored to the first accepted input, and preserves supported same-actor live-turn folding.
- Requires signed `member-home` confirmation before marking legacy Linq routes. The bounded repair runs before cron only when no selected input or maintenance-yield condition already defers cron; it does not restore the retired pending-proof/current-inbound compatibility machinery.
- Carries the direct-home-only assertion on the exact outbox intent through retries and final provider admission. New intents persist an explicit classification; legacy unanchored direct sends fail closed while reply-anchored intents retain ordinary authority. Web rejects the claim if the target has any durable thread route.

## Invariants

- One member has one payment authority: direct or Family, never both.
- Accepted Stripe cleanup is durable, idempotent, and executable from its owning event; invoice refunds never guess a charge.
- Account deletion fails closed until every known Checkout, subscription, customer, and accepted cleanup receipt is settled.
- A failed Family checkout cleanup can retire only its exact attempt/session, never a newer concurrent attempt.
- Active turns read only exact notified IDs; actor cursors never cross a different actor or configured account, and later same-actor inputs cannot rewrite the causal anchor.
- Hosted Linq provider admission requires a marked canonical or authority-matched device parent; copied, group, unknown, and unparented legacy targets remain fail-closed.
- A direct-home-only delivery cannot target any durable thread route, including a route currently owned by the same member.
- Selected foreground/pending input, maintenance yield, or abort prevents compatibility repair from delaying the critical reply path or writing a partial confirmation batch.

## Validation

- Latest `main`, including TypeScript 7, is merged. Hosted web, assistant-engine, and assistant-runtime typechecks pass.
- Round 4 focused proof: account deletion and Family plan 169/169; active-turn controller 17/17; hosted turn input 13/13 plus the focused initial actor-prefix regression; rapid mailbox exact-ID admission and isolated outbox timing regressions pass.
- Diff-aware verification passed dependency, workspace-boundary, hosted-runtime, Temporal, crypto, log, and all affected typecheck stages. Its broad parallel test stage exposed one setup-wizard assertion that reproduces unchanged on exact `origin/main`; change-related failures passed after focused correction or isolation.
- Coverage-write added the newer-concurrent-Family-attempt regression and found no unresolved proof gap. Security/privacy review found no evidence-backed Medium-or-higher issue.
- Final conflict-marker, diff, identifier/private-path, and raw authorization-value scans passed.
- Exact pushed-head CI and ReviewGPT remain required merge gates.

## Deployment compatibility

Deploy Web first so the signed control plane understands and enforces `directHomeRouteOnly`, then deploy Cloudflare Worker/runner with `container_rollout=immediate` and wait for every runner to report the new bundle fingerprint before new classified outbox writes. Do not roll runners below the boolean-aware schema afterward. Verify direct-home rejection against a durable thread route, one server-confirmed legacy-home repair, normal reply delivery, and the Family/Stripe post-deploy contract migrations.

## Change shape

- Source: +4,563 / -506
- Tests: +6,769 / -478
- Docs: +542 / -14
- Config/tooling: +44 / -0
- Generated/other: +0 / -0


rocketman-21:
Superseded by three independently reviewable drafts rebuilt from current `main`:

- #637 — narrow proactive Linq route authority (+49/-18 source)
- #638 — exact inactive legacy Family cleanup (+214/-5 source)
- #639 — exact active-turn group actor ordering (+193/-99 source)

The replacement patches deliberately drop #606's route scanner/maintenance loop, schema and migration work, generic billing compensation machinery, owner-migration/payment-authority expansion, persisted route index, candidate buffer, and broad input-source query API.

Each replacement is running its own exact-head CI and ReviewGPT gate. #606 is being closed without merge; no finding or gate is being bypassed.

PR #607, opened 2026-07-13
Fix proactive thread usage-limit notices

rocketman-21:
## Why this PR exists

PR #570 introduced usage-aware allowance classification, but a proactive crossing notice still needs exact authority for the originating provider thread. Transient mailbox route state can disappear across a restart or deferred post-checkpoint flush, which previously caused the crossing path to skip with `home_route_missing`.

This PR reconstructs the target from durable accepted assistant-input events. It does not invent a route lookup and never falls back from accepted but unusable provenance to a participant's personal home.

## User goal / visible behavior

When an authorized Linq or Telegram thread crosses its Murph AI allowance, that same thread receives one neutral `thread_usage_limit_reached` reply immediately. If persisted accepted inputs do not prove one exact permitted route, no proactive message is sent; the existing next-current-inbound reset backstop remains.

## Invariants

- Derive the delivery target after the foreground checkpoint by rereading every provider-accepted input ID from durable assistant-input state.
- Use the shared durable reply-target route parser; optional Telegram enrichment metadata must not gate an otherwise exact Telegram reply target.
- Require all accepted inputs to prove the same route; choose the newest reply target with the canonical assistant-input cursor, so durable mailbox lane sequence outranks provider timestamps and caller order.
- Preserve legacy home fallback only when there are zero accepted input IDs.
- Return explicit `null` for every non-empty accepted-input set that is missing, unreadable, mixed, unsupported, invalid, or different-route, preventing unrelated home fallback.
- Preserve direct Linq (`routeAuthority: null`) and Telegram reply semantics from PR #570.
- Require explicit external route authority and confirmed group directness for Linq group reconstruction.
- Do not persist or merge a parallel mailbox route projection.
- Keep web as the owner of usage accounting, notice claims, and provider dispatch; runtime carries only the bounded target.
- Revalidate provider authority against current canonical routing state before dispatch.
- Never derive a thread-container recipient from a participant's personal home route.

## Verification

- Final cursor-order regression: mailbox sequence 42 wins even when its provider timestamp is earlier than sequence 41; focused workspace-runner suite 95/95 and assistant-runtime typecheck passed.
- Focused security/privacy re-audit: zero medium-or-higher findings; exact route agreement and every fail-closed boundary remain unchanged.
- Focused coverage-write re-audit: no proof gap; independently confirmed the regression fails the former `Date.parse` loop and passes the canonical comparator.
- Post-rebase web usage suite: 38/38.
- Production-shaped Telegram regression: ordinary import persists `sourceMetadata: null` plus an exact reply target, and the durable resolver returns that exact Telegram thread.
- Unsupported-origin regressions: zero accepted IDs resolve to omitted/`undefined`; accepted email and WhatsApp inputs resolve to explicit `null`; web suppresses home fallback for both ordinary and thread notices.
- Current exact-head assistant-runtime sweep reached 1,606 passing and 2 skipped; seven overload-sensitive failures outside this diff were isolated, while the affected workspace-runner file passed 95/95.
- The prior pushed head completed all GitHub Actions and Vercel successfully; the final cursor-order head is receiving a fresh complete CI cycle alongside ReviewGPT.
- `git diff --check`, clean-worktree review, exact current-`main` merge-base proof, range comparison, and added-line privacy scan passed.

## Deployment concerns

Deploy the Cloudflare Worker and runner bundle first with `container_rollout=immediate`; confirm the managed-container smoke reports the new bundle fingerprint, then deploy web. Old web ignores the additive target and retains the next-inbound notice. New web receiving an old/no-target request refuses personal-home fallback for `thread_usage_limit_reached`, so reverse skew is safe but defers the proactive notice.

Post-deploy, cross a test group allowance and verify exactly one neutral reply in the same thread. Also verify exact Telegram and direct Linq delivery, plus suppression for missing and unsupported accepted origins.

## Diff breakdown

- Source: +166 / -162
- Tests: +451 / -157
- Docs: +147 / -0
- Config/tooling: +0 / -0
- Generated/other: +0 / -0


rocketman-21:
Temporarily closing and immediately reopening to retrigger the missing GitHub Actions pull_request event for head 72da02362680; the pushed head and PR diff are unchanged.

PR #625, opened 2026-07-14
fix(hosted): preserve reply and home-route continuation

rocketman-21:
## Why this PR exists

Production reply liveness regressed across a large hosted-runtime merge. The audit found no basis for another broad revert: the fresh-input priority fix, system-note exclusion, checkpoint timing, pending index, and durable route-repair pieces are directionally correct and still required. This PR closes the remaining owner and continuation gaps without restoring reverted authority logic or adding another state owner.

## User goal and experience

Supported email, Linq/iMessage, and Telegram inputs should receive a current reply even when they arrive in a mixed mailbox burst, contain audio/video that needs parsing, or resume after a checkpoint/restart. Text-bearing messages stay on the fast path; media-only messages wait for a transcript or explicit terminal parser evidence, then receive one reply. A personal Linq reminder may follow an authorized direct-home transition, while explicit and group targets remain fail-closed.

WhatsApp is unsupported and is intentionally outside this PR. All PR-specific transient WhatsApp auto-reply support was deleted; baseline inbound/provider code on `main` is unchanged.

## What changed

- Treat the optional staged assistant-input ID as the fresh foreground capability. Context-only rows remain durable, and a mailbox pass leaves later fresh reply work under the mailbox owner with an immediate continuation.
- Re-probe the durable pending-input owner after an empty fresh pass so an older ineligible row cannot strand later supported-channel work.
- Keep text-plus-media replyable immediately, but hold media-only audio/video until semantic evidence is terminal.
- Use the existing durable inbox parser jobs as the sole parser owner. Idle maintenance drains one exact capture/job at a time, survives abort/restart, refreshes the indexed event, and carries the earliest retry into the existing workspace wake.
- Restore only the exact canonical capture/month when a local projection is missing, atomically restoring the projection and parser job. Missing capture/job state terminalizes explicitly instead of looping.
- Permit stale implicit Linq home routing only to a distinct current personal home, with target locks, revalidation, and the provider-dispatch fence retained.

## Invariants

- Fresh foreground input outranks stale pending input, without dropping or starving either owner.
- A media-only turn never reaches the model with only generated placeholder text and pending parser evidence.
- Parser work is bounded and exact-owner; no unowned job drain, in-memory duplicate queue, new schema, or new durable state owner.
- Every accepted supported-channel input remains either replyable, durably pending, or explicitly terminal.
- Email privacy routing remains fail-closed; group-scoped raw content is not projected into the personal vault.
- Explicit Linq targets, replies, recipient-bearing sends, same-target fallbacks, and foreign group routes remain rejected.
- No dependency, environment variable, public API, or migration was added.

## ReviewGPT lineage and change shape

ReviewGPT first-reviewed head: e1c1e9c7c4038231e41a1d2a738c74439dfeb98c

Classification rule: paths under `test/` or named `*.test.*` are tests/fixtures; `agent-docs/**` and `docs/**` are docs; dependency/build/script manifests are config/tooling; remaining authored runtime files are source. No binary or generated files are present.

First-reviewed baseline:

| --- | ---: | ---: |
| Source | 36 | 4 |
| Tests / fixtures | 347 | 37 |
| Docs | 48 | 0 |
| Config / tooling | 0 | 0 |
| Generated / other | 0 | 0 |
| **Total** | **431** | **41** |

Current head `eb7c0fa8a2d1c02e0ff606559d1d7663806759e9`:

| --- | ---: | ---: |
| Source | 1,180 | 304 |
| Tests / fixtures | 3,181 | 1,001 |
| Docs | 93 | 8 |
| Config / tooling | 0 | 0 |
| Generated / other | 0 | 0 |
| **Total** | **4,454** | **1,313** |

Authored-source additions grew by 1,144 lines and authored-source churn grew from 40 to 1,484 lines after review remediation. The larger proof surface is kept separate in the table.

## Anomaly retrospective

Original requirement: restore durable current replies after the incident and audit the risky merge without broad-reverting still-correct route repair, pending recovery, or checkpoint behavior.

- Round 1 accepted that projection/enrichment could block a later replyable row and that foreground admission had competing signals. The correction made the optional input ID the single foreground capability and kept already-durable parser failure metrics off the reply lane.
- Round 2 accepted that synchronous parser drain still blocked the sequential mailbox lane. The initial correction moved it out of that lane.
- Round 3 accepted the media-only semantic gap and the duplicate in-memory parser-effect owner. The current patch gates only media-only input and deletes the transient closure/effect lifecycle in favor of the existing durable parser queue.
- Round 3's WhatsApp finding is out of scope after the product decision that WhatsApp is unsupported. Rather than add consent/capability infrastructure, the branch-specific support was deleted.
- A final local proof audit found that a bounded scan could drain a fifth parser job without its exact event owner. The correction requires the exact semantic-pending event inside the four-input window; the hidden job remains durable and receives the existing bounded continuation.

Decision: continue this PR because the remaining source is one coupled hosted reply-liveness path across existing mailbox, pending-input, inbox projection, parser-job, and workspace-wake owners. The remediation removed unsupported scope and the duplicate lifecycle instead of adding a queue, state machine, reconciliation pass, or compatibility layer. Source churn remains below the 2,000-line retrospective threshold and the patch is indivisible at the production behavior boundary.

## Verification

- TypeScript: `pnpm exec tsc --version` reports `7.0.2` from latest `main`.
- Typechecks passed for assistant-engine, assistant-runtime, inboxd, parsers, web, and Cloudflare.
- Assistant-engine focused coverage: 153 passed.
- Assistant-runtime focused coverage: 341 passed; parser-maintenance: 7 passed.
- Inboxd idempotency/rebuild: 17 passed; parsers: 54 passed; web Linq authority: 40 passed.
- The unchanged baseline WhatsApp regression passed; no branch-specific WhatsApp behavior remains.
- `git diff --check`, unsafe-cast/identifier/credential scans, clean worktree, and merge-tree proof against latest `main` passed.
- Independent final owner, parser, and authority reviews found no Critical, High, or justified Complexity Collapse finding. They also confirmed `6e1759b3` should remain and found no remaining broad-revert target from `67e98e2c`.

The local Cloudflare Linq audio E2E did not reach its test: its full-stack `beforeAll` setup timed out after 300 seconds while the workstation load average exceeded 130. The previous pushed CI failure was behavioral and showed pending parser evidence reaching the assistant; this head closes that exact mechanism. The exact-head hosted E2E in PR CI is the required end-to-end result.

## Deployment and compatibility

No production or manual deployment is part of this PR, and none was performed. Keep this PR draft; do not merge or deploy it tonight.

There is no new persisted shape or cross-version protocol. Web routing and Cloudflare/runner reply recovery are independently backward-compatible, so no tandem deploy or `container_rollout=immediate` is required for compatibility. If this is deployed later, ship Cloudflare/runner first to shorten the incident window; web may follow independently. Warm old runner containers can only continue the old missed-reply behavior until recycled—they cannot misread new state. The maximum skew exposure is therefore the supported-channel inputs handled by still-old containers during that rollout, not data corruption or an irreversible migration. Current production volume and rollout duration were not queried because deployment was not authorized. Monitor reply latency, pending parser jobs/evidence, and Linq authority rejections; smoke a normal reply, media-only reply, mixed mailbox burst, restart continuation, and stale personal-home reminder after convergence.


rocketman-21:
Closing in favor of #684. The broad parser/assistant-runtime architecture in this PR is no longer worth landing after the related fixes that have since reached main. #684 carries only the independently reproduced parser-abort bug as a 9-line production change plus focused regression coverage.

PR #637, opened 2026-07-14
Tighten proactive Linq route authority

rocketman-21:
## Why

PR #606 mixed billing, active-turn ordering, and Linq authority into one oversized patch. Current `main` now owns scheduled-route resolution, so this PR keeps only the missing delivery invariant: a proactive home send must still be the member's current home route at provider entry, while route-independent overnight maintenance must not be treated as a recipient send.

## User-visible goal and flow

Scheduled Linq notifications with an unverified audience fail before model work through the current `main` route resolver. A valid proactive home send may resolve a stale home chat during preflight, but provider entry verifies the exact resolved chat and never retargets it. Reply-anchored sends retain their existing authority. Overnight memory maintenance still runs because it has no recipient.

## How it works

- Classifies a send as current-home-only from existing facts: home fallback is allowed, no reply is anchored, and no mailbox items are being answered.
- Permits home-route replacement only during the authority-only preflight.
- At provider entry, verifies the resolved target exactly and rejects any durable group route, including one owned by the same member.
- Neutralizes only the ephemeral target on an overnight maintenance execution claim; the persisted automation is unchanged.
- Adds no scanner, migration, compatibility flag, queue, repair loop, or new state owner.

## Invariants

- Unverified proactive work that can notify a recipient cannot reach the model.
- Route-independent maintenance reaches its existing no-send path.
- Preflight may resolve a home target; provider entry may verify it but never retarget it.
- A proactive home send cannot become a group-thread send.
- Reply-anchored delivery behavior is unchanged.
- Web remains the route-authority and provider-dispatch-claim owner.

## ReviewGPT rounds

- **Round 1 — `d965a0c5f3`:** found that the recipient guard also blocked route-independent overnight maintenance. Accepted. The first correction added one maintenance predicate and no state or abstraction, but it was incomplete.
- **Round 2 — `7505cb2880`:** `RETROSPECTIVE_REQUIRED`; found that maintenance still carried its Linq target into audience planning, so the provider was skipped and the occurrence could be falsely consumed as success. Accepted. The correction clears only the ephemeral maintenance target and deletes the tactical exemptions. [Retrospective](https://github.com/cobuildwithus/murph/pull/637#issuecomment-4974408463).
- **Round 3 — `10bacfc8f2`:** `PASS`; no qualifying findings. It confirmed that the exact maintenance claim loses only its ephemeral target before audience planning, the canonical route is reprojected unchanged at finalization, and the maintenance turn has no delivery, dynamic-tool, session-persistence, or conversation-mutation authority. No fix or complexity was added.

No ReviewGPT suggestion introduced a new owner, queue, checkpoint, retry path, scanner, or compatibility mechanism.

## Validation

- The affected package/app typechecks pass.
- The complete diff lane reaches affected package tests, where it is blocked by an unrelated current-`main` prompt-size guard (`61,018 > 61,000`); this PR changes no prompt surface.
- Focused Web, assistant-engine, and assistant-runtime tests cover preflight-only retargeting, exact provider-entry recheck, maintenance no-send behavior, and text/voice delivery.
- The existing hosted Linq authority E2E was green on the prior pushed head; final-head CI will rerun it.
- `git diff --check` and merge-tree conflict checks pass.

## First-review change shape (immutable baseline)

Classification: authored runtime files are source; test files are tests/fixtures; the completed execution plan is docs/process. No binary or generated files are included.

| --- | ---: | ---: |
| Source | 50 | 18 |
| Tests/fixtures | 46 | 20 |
| Docs/process | 60 | 0 |
| Config/tooling | 0 | 0 |
| Generated/other | 0 | 0 |
| **Total** | **156** | **38** |

ReviewGPT first-reviewed head: d965a0c5f3b6258810fd09abaa8e59fea270acf4

## Current change shape

| --- | ---: | ---: |
| Source | 50 | 19 |
| Tests/fixtures | 121 | 13 |
| Docs/process | 60 | 0 |
| Config/tooling | 0 | 0 |
| Generated/other | 0 | 0 |
| **Total** | **231** | **32** |

## Deployment concerns

Deploy Web/Vercel first, then the Cloudflare Worker/runner with `container_rollout=immediate` and verify the new bundle fingerprint. Roll back the runner before Web. After rollout, confirm a stale home target resolves only to the current home chat and a durable group target is rejected.

## Supersedes

This replaces only the Linq-authority slice of #606. It intentionally omits that PR's scanner, maintenance repair loop, acknowledgement field, device compatibility state, and migrations.


rocketman-21:
## ReviewGPT Round 2 retrospective

**Decision:** continue this narrow PR, but replace the Round 1 layer exception with one route-neutral execution disposition and real-boundary proof. This is an ownership error, not a size or split problem: the first-reviewed production patch was small, Round 1 added one source predicate, and no state, service, helper, or fallback was introduced.

**Requirements reaffirmed**

- Unmarked proactive Linq work that can notify a recipient must fail before model/provider work.
- The exact overnight memory job is route-independent, must reach its existing maintenance provider path, and must never deliver.
- Existing intent/route facts remain the authority. No persisted boolean, scanner, repair loop, compatibility state, queue, or new lifecycle owner.
- Marked direct Linq is still rechecked at final provider entry and may never turn into a durable group send.

**Why the mechanism repeated**

Round 1 added `!maintenanceJob` only at the new cron marker guard. Cron then continued forwarding the retained Linq recipient/session fields. Notification rebuilt those fields into the maintenance message and ephemeral binding, ordinary conversation policy classified the audience as unverified external, and notification skipped before provider start. Cron separately exempted maintenance from that result and consumed the occurrence as a successful no-op. The existing regression mocked the notification boundary, so it stopped immediately before the failing owner.

**Chosen correction**

Cron will derive one ephemeral nullable notification target: null only for the already identified overnight maintenance job, otherwise the claimed target. Target validation, the unmarked-Linq guard, and all recipient/session/delivery fields will use that one disposition. This deletes the standalone marker exception and the post-turn maintenance audience exemption. The existing notification maintenance profile, evidence, provider-start replay barrier, and exact-skip enforcement remain unchanged.

I am explicitly rejecting maintenance-aware conversation policy, a generic allow-unverified flag, treating targetless Linq as private, persisted route rewrites, migrations, scanners, and a second provider execution path.

**Required proof**

A production-faithful regression will traverse cron, real notification session construction, real shared planning, and real conversation policy while mocking only provider and outbound edges. It must show retained unmarked Linq maintenance reaches provider admission and consumes only after provider work, sends nothing, and carries no recipient audience; the equivalent notifying automation must still fail before provider admission and remain retryable.

This decision stays inside the original scope and adds no new state or ownership concept.

PR #651, opened 2026-07-14
Keep the latest ten group reactions as context

rocketman-21:
## Why

PR #644 landed the minimal one-shot group-reaction context path, but it retained only the latest reaction and omitted the reacting participant from the hint. This follow-up gives Murph the useful context the product needs while preserving the same small ownership boundary.

## User flow

For a verified reaction in an established group chat, the web route appends one bounded entry containing:

- the canonical active participant handle
- whether the reaction was added or removed and its type
- the exact bounded reacted-to message or selected part

The existing encrypted nullable route field keeps at most the newest ten entries in insertion order. The next normally admitted group message consumes and clears them together through the existing `conversation.message` sidecar. Reactions still create no mailbox item, wake, reply candidate, or separate lifecycle.

The follow-up also keeps the narrow hardening that missed #644: broader URL scrubbing without damaging labels such as `HRV:42`, a 512-character pre-scrub work bound, explicit active-roster checks, and a 500 ms bound on optional route crypto.

It also compresses one inherited join-offer instruction by 34 characters while preserving the required `Like this message` opening, keeping the stable assistant prompt inside its existing cache budget without raising the ratchet.

Exact-head CI also exposed an inherited mailbox-continuation scheduling gap. The canonical intermediate import checkpoint now remains immediately due when imported mailbox work has not yet reached the assistant phase. This is a two-line wake projection at the existing checkpoint owner, with no new state or lifecycle.

The Linq scheduled-reminder E2E still proves that setup completes and checkpoints before a persisted future wake fires without another nudge. Its CI-only fast profile now leaves 90 seconds of runway because loaded shared runners repeatedly needed 47-53 seconds to finish that checkpoint; production timing is unchanged.

## Invariants

- One existing encrypted route column remains the only transient state owner; there is no new table, service, worker, cursor, scanner, reconciliation loop, or reaction queue lifecycle.
- The buffer is FIFO and capped at ten. An eleventh append drops only the oldest entry.
- Actor, action/type, and target text are validated and bounded before encryption. Provider IDs, URLs, and attachment metadata are excluded.
- Join-offer reactions remain separate, and duplicate provider events cannot append twice.
- The next ordinary admitted message consumes the whole buffer once. A failed or raced mailbox append rolls consumption back.
- Legacy single-string ciphertext remains readable and is upgraded naturally on the next append.
- Missing, malformed, corrupt, or crypto-unavailable optional context fails open and never blocks an ordinary inbound message.
- Downstream storage remains the existing optional scalar sidecar; no strict assistant event or durable conversation history gains reaction context.
- An intermediate canonical mailbox-import checkpoint remains immediately scheduled until the existing assistant phase can consume the imported work.

## Change shape

Current head:

- Source: +95/-32
- Tests: +563/-41
- Docs: +27/-15
- Config/tooling: +0/-0
- Generated/other: +0/-0
- Total: +685/-88 across 22 files

Immutable round-1 baseline:

- Source: +91/-31
- Tests: +514/-19
- Docs: +24/-14
- Config/tooling: +0/-0
- Generated/other: +0/-0
- Total: +629/-64 across 18 files

ReviewGPT first-reviewed head: 0414fae912aee1db361982b12549a495b8994b39

An earlier ReviewGPT invocation on #644 returned `INVALID` because its evidence package omitted the PR body and immutable first-review change-shape baseline. It made no substantive code disposition.

Round 1 at `0414fae912aee1db361982b12549a495b8994b39` found that the expanded bare-domain scrub could perform quadratic regex work before the retained output bound. The finding was accepted. The remediation only reorders the existing operations to slice each normalized candidate to 512 characters before scrubbing, plus one adversarial 32-part regression test; it adds no parser, queue, cache, or state owner.

Round 2 at `9e0573024bf405423aa95693d542135de1ed284c` passed. It verified the URL-bound correction, accepted the inherited mailbox-continuation fix as reuse of the existing checkpoint/wake owner, and found no qualifying review-induced issue. It required no further code change. The later exact-head delta is only the isolated E2E timing correction above plus merged base history, so repo policy does not require another substantive round.

## Deployment concerns

Deploy the hosted runner bundle before the web producer. The existing optional scalar contract now allows the joined ten-entry hint (up to 5,129 characters); an older runner still enforces the former 512-character maximum. No schema migration or tandem state transition is required, and the safe order avoids sending a larger optional hint to an older parser.

After the new web producer begins writing the JSON-array slot shape, forward-fix web rather than rolling it back until transient slots have been consumed or cleared. This is rollout discipline only; the PR deliberately adds no version handshake or migration state.


rocketman-21:
Root-cause evidence for the remaining Linq reminder failure: the immediately preceding exact tree in #657 passed this same lane (run 29381438788, job 87246360819). On this PR, the only workspace-runner production delta is the two fields added to the canonical mailbox-progress checkpoint: `nextWakeAt: resolveHostedWorkspaceRunnerNowIso(input.now)` and `nextWakeReason: HOSTED_ASSISTANT_WAKE_REASON`. The failing exact-head run then expected a wake no later than 01:44:16Z but persisted 01:51:23Z after the forced extra pass reached normal retry/backoff. Smallest correction: delete those two fields and revert only the unit assertions that require the intermediate checkpoint/result to stay immediately scheduled. Preserve the existing checkpoint/session metadata owner; do not add timing, retry, or checkpoint infrastructure. This is read-only diagnosis from run 29382021568/job 87248164535; I did not rerun or edit the branch.

rocketman-21:
Correction to my prior diagnosis: concurrent exact-head runs for #637 and #639, neither of which contains this PR's two workspace-runner wake fields, have now failed the same Linq reminder lane on the same base. That disproves treating those two fields as the sole/root cause. Please do not delete them solely on my earlier comment. The shared evidence now points at the CI-only 30-second fast-gate window: once setup misses exposing a future wake before the due timestamp passes, the current waiter condition can no longer succeed but keeps polling to 120 seconds. I am continuing read-only comparison and will treat reruns as flake confirmation rather than prescribe runtime machinery.

rocketman-21:
Confirmed common root cause from #637 attempt 1 (run 29382419322/job 87249291730): the first runtime execution took 50.755s, while `MURPH_HOSTED_LOCAL_E2E_FAST_GATE=1` gives the scheduled reminder only a 30s lead. The allowed due time was 01:53:32Z; the first checkpoint completed at 01:53:53Z, 21s after due. At that point `latestNextWakeAt > Date.now() && latestNextWakeAt <= dueAt` is impossible, so the helper polls until its 120s timeout and later reports the ordinary retry wake. #639 reproduced the same lane independently. Smallest durable correction is test-only: keep the fast gate's 1ms idle-checkpoint delay, but restore its reminder `leadMs` and setup text to the existing 60s / about-one-minute values (and update the timing assertion). No runtime scheduler/checkpoint/retry change is warranted by this failure.

PR #677, opened 2026-07-15
Preserve foreground reply authority during receipt recovery

rocketman-21:
## Why

Canonical receipt replay runs during workspace restore, before mailbox fetch. A malformed or irreconcilable receipt therefore had boot authority and could retry forever while a durably accepted current conversation never reached the assistant. The first liveness correction also needed a clear boundary between corrupt recovery input and an artifact-store outage: the latter is not evidence that a valid durable receipt should be discarded.

## User-visible goal and flow

A durably accepted current conversation can still receive a reply when deterministic receipt recovery rejects unsafe state. The runtime discards partial local mutation, reloads the authoritative snapshot, removes the rejected active receipt-log reference, emits a secret-safe degraded diagnostic, and admits foreground work. An explicitly retryable artifact transport/service failure instead clears partial local state, preserves the unchanged durable receipt reference, and fails that invocation so the next invocation retries the same canonical input. A permanently malformed encrypted object is terminal and follows the reject-and-continue path in the current invocation.

## Invariants

- Foreground reply authority cannot be indefinitely withheld by recovery, replay, audit, checkpoint, projection, maintenance, or diagnostics.
- Unsafe receipt mutation remains fail-stop; partial replay is never checkpointed and conflicting canonical state is never fabricated or silently appended.
- Retryable artifact transport failure is not proof of corrupt receipt input and cannot authorize deletion of its durable reference; proven terminal persisted-content failure cannot retain foreground authority.
- The authoritative snapshot is the sole rollback and continuation boundary after rejected recovery input.
- Rejected deterministic recovery input does not retain replay, append, checkpoint, repair, or foreground-admission authority.
- Host cancellation still wins exactly.
- Diagnostics remain secret-safe and non-blocking.
- Non-audit JSONL receipt replay remains byte-strict; audit recovery uses schema-valid immutable IDs.

## Architecture

The restore owner keeps canonical receipt replay fail-stop. Existing owners carry one explicit disposition: the canonical encrypted-envelope parser rejects malformed metadata before lookup, the encrypted R2 reader identifies deterministic envelope/decryption failure, the artifact HTTP route transports it, and the runtime artifact port exposes `retryable`. Receipt recovery retries only that explicit retryable case; terminal, missing, malformed, conflicting, or inapplicable content reloads the authoritative snapshot and removes the rejected active ref before foreground admission. There is no repair metadata owner, persisted retry count, collection, chain, queue, retry worker, recovery service, compatibility layer, or web contract.

## Review remediation

- Round 1 accepted one finding: catch-and-continue could expose and checkpoint a partial multi-action replay while clearing its retry reference. The first correction restored the authoritative snapshot and added a singleton repair reference.
- Round 2 required a retrospective because a second failed log could overwrite that singleton, repeating the same reference-loss mechanism.
- Retrospective decision: complexity collapse by deletion. Repair ownership was removed instead of expanded into a collection or lifecycle. A failed deterministic receipt batch is rejected as unauthorized recovery input; the authoritative snapshot remains canonical and foreground work continues.
- Round 3 accepted one boundary finding: a transient artifact 503/timeout was treated like corrupt input and could checkpoint away a valid receipt ref. The correction distinguishes the existing store outcomes structurally: a returned missing object remains deterministic rejection, while a thrown read preserves the durable ref and retries on a later invocation.
- Round 4 required a retrospective because every thrown read was treated as retryable, so permanently malformed encrypted objects could again withhold foreground authority forever.
- Round 4 retrospective decision: assign disposition to existing owners only. The encrypted R2 reader proves terminal persisted-content failure; the existing route and artifact port carry that fact. Delete the receipt-specific catch-all wrapper and add no repair/retry state or lifecycle.
- Round 5 accepted one residual owner-boundary finding: the canonical encrypted-envelope parser accepts malformed key IDs, so the historical-root resolver can turn deterministic persisted corruption into a retryable HTTP failure and recreate the foreground-blocking loop.
- Five-round retrospective: the original requirement remains indivisible—foreground authority requires a truthful terminal/retryable distinction—but the repeated findings came from assigning that distinction below an overly permissive canonical parser. The smallest continuation is to tighten the existing envelope parser so key IDs are canonical and bounded before lookup. This adds no owner, state, retry mechanism, repair path, queue, service, lifecycle, or compatibility layer; malformed persisted content then reaches the already-implemented terminal path.
- Explicit continuation decision: land that parser-boundary correction, make every other completion audit, verification check, and CI job green, then run round 6. Completion policy now states that the five-round cap pauses another review; it does not prohibit fixing a known round-five bug.
- Round 5 correction: the canonical parser and writer reject surrounding-whitespace and over-256-character key IDs before lookup or encryption. Reader coverage proves the resolver is not called, and the production runner route returns terminal 422 for malformed encrypted metadata.
- Round 6 required another retrospective because embedded control characters still passed the partial trim/length rule and could fail database lookup as retryable infrastructure.
- Round 6 retrospective decision: cipher-envelope key IDs are closed portable identifiers. Every production producer emits an ASCII token (`udrk:...`, `hdk:...`, or a browser-replica hash-derived ID). Replace the partial rules with one canonical parser language: 1–256 ASCII alphanumeric/period/underscore/colon/hyphen characters, beginning with an alphanumeric. This keeps validity and terminality at the existing envelope owner, rejects NUL and all other control/Unicode input before lookup, and adds no reader heuristic, state owner, service, queue, retry mechanism, repair path, lifecycle, or compatibility layer.
- Round 6 correction: the shared envelope parser and writer now enforce that complete portable-ID language. Boundary coverage proves the minimum, permitted punctuation, and exact maximum; malformed-content coverage proves embedded NUL and non-ASCII input never reach historical lookup and return terminal 422 through the production artifact route.
- Round 7 correction verification passed with zero findings. The verified production producers all emit IDs inside the closed grammar, and the end-to-end terminal classification continues through foreground admission without adding compatibility or recovery machinery.
- Current-main reconciliation had one additive barrel-export conflict and three semantic test-wrapper conflicts after `main` made artifact read context mandatory. The merge retains both existing platform exports and transparently forwards the required context through the PR-owned wrappers; no production behavior or new compatibility path was added.
- Fresh CI exposed one more PR-owned Cloudflare test call that predated the mandatory context. The test-only follow-up supplies the adjacent production-path `workspace_restore` context; it changes no runtime behavior and therefore does not create another substantive review round.
- Round 8 accepted that same concrete Cloudflare test failure from its immutable merge-head snapshot and found no other issue. The already-pushed correction supplies the required context, passes Cloudflare typecheck and the full runner-platform suite, and is being verified on the current head.
- Round 9 passed with zero findings on the final head and verified that the test-only correction preserves 422-terminal/503-retryable classification without changing production behavior.

## Verification

- Required coverage-write audit: no unresolved findings.
- Assistant-runtime exact-diff tests: 1,666 passed, 2 skipped.
- Complete runtime entrypoint suite: 220 passed.
- Transient-read proof: receipt-log, receipt, and second-payload failures all preserve the original error/ref and create no checkpoint or foreground mutation; the second-payload case proves an earlier partial action is cleared.
- Terminal-read proof: malformed encrypted content is terminal through reader, route, platform, and assistant boundaries; receipt recovery clears partial effects, rejects the active ref, imports the pending conversation, and reaches the assistant in the same invocation.
- Assistant-runtime coverage: 88.27% statements, 81.52% branches, 91.54% functions, 88.55% lines.
- Runtime-state verification: 177 passed; focused canonical parser: 5 passed.
- Cloudflare verification: 1,827 passed; focused reader/route proof: 208 passed.
- Current-main conflict proof: assistant-runtime typecheck and all 223 workspace-entrypoint tests passed.
- Post-merge CI proof: Cloudflare typecheck and all 129 runner-platform tests passed.
- `pnpm test:diff`: passed on the exact pushed source and test diff, including affected typechecks, architecture/dependency/privacy guards, and production builds.
- Scenario-manifest integrity: 204 scenarios, 11 sample inputs, and 28 golden-output directories passed.
- Privacy scan and `git diff --check`: passed.

## Change shape

| Area | Added | Deleted |
| --- | ---: | ---: |
| Production source | 269 | 107 |
| Tests | 1,117 | 87 |
| Durable docs | 32 | 5 |
| Execution plans | 474 | 0 |
| Config/tooling/generated | 0 | 0 |

## Review heads

| Round | Head | Scope | Outcome |
| --- | --- | --- | --- |
| 1 | `904912706e8818c5329480791ff59b6036759c3b` | Full patch | One accepted finding |
| 2 | `e264a17526c45563549816ab70ce5c93a13f2007` | First correction | Retrospective required |
| 3 | `70bfd43ab5be9b3c4e90ec7f96c7b2074e135f8a` | Complexity collapse | One accepted finding |
| 4 | `6e54c7f478c494f8b1714733a7eff1f57f99c1c0` | Artifact availability boundary | Retrospective required |
| 5 | `f320ca7088d761b25aab2dafc36e9d422f6384a3` | Terminal artifact disposition | One accepted finding; hard cap reached |
| 6 | `4aae18e3a3964d4c98ea61cd5c178ec8f3a9c2d4` | Canonical key-ID boundary | Retrospective required |
| 7 | `b43e5d00031d294697df36797c5023b4e0bbfb1c` | Portable key-ID correction | PASS — zero findings; GPT-5.6 Pro verified |
| 8 | `e64a83d9ee19a4f1bc054ab33f2eeb8f7983e7ea` | Current-main manual conflict resolution | One accepted test-call finding; already corrected |
| 9 | `eed10da4670489a4220688d98b694b423cda8b29` | Cloudflare context correction | PASS — zero findings; GPT-5.6 Pro verified |

## Deployment

This is Cloudflare-only at runtime. After merge, deploy Cloudflare with an immediate container rollout. Verify the affected runtime advances mailbox conversation progress, accepts the next foreground canonical write, and no longer repeats the deterministic restore failure loop.

ReviewGPT first-reviewed head: 904912706e8818c5329480791ff59b6036759c3b


rocketman-21:

- Original requirement: unsafe receipt replay must never corrupt canonical state or indefinitely withhold a durably accepted foreground reply.
- First-reviewed shape: per-receipt catch-and-continue with no repair owner. Round-one remediation delta: production +200/-126, tests +270/-38, docs +7, plan +53. It added snapshot rollback plus one singleton repair-log reference.
- Repeated mechanism accepted: after repair log A exists, a fresh active log B can fail and replace the sole reference to A. This repeats the prior durable-reference-loss mechanism.
- Decision: complexity collapse by deletion. Remove repair-log ownership entirely rather than adding a collection, chain, queue, resolver, or lifecycle. A failed active receipt log is explicitly rejected as unauthorized recovery input; its partial local effects are discarded and the authoritative snapshot remains the sole canonical source. The runtime emits a secret-safe degraded diagnostic and continues foreground work, but does not claim the rejected batch remains repairable.
- Required proof: two consecutive failures separated by a successful foreground canonical write and crash boundary; each must restore only the authoritative snapshot, clear the rejected active ref, never checkpoint partial state, preserve exact cancellation, and retain foreground admission.

This stays inside the durable foreground-reply-authority invariant while removing the review-introduced state owner.

PR #678, opened 2026-07-15
Fix low-risk assistant configuration approval

rocketman-21:
## Why this PR exists

Murph generated a secure-action approval for a simple model/reasoning preference change, and the assistant then shortened the returned capability URL by one character, producing an unavailable page. Low-risk configuration should not require a passkey/browser handoff, while sensitive vault delivery must keep its approval boundary and deliver an exact capability without exposing that capability to later model turns.

A member can explicitly ask Murph to change the model or reasoning effort and receive an immediate saved/unchanged/upgrade-required result for the next turn, without an approval link. Vault file sending remains approval-gated. The runtime appends the exact owner-issued URL to the user delivery response, while durable transcript and all model-history paths retain only capability-free semantic content.

## User experience

The member asks for an eligible model or reasoning change in a private conversation. Murph validates the current saved target and entitlement, saves only the requested fields, and confirms that the change starts next turn. Sol still reports the Edge upgrade requirement when unavailable. Vault file requests still park the file action and require the normal secure approval page; denied, expired, wrong-account, or otherwise invalid approvals continue to fail closed.

## Invariants

- Web/Postgres remains the only durable owner of model and reasoning preferences.
- A direct assistant update requires eligible accepted user intent in a locally revalidated bounded exact-successor provider batch. The runtime forwards only the terminal input id, which web binds inside the transaction to the callback member and exactly one live conversation mailbox row.
- Web re-derives active access, model availability, and Sol entitlement; model-provided identity, plan, target claims, or causal sequence are never authority.
- Sensitive disclosure, vault delivery, delegated access, identity, payment, export/delete, and irreversible actions retain their existing approval or stronger authenticated handoff.
- Vault approvals remain bound to file/action/destination/provider identity. The exact approval URL stays runtime-owned and is appended only to the delivery response.
- Durable assistant transcript, native provider continuation, committed-history fallback, stale-resume fallback, and provider prompt serialization remain capability-free.
- Required user-visible handoff or approval links are known before durable no-reply suppression/completion callbacks settle.
- Saved configuration applies only on the next hosted invocation; no mailbox item, wake, second preference store, schema, queue, dependency, or new lifecycle owner is introduced.

- Runtime-appended computer-handoff capabilities use the same delivery/history boundary as vault approval capabilities. When the model omits the browser handoff URL, the runtime still delivers the exact owner-issued URL to the member, but the URL is not persisted into committed transcript or replayed through fresh-thread/stale-resume model history. A handoff-only delivery therefore has no semantic assistant transcript entry. This is intentional: the browser takeover token is transport authority owned by the computer workflow, not conversational content for later providers.
- `packages/assistant-engine/test/assistant-codex-runtime.test.ts` proves exact computer-handoff delivery with a null semantic transcript when the model supplied no semantic message. `apps/cloudflare/test/hosted-local-computer-handoff-linq-roundtrip-e2e.test.ts` proves the hosted delivery/completion/resume workflow, and the current-head `Sensitive effect + handoff E2E` CI lane passed. `packages/assistant-engine/test/assistant-protocol-index-planning.test.ts` proves fresh-thread fallback consumes only committed semantic transcript history.

## Verification

- `pnpm test:diff packages/hosted-execution packages/assistant-engine packages/assistant-runtime apps/web apps/cloudflare`
  - all affected owner/reverse-dependent typechecks, guards, package tests, app tests, lint, smoke, and production builds passed
  - assistant engine: 2,228 passed
  - assistant runtime: 1,674 passed
  - web: 5,128 passed
  - Cloudflare: 1,819 passed
  - CLI: 1,087 passed
- Focused capability-delivery/transcript/fresh-thread suite: 368 passed.
- `pnpm hosted-local e2e vault-file-approval-resume`: passed on the current head, including exact persisted approval capability and resumed Linq attachment upload.
- Runner assembly measured entry 1,494,107B, static boot closure 7,177,743B, and total 8,865,730B. The reviewed entry baseline advanced by the exact 633B overage while preserving the existing 48KB noise band.
- Required coverage-write audit completed with no unresolved findings.

## ReviewGPT round 1 dispositions

- Accepted, high: `finish_without_reply` persisted suppression/completion state before a later vault approval link was known. The runtime now records no-reply as a turn-local candidate and settles its durable callbacks once, only after pending tools drain and only when no required user-visible link exists. Focused provider-failure coverage proves a pending vault link leaves no suppression/marker and remains retryable; genuine no-reply failure still settles once.
- Accepted, complexity collapse: the model-facing vault tool result exposed `approvalUrl`, while hidden runtime metadata carried the same capability and a URL parser/regex repair layer tried to correct model transcription. The model result now contains only filename/status, the exact owner URL stays hidden, and the runtime deterministically appends it. URL parsing, regex repair, punctuation cleanup, and dedupe machinery were deleted.

## ReviewGPT round 2 disposition and retrospective

- Accepted, high, review-induced: round 1 kept the capability out of the tool result but appended it to the provider `response`; local-service persisted that transport-complete value into the assistant transcript. A later model/reasoning route change could force a fresh thread and replay the still-live URL from committed history into the next provider prompt.
- Landed correction: the app-server/provider result now carries separate ephemeral delivery and semantic transcript representations. Delivery keeps the exact URL; transcript persistence consumes only the capability-free semantic response. Native continuation never receives the runtime-appended value, and fresh/stale fallback reads only the semantic transcript. Other providers retain their prior behavior through the optional-field fallback.
- Retrospective: required by round number and the repeated capability-ownership mechanism, and completed before the correction. Decision recorded in the PR comment: durable transcript means semantic assistant content, not transport-complete delivery text; continue with the smaller split representation. No parser/redactor, second persistence path, schema, queue, state machine, lease, migration, or reconciliation owner was added.
- Regression chain: exact approval delivery contains the URL; provider mapping keeps a capability-free transcript value; local-service gives the finalizer only that semantic value; a forced fresh thread after a model-route change replays semantic history and excludes the URL.

## Change shape

Classification rule: base-to-head authored line counts; test paths and `*.test.*` are tests/fixtures, Markdown is docs, remaining runtime/app code is source. Generated output is separate; there are no binary files.

| --- | ---: | ---: |
| Source | 262 | 181 |
| Tests/fixtures | 782 | 160 |
| Docs | 130 | 28 |
| Config/tooling | 0 | 0 |
| Generated/other | 0 | 0 |
| **Total** | **1,174** | **369** |

From the immutable first-review baseline, source additions moved +31 and deletions +44; authored-source churn rose from 368 to 443 lines. Review-driven growth is confined to the no-reply settlement correction and the ephemeral delivery/semantic transcript split. No new durable owner or compatibility mechanism was introduced.

Deploy web first: it accepts both the new direct input-bound request and the legacy exact-target approval request. Then deploy Cloudflare/runner with `container_rollout=immediate` and require managed-container smoke to report this runner-bundle fingerprint. Old warm runners remain compatible with new web by issuing the legacy approval shape. New runtime against old web would fail direct writes closed, so web is the rollout and rollback floor; roll Cloudflare back first if needed.

The incident evidence is one observed configuration request; no production-wide event count was queried. The new request shape is not persisted, so there are zero incompatible stored records to migrate or repair. During a misordered window, the maximum exposed operation is a member-initiated model/reasoning change, which fails closed and is reversible through Settings or a later turn; ordinary replies and vault approvals remain available. Post-deploy, smoke one direct reasoning change without a URL, verify the next invocation uses it, and request one vault file to confirm the exact approval URL opens and completes delivery. Then change model/reasoning again within the approval TTL and confirm the new provider prompt does not contain the URL. Remove the legacy parser/handler only after old runners and their bounded idle windows drain.

Current PR head: 4538ff390af7b0e5da7d7bec445e4197fe249c8f

ReviewGPT first-reviewed head: eb2b2c8ab8e3ca9b41abbc60c9bb76cddde374e4

### ReviewGPT first-review baseline

| --- | ---: | ---: |
| Source | 231 | 137 |
| Tests/fixtures | 468 | 145 |
| Docs | 114 | 21 |
| Config/tooling | 0 | 0 |
| Generated/other | 0 | 0 |
| **Total** | **813** | **303** |


rocketman-21:
## ReviewGPT round 2 retrospective decision

Round 2 outcome: `RETROSPECTIVE_REQUIRED`.

Accepted finding: the immediate vault tool result no longer exposes `approvalUrl`, but the runtime-appended delivery URL is still persisted in the assistant transcript. A next-turn model/reasoning change changes the Codex route fingerprint, so fresh-thread transcript fallback can place the still-live capability into model context. This repeats round 1's capability-ownership mechanism.

Requirement restatement: low-risk model/reasoning changes remain direct and input-bound; sensitive vault delivery remains approval-gated; the exact owner URL reaches the member once, overrides no-reply, and never enters any model-context source. Web/Postgres remains preference truth. No new persistence, schema, queue, dependency, lifecycle state, or durable owner is permitted.

Decision: the durable assistant transcript represents semantic assistant content, not transport-complete delivery text. The runtime will carry a separate ephemeral delivery representation for owner-appended capability URLs while transcript/fresh-thread/stale-resume/prompt history consume only the capability-free semantic response. Native continuation is already capability-free because the URL is appended after the model turn. The correction will not parse, redact, regex-match, deduplicate, or reconcile URLs.

Validation decision: add a focused end-to-end runtime test that creates a pending vault approval and next-turn model/reasoning change, forces the resulting fresh-thread fallback within the approval TTL, and proves the delivered reply contains the exact URL while the next provider prompt/transcript does not. Preserve exact link delivery, no-reply override, and existing web-owned approval binding.

Structural decision: continue with this smaller split representation. It removes the repeated dual-purpose-text mechanism without adding durable state or another owner. The incoming mainline exact-successor batching change will be reconciled by reusing its existing terminal preference-input ID primitive for both personalization and assistant configuration.

rocketman-21:
ReviewGPT round 3 completed on exact head 4538ff390af7b0e5da7d7bec445e4197fe249c8f. The initial round accepted one disclosure-only Purpose Drift finding for the intentional computer-handoff history effect; no code change was required. The PR body now names that non-obvious surface, necessity, and regression proof. One same-head retry was INVALID because its prompt omitted the immutable first-reviewed head even though the packaged metadata was correct; the corrected evidence retry preserved first-reviewed head eb2b2c8ab8e3ca9b41abbc60c9bb76cddde374e4 and previous reviewed head 2a362d785c3e862416e60f54ea35a9b15f5af453, then returned ROUND_OUTCOME: PASS with no qualifying findings. CI is green on the reviewed head.

PR #683, opened 2026-07-15
fix(ops): require verified member email

rocketman-21:
## Why this exists

The member-email composer was merged with an unsafe fallback to an unverified Stripe checkout address and a feature-only form abstraction. This follow-up removes both before operators use the new surface.

## User-visible goal and UX

- Keep `/ops/email` limited to the member's current verified email.
- Show Stripe-only members as missing email and skip them without exposing an address.
- Preserve the existing Preview, exact-draft Send, retry, and accessibility behavior with direct form primitives.

## Invariants

- A Stripe checkout or payer address is never recipient authority for free-form member correspondence.
- Unknown, suspended, Stripe-only, and addressless members are never sent an email.
- Recipient addresses remain server-only and absent from responses and logs.
- Existing fixed-purpose welcome and cancellation-feedback email rules are unchanged.
- Keep the form accessible without adding a feature-only component framework.

## Verification

- Hosted-web test run: 428 files passed plus 2 skipped; 5,163 tests passed plus 139 skipped.
- Hosted-web typecheck: passed.
- Hosted-web lint: 0 errors; 12 unrelated pre-existing warnings.
- Diff/privacy checks: passed.
- No real email was sent.

## Change shape

| --- | ---: | ---: |
| Source | 63 | 293 |
| Tests | 11 | 17 |
| Docs | 3 | 3 |
| Config/tooling | 0 | 0 |
| Generated/other | 0 | 0 |

Total: 77 added / 313 deleted.

ReviewGPT first-reviewed head: e121d1ece3ce5da93fc41a8519a6c6ac05232852


rocketman-21:
Closing at operator direction. No changes from this follow-up should be merged or deployed.

PR #719, opened 2026-07-15
Bound stale signup welcome retries

rocketman-21:
## Why this PR exists

A signup welcome remained retryable after the member had already established a healthy conversation, producing an unbounded 30-minute wake loop and risking a duplicate welcome if the provider later recovered. The shared outbox also had no finite no-progress exit for definite automatic delivery failures.

Members keep receiving fresh signup welcomes and current-inbound replies normally. A stale welcome is retired once durable auto-reply provenance proves a newer accepted reply for the same direct recipient, and definite automatic sends stop after 48 persisted provider attempts instead of cycling forever.

## User experience

There is no new UI. Healthy foreground conversations remain unchanged. Background delivery failures retry with the existing backoff through attempt 48, then become terminal with an explicit diagnostic. Ambiguous non-idempotent sends remain parked for confirmation rather than being mislabeled or replayed.

## Invariants

- Attempt 48 may run; no attempt 49 provider call may begin.
- A delivery that may already have succeeded is not terminalized as an ordinary exhausted failure.
- Replay-safe stale claims reconcile persisted/provider delivery proof before terminalization.
- Only pending or retryable signup welcomes with newer accepted auto-reply evidence for the same recipient are superseded; sending welcomes retain normal confirmation handling.
- Foreground replies retain priority and are never disabled or delayed by this cleanup.
- Newsletter parent and recipient replay cannot reset the logical retry budget by creating a fresh token.

- Group newsletter parent and per-recipient replay: exhausted logical sends are now terminal for the occurrence. Regression proof: `newsletter-outbox.test.ts` and hosted callback fan-out coverage.
- Hosted delivery collection: durable auto-reply provenance is consulted across restored outbox history so an already-persisted stale welcome self-heals. Regression proof: same-recipient, negative recipient/group/older-reply, background-mode, and dispatch-race cases in `hosted-runtime-callbacks.test.ts`.
- Shared mirror dispatch: the hosted prepared-claim path now shares the 48-attempt terminal bound. Regression proof: direct attempt-48 failure and re-claim coverage in `assistant-outbox-thresholds.test.ts`.

The persisted outbox schema is unchanged, so old and new runner bundles can read the same state. During a gradual Cloudflare rollout, an old warm runner can continue the legacy retry loop until it is replaced; deploy the Cloudflare runner bundle with `container_rollout=immediate`, then require managed-container fingerprint smoke before considering the incident fixed. `apps/web` is unchanged and needs no tandem deployment. The observed incident involved one stale welcome; realistically exposed operations during the short rollout are due retries handled by old warm containers. Rollback is data-compatible but reintroduces the unbounded loop. Post-deploy, verify the observed welcome becomes `abandoned` without provider entry, its automatic wake disappears, and no duplicate welcome is delivered.

## Change shape

Classification rule: production TypeScript is source; `*.test.ts` is tests; `agent-docs/**` is docs; no binary or generated files are included.

| --- | ---: | ---: |
| Source | 346 | 38 |
| Tests / fixtures | 758 | 13 |
| Docs | 96 | 5 |
| Config / tooling | 0 | 0 |
| Generated / other | 0 | 0 |
| **Total** | **1200** | **56** |

ReviewGPT first-reviewed head: 5ebca08ec955323a86966c704994d872ac8230cb

| Review round | Current head | Scope |
| --- | --- | --- |
| 1 | `5ebca08ec955323a86966c704994d872ac8230cb` | Full patch |

---
<a href="https://app.blacksmith.sh/cobuildwithus/codesmith/murph/pr/719"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786747220&installation_id=127572939&pr_number=719&repository=cobuildwithus%2Fmurph&return_to=https%3A%2F%2Fgithub.com%2Fcobuildwithus%2Fmurph%2Fpull%2F719&signature=335de336b9a9a8daffb8684b4fce577bdf21f38e1c6a7813ed26c3097a0742b6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


rocketman-21:
CI blocker evidence (2026-07-15): `Runtime checkpoint durability E2E` fails before this PR's retry-bound path runs, during test artifact setup in `hosted-local-canonical-receipt-lost-ack-recovery-e2e.test.ts`. The exact base commit (`743aa24537de1ee21d18788403368eb4dc835d7d`) fails at the same test and stack with `Hosted runtime crypto context fetch failed with HTTP 403`; the PR run and one isolated rerun reproduced that identical 403. All task-relevant package tests/typechecks/coverage, ReviewGPT, build, and the other hosted matrices passed. No unrelated harness fix is included in this PR.

PR #723, opened 2026-07-15
test(cloudflare): provision checkpoint recovery workspace

rocketman-21:
## Why

The required Runtime checkpoint durability E2E was deterministically failing before encrypted fixture upload because its second scenario activated a member but never provisioned that member's hosted workspace. The test-only artifact route therefore reached the production crypto-context gate without its required workspace and correctly received HTTP 403.

## User-visible goal and UX

There is no product UI or runtime behavior change. This restores a trustworthy required release gate so unrelated production changes can proceed only after the checkpoint recovery scenario proves its intended behavior.

## Invariants

- Runtime crypto context remains fail-closed for inactive members and members without a hosted workspace.
- The scenario provisions the workspace through the existing activation wake path; no authorization or harness shortcut is added.
- Provider and delivery baselines remain captured after activation.
- Production source and runtime configuration are unchanged.

## Non-obvious surfaces

The second preference-recovery scenario now waits for activation completion and asserts a non-null workspace before binding its chat and uploading encrypted incident artifacts.

## Change breakdown

- Source: +0 / -0
- Tests: +8 / -0
- Docs: +46 / -0 (completed execution plan)
- Config/tooling: +0 / -0
- Generated/other: +0 / -0

## Verification

- `pnpm hosted-local e2e canonical-receipt-lost-ack-recovery` — passed, 2/2 tests
- `pnpm test:diff apps/cloudflare/test/hosted-local-canonical-receipt-lost-ack-recovery-e2e.test.ts` — passed, 105 files / 1,833 tests
- Required coverage-write audit — passed with zero findings and no edits
- Parent final review, diff check, and privacy scan — passed

### DIFF
--- a/agent-docs/exec-plans/completed/2026-07-15-checkpoint-ci-workspace-seed.md
+++ b/agent-docs/exec-plans/completed/2026-07-15-checkpoint-ci-workspace-seed.md
@@ -1,0 +1,46 @@
+# Checkpoint CI workspace seed
+
+Status: completed
+Created: 2026-07-15
+Updated: 2026-07-15
+
+## Goal
+
+- Restore the required runtime-checkpoint CI lane by provisioning the fixture member's hosted workspace before the test-only encrypted artifact route requests that member's runtime crypto context.
+
+## Success criteria
+
+- The recovery scenario uses the existing activation path before seeding encrypted artifacts.
+- The focused hosted-local checkpoint E2E passes without weakening crypto-context authorization.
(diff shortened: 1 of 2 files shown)


rocketman-21:
Superseded by #720, which landed the same root-cause correction more minimally by seeding the checkpoint/runtime crypto context before encrypted artifact upload. #720 passed the exact Runtime checkpoint durability E2E and both required hosted gates. Closing this duplicate without merge.

PR #764, opened 2026-07-16
fix(hosted): prune transient delivery artifacts before checkpoints

rocketman-21:
## Why this PR exists

Cold-path tracing found an already-terminal assistant-generated delivery artifact retained across 19 hosted checkpoints. It inflated the encrypted workspace payload and added about 1.1 seconds to a representative restore. Reader compatibility shipped first in PR #768; this PR activates the narrow writer/cleanup side.

## User goal and visible behavior

Cold Murph replies should stop restoring obsolete one-time generated delivery files without adding a processing tier or weakening approval, retry, restart recovery, or ordinary vault-file safety. There is no new user-facing flow or copy.

## Invariants

- The only reserved ref is the exact flat `.runtime/operations/assistant/generated-deliveries/<filename>` shape. Only a newly generated, same-turn delivery obligation may use it; ordinary vault files and `exports/**` remain ordinary user data.
- Generated `send_vault_file` calls use the existing serialized dynamic-tool chain. A second call cannot race adoption, approval, or response-media mutation against the first.
- Generated sends require the provider's semantic tool-call id. The process-local JSON-RPC request id is not a substitute, and missing identity fails before adoption.
- The owned ref is deterministic per `[sessionId, turnId, toolCallId, ref]`, keeping distinct calls separate and exact call re-delivery idempotent. Identity remains intentionally attempt-scoped; a replacement provider call has a new id. Solving that would require a new durable send fact or an explicit coalescing rule, neither of which is justified here.
- Adoption has one filesystem owner transition: validate a single-link source, atomically hard-link it to the no-clobber deterministic target, verify the captured inode at both paths, keep the verified target handle open while unlinking the friendly source, and chmod that handle to `0600`. Safe prior targets are idempotent results; exact interrupted two-link transfers recover. Source swaps, destination swaps, and Linux inode reuse fail closed.
- Runtime parents are `0700`; adopted generated files are regular, single-link, and `0600` before hashing, approval, or persistence. Ordinary vault-file modes are never tightened through this path.
- Quiescent checkpoint cleanup validates the complete flat inventory and live outbox state before deleting anything. Active descriptors must match ref, content type, size, and SHA-256 exactly. Active hardlinks fail closed; an orphan hardlink may remove only the runtime-owned link. Unsafe paths, entries, or live inventory retain the staging set.
- This adds no schema, snapshot format, registry, sidecar, lifecycle manager, queue, or second state owner.

## Audit remediation on the resolved head

The post-conflict parent audit found and fixed five concrete problems in the prior implementation:

1. `send_vault_file` was omitted from the existing serialized tool chain, allowing overlapping media mutation.
2. Exclusive adoption used a clobberable check-then-rename sequence.
3. Staging hardlinks could chmod an ordinary vault inode and make active checkpoints fail unpredictably.
4. Missing semantic provider call identity could alias distinct same-turn generated sends.
5. The destination could be swapped after link-pair validation but before chmod; Linux delete/recreate could also reuse the same inode number and defeat a reopen-and-compare repair.

The repair reuses the runtime-state file owner and existing dynamic-tool serialization. A proposed permanent quarantine registry was rejected: malformed live inventory already quarantines and retains in that pass, while a later trusted quiescent pass can prune a proven orphan. Adding durable registry state would create a second source of truth without a product requirement.

- `packages/runtime-state` owns no-clobber adoption, interrupted-transfer recovery, link-count enforcement, and destination identity through chmod.
- `packages/assistant-engine` owns generated-ref identity, serialized sends, exact media reads, outbox binding, and fail-closed residue cleanup.
- The hosted checkpoint bridge opts into cleanup only at the existing quiescent boundary; active files remain restart-safe in encrypted checkpoints.
- Hosted-local approval/resume coverage exercises pending approval, checkpoint, container loss, cold restore, and attachment delivery without another provider request.
- The current-main runner budget remains a `7,500,000`-byte static-closure baseline plus `96,000` bytes tolerance and `250,000` bytes explicit headroom. The measured resolved bundle is `7,531,562` bytes static closure, `1,549,429` bytes entry, and `9,238,580` bytes total.

## Verification

- Affected package/app typechecks: passed.
- Assistant-engine canonical coverage: 167 files passed, 1 skipped; 2,507 tests passed, 5 skipped.
- Runtime-state final coverage: 27 files and 192 tests passed after pinning the verified target handle across unlink/chmod.
- Focused vault-send/residue coverage: 51 tests passed; overlapping-send regression passed.
- Runner-bundle ratchet: 28 tests passed; full bundle assembly passed within all budgets.
- Scenario integrity: 204 scenarios, 11 sample inputs, and 28 golden-output directories passed.
- Full `verify:acceptance` cleared syntax, dependency policy, package boundaries, runtime/security guards, all workspace typechecks, documentation hygiene, and every changed-owner coverage suite. Its aggregate parallel lane reported two untouched load-sensitive tests: a core receipt stress case timed out at 60 seconds and a setup-wizard TTY selection mismatched. Their exact isolated reruns passed in 28 seconds and 158 milliseconds respectively.
- Post-merge onboarding/system-prompt overlap: 7 tests passed.
- Diff check and identifier/privacy scan: passed.

## Deployment skew and rollback

- No Vercel/web deployment, database migration, secret change, or tandem release is required.
- PR #768 remains the rollback floor while a checkpoint or active outbox descriptor can contain the hidden runtime ref.
- Deploy the Cloudflare writer only after protected-main gates and exact runner fingerprint convergence. Older warm runners may retain residue for one additional checkpoint; new readers accept either checkpoint shape.

## Change shape

| --- | ---: | ---: |
| Source | 1,015 | 52 |
| Tests/fixtures | 1,699 | 144 |
| Docs | 366 | 45 |
| Config/tooling | 6 | 6 |
| Generated/other | 0 | 0 |
| **Total** | **3,086** | **247** |

## Round 4 anomaly retrospective

**Original requirement.** Remove terminal one-time generated-delivery residue at the existing quiescent checkpoint boundary while preserving approval, active retry/restart recovery, cold delivery, and ordinary vault-file safety. Do not add a user flow, schema, registry, queue, lifecycle manager, or second owner.

**Immutable first-reviewed shape (`97d6dc37d77df60314ff71b07f724b964333cd45`).** Source 619/46, tests 892/112, docs 203/42, config/tooling 6/1; total 1,720/201.

**Current shape.** Source 1,015/52, tests 1,699/144, docs 366/45, config/tooling 6/6; total 3,086/247. Authored-source churn moved from 665 to 1,067 lines: +396 additions and +6 deletions. This is below the 500-line remediation trigger; the retrospective is required because this is substantive round 4.

**Attribution of production growth.** The first head owns the cleanup planner, exact active-descriptor matching, quiescent snapshot integration, permission adoption, and cold approval/resume path. Round 1 replaced the shared friendly filename as physical identity with a per-send owned ref, preserved the display filename separately, and bound cleanup to the exact physical descriptor after the same-name collision finding. Round 2 removed the interim random identity, derived the owned ref from the provider call fact, threaded that fact through existing tool context, and added exact-call interrupted adoption after the retry-identity finding. Round 3 added no new owner and narrowed the promise to the documented attempt-scoped limitation. The resolved-head parent audit then added one case to existing tool serialization, rejected missing semantic identity, and replaced check-then-rename with runtime-state-owned no-clobber transfer, single-link policy, and hardlink-aware cleanup. The exact-head Linux CI failure proved that reopen-plus-dev/inode comparison still allowed immediate inode reuse; the current delta holds the already-verified destination handle across unlink/chmod instead.

**Concept and owner accounting.** Added concepts are semantic call identity, serialized generated sends, deterministic owned refs, atomic link/unlink transfer, exact interrupted two-link recovery, single-link enforcement, and active-versus-orphan hardlink cleanup. Removed concepts are random UUID identity, shared friendly-name physical ownership, clobberable check-then-rename, catch-all target acceptance after arbitrary failure, and the false promise of replacement-call retry stability. Owners did not grow: runtime-state still owns file adoption, assistant-engine still owns tool/outbox/residue behavior, and the snapshot bridge still invokes cleanup only at its existing quiescent boundary.

**Decision: justified continuation, with no split or new machinery.** Deletion cannot preserve both portable no-clobber creation and identity-safe permission tightening: Node rename overwrites, while atomic link plus an open verified handle is the smallest cross-platform primitive that survives the reproduced races. Interrupted-link recovery is the unavoidable completion case of that two-operation transfer. Serialization reuses an existing chain, semantic identity is derived from the provider fact, and cleanup derives entirely from existing outbox plus live inventory. Conflict-only base history is excluded from the PR patch; retaining current-main runner baseline/tolerance/headroom is policy preservation, not generated-delivery behavior. The deliberate limitations remain: provider identity is attempt-scoped, and this PR will not add a durable send fact, registry, lifecycle manager, replay mechanism, or second cleanup owner.

**Round 4 outcome.** ReviewGPT checked `66cc8761f9f3a0c3d5c5d88e863c9ae91ff87881` and returned `RETROSPECTIVE_REQUIRED` solely for the mandatory round-number gate; it emitted no tactical correction finding. This retrospective satisfies the requested accounting before any later correction-verification round.

ReviewGPT first-reviewed head: 97d6dc37d77df60314ff71b07f724b964333cd45

ReviewGPT previous substantive head for round 4: 6a6593e78ed19344f18abf262a5bf7a6537204a3


rocketman-21:
Audit lineage note: a draft-stage ReviewGPT round ran against the pre-redesign head 32e89c8412 and returned one accepted High finding — reserving `exports/assistant-deliveries/**` (ordinary vault space) as cleanup authority created a deterministic data-loss hazard for pre-existing files at that path. The accepted remediation was the reviewer's own smallest-safe-correction: stage one-time generated deliveries under the runtime-owned `.runtime/operations/assistant/generated-deliveries/` subtree via an expand-then-produce rollout. Phase one (reader compatibility, writer closed) landed and deployed in PR #768 with fingerprint convergence; this PR is the phase-two producer. `exports/assistant-deliveries/**` retains no ownership, deletion, or packaging semantics on this head. A fresh full-patch round is running against the current head as the substantive audit of the redesigned patch.

rocketman-21:
Round 1 result recorded: one accepted High. The shared per-vault staging namespace let a later turn that reused a friendly staging name (e.g. `report.pdf`) overwrite the bytes of an earlier still-pending delivery, and cleanup's `some()` match hid the resulting conflict so an approved delivery could terminally fail after its bytes were gone.

Remediation (head 7c78b4c2a863072377847f0437da6528f5b39c21, commit 7c78b4c2a8), following the reviewer's smallest-safe-correction and adding no registry/state-owner/service:
- `send_vault_file` now consumes the model-selected staging file into a runtime-generated collision-free flat ref (`<stem>-<uuid>.<ext>`) via a same-directory `rename` before hashing or approval; the user-facing basename is preserved in the descriptor's `filename`. New `adoptAssistantStateFileIntoExclusiveName` performs the O_NOFOLLOW adoption then fails closed if the exclusive target already exists.
- Cleanup no longer binds on descriptor filename (now runtime-owned); it matches on the physical ref's content-type + exact size + SHA-256, requires every active descriptor sharing a ref to agree on that fingerprint (fail-closed, retain-all on conflict), and retains a file only when it matches every such descriptor.
- Regression coverage: two sequential pending sends reusing one staging name now persist distinct owned refs with distinct bytes and both verify; conflicting active descriptors for one ref skip cleanup and retain all files; exclusive-name adoption rejects clobber, cross-directory, and symlink sources. Full-stack approval-resume E2E still green.

Round 2 will review only this remediation delta and its directly affected paths.

rocketman-21:
Round 2 outcome: RETROSPECTIVE_REQUIRED (anomaly), with a High REVIEW_INDUCED finding against the round-1 remediation itself.

Finding: the consume-into-owned-name step renames the model-written staging file to a random-UUID ref *before* a durable outbox descriptor exists. If `requestAssistantVaultFileSend` is interrupted after that rename but before `createAssistantOutboxIntent` persists (approval-port HTTP timeout, lost response, process kill, or outbox-write failure), the model-known ref is gone, no durable record maps the logical send to the UUID file, and the next quiescent checkpoint prunes it as an orphan. The prepared one-time artifact is then unrecoverable without regeneration, and a remotely-committed-but-lost approval response cannot be re-derived. This fixes cross-send uniqueness (round-1 High) but introduces same-send retry/restart instability — the reviewer's point is that physical-storage naming is still standing in for a stable *logical* send identity.

The reviewer asks for a requirement-level retrospective before any further tactical patch: define one logical send identity that is unique across distinct sends and stable across retry/replay/ambiguous-approval/restart, and derive the physical owned ref from that stable key (from an existing identifier, not new lifecycle state), so an interrupted send can re-locate and validate already-owned bytes and recover the same approval identity.

Note: this is scoped to the narrow pre-outbox-persist interruption window and concerns freshly-generated one-time artifacts (regenerable), not stored user/vault data. The persisted-outbox retry path (post-`createAssistantOutboxIntent`) remains safe and is covered by the E2E. Pausing the loop for a direction decision rather than applying a third tactical patch, per the change-shape retrospective rule.

CI status: the runtime-checkpoint-durability E2E (the shard that directly exercises this change) is green. The 5 red checks (Codex media / Linq reminder / Telegram E2E, Release app verification/checks) all reproduce on origin/main's own tip 8331f28d9a and are upstream: main's gpt-5.6-terra default swap zeroed tool advertisement in the stub E2Es, and the Murph Safe apps/web unit test fails on main independently. This branch touches no apps/web files.

rocketman-21:
Round-2 retrospective remediation applied (head 6a6593e78ed19344f18abf262a5bf7a6537204a3, commit 6a6593e78e), authored via a gpt-5.6-sol redesign pass and verified locally:

- The owned physical ref is now DETERMINISTIC: `buildAssistantGeneratedDeliveryOwnedRef` derives the basename from `sha256([sessionId, turnId, toolCallId, ref])` instead of `randomUUID()`. Same send (same session/turn/tool-call/ref) re-derives the same ref (retry-stable); distinct tool calls reusing friendly `report.pdf` get different refs (cross-send unique).
- The stable `toolCallId` (already on `ParsedDynamicToolCallRequest`, previously dropped) is threaded through `readMurphDynamicToolRequest` → `executeMurphDynamicToolRequest` → `sendVaultFile(ref, toolCallId)` → hosted-tool-context → local-service → `requestAssistantVaultFileSend`.
- Adoption is now idempotent: `adoptAssistantGeneratedDeliveryIntoStableName` first tries adopting the deterministic target in place (recovering an interrupted attempt's already-owned bytes), and only falls back to the exclusive rename when the target is absent. An interrupted send (approval-HTTP failure / crash before outbox persist) is now recoverable — the retry re-derives the same ref, reuses the same bytes, and rebuilds the same approval identity.
- No new lifecycle state, registry, sidecar, or reconciliation loop; outbox schema and snapshot format unchanged.

Regression proof added: two distinct sends reusing `report.pdf` keep separate bytes; the same send retried after an outbox-persist interruption reuses one owned file and one approval identity (approval request bytes identical across both attempts). Full assistant-engine suite (2,453 passed) and the full-stack approval-resume E2E green; runner-bundle budget re-ratcheted by the exact 1,284B measured overage.

Round 3 (remediation-delta verification) starting now.

rocketman-21:
Round 3 (valid, remediation-delta scope): RETROSPECTIVE_REQUIRED again, with a High REVIEW_INDUCED finding that invalidates the round-2 fix's premise.

The catch: `toolCallId` is a provider *attempt* identity, not a logical *send* identity. When `send_vault_file` fails and the error is returned to the model (per RELIABILITY.md), the model recovers with a NEW provider function call carrying a NEW call_id — not an exact replay. The App Server stubs prove this (assistant-codex-scripted-runtime.test.ts:781-787 and hosted-local-e2e-support.ts:812-899 mint a fresh call_id per Responses request). So: call A hashes to owned-ref A, renames staging → owned-ref A, approval HTTP fails; model recovers with call B → owned-ref B (doesn't exist), staging gone → adoption ENOENT, nothing persists, and owned-ref A is pruned as unclaimed at the next checkpoint. Same stranding as the UUID version — a deterministic hash of an attempt id is still an attempt id.

The round-2 regression test is insufficient: it re-invokes with the SAME toolCallId, proving exact-RPC replay only, not the real A→B recovery path.

Reviewer's requirement-level decision (correct): no local hash of currently-available inputs can satisfy this, because a recovery call and an intentional second send can BOTH carry the same session/turn/ref with a new call_id. It must be decided whether identical sends coalesce (key on a logical fact like content, owned before the destructive move) or whether a durable send identity is genuinely required — without adding a registry/sidecar/reconciliation loop.

Scope/impact: reachable only on the approval-failure recovery path; strands a freshly-generated (regenerable) one-time artifact, not stored user data. The persisted-outbox retry path (after the intent is saved) remains safe.

Pausing the automated loop for a requirement decision rather than applying a third tactical patch, per the stop-patching rule (00-invariants.md:33-38).

rocketman-21:
Decision recorded: accept the round-3 narrow risk rather than pursue a further redesign or revert. The attempt-scoped generated-send identity limitation is now documented (commit 56f8d7f3bf) in agent-docs/references/hosted-runtime-protocol.md, at the consume site in vault-file-send.ts, and in the PR body's 'Known limitation (accepted)' section; the retry regression test is rescoped to accurately cover only exact-call-id re-delivery. The ReviewGPT remediation loop is stopped here by decision (not a zero-finding pass). Merge remains gated on CI, which is currently red only from upstream main breakage (gpt-5.6-terra tool-advertisement E2Es + Murph Safe apps/web test), not this PR.

PR #816, opened 2026-07-21
Normalize equivalent lab biomarkers

rocketman-21:
## Why

Equivalent lab analytes currently split into separate histories when one source uses an abbreviation and another uses the expanded name. That makes BUN, TSH, MCH, and MCHC appear duplicated and obscures their longitudinal history.

## User-visible goal and UX

- Combine explicit BUN, TSH, MCH, and MCHC aliases into one biomarker history.
- Keep generic urea and BUN/creatinine ratio as distinct measurements.
- Show device-derived biomarkers before lab content, including before the lab-empty state.
- Keep lab health-area sections closed until the member opens them.
- Frame the page accurately for both device and lab sources.

## Invariants

- Use explicit aliases only; no fuzzy medical matching.
- Preserve each original reported value and unit in history rows.
- Normalize only comparable chart values; convert BUN mmol/L to mg/dL and treat equivalent TSH concentration units as mIU/L.
- A shared metric consumer may use a canonical value/unit pair or a finite raw value with an explicitly reported raw unit. It must never attach a catalog display unit to a unitless numeric result.
- Keep unitless lab observations visible as raw history without making them selectable, chart-comparable, or usable by goals or experiments.
- Rebuild persisted query projections when catalog semantics change; do not rewrite source vault data or widen the browser replica.
- Use native disclosures without a new UI state owner or dependency.

- The shared health-metric catalog controls metric and biomarker identity in browser-vault projection, list/detail routes, and persisted or hosted-portable query SQLite projections. Query projection version 18 forces unchanged-manifest v17 stores to rebuild with the canonical identities and values.
- Shared metric selection and series construction own typed comparability for generic rows, goals, and experiment analysis. They now use one canonical-or-explicitly-reported value/unit rule rather than independent display-unit fallbacks.
- Experiment primary-outcome admission, readiness, analysis, and old anchored-point retention consume the same catalog. Anchor aliases are canonicalized once before the 365-day browser retention filter.
- Metric goals consume the catalog for canonical identity, explicit unit comparison, latest-lab selection, and the existing 365-day lab staleness policy.
- Page-level source framing is neutral because the header count and first section may include device metrics.

## Verification

- Health-metrics suite: 64 passed.
- Full query package suite: 559 passed across 58 files. The three CLI regression files that exposed metric-slug fallback loss also pass (91 tests with the health regression file).
- Focused browser regressions: lab aliases 9 passed; metric goals 10 passed; experiment results 86 passed; browser replica 14 passed; projection/runtime 94 passed.
- Biomarker UI focused tests: 26 passed on the earlier frontend head; the review remediation does not change frontend files.
- Health-metrics, query, and hosted-web typechecks passed.
- Dependency, workspace-boundary, Temporal, crypto, and raw-health-log guards passed in the affected-scope lane.
- The prior affected-scope lane reached 2,515 passing tests before an unrelated assistant-engine Vitest worker exhausted its 4 GB heap; all exact-head required CI checks passed on the current head.
- Diff/privacy checks passed. Coverage-write and remediated frontend-review passes have zero unresolved findings.
- Browser backends were unavailable for desktop/mobile visual proof.

## ReviewGPT round 1 dispositions

- Accepted projection-version finding: bumped the query SQLite version to 18 and added an unchanged-manifest v17 upgrade regression that proves aliases rebuild into one canonical history.
- Accepted unitless-lab finding: unitless test results retain raw value and identity but leave canonical value/unit null.
- Accepted shared-catalog purpose finding: disclosed experiment and goal consumers and added focused alias regressions for admission/analysis, goal selection, conversion, and freshness.

## ReviewGPT round 2 retrospective and disposition

Round 2 returned `RETROSPECTIVE_REQUIRED` because generic selection/series could still attach a display unit to the raw unitless value, and old experiment-anchor retention still compared the raw alias after projection canonicalization.

The recorded retrospective chose a same-PR owner-bound redesign: one shared canonical-or-explicitly-reported value/unit rule for selection and series, plus catalog canonicalization at the existing experiment-anchor retention boundary. This removes competing authority without adding a state owner, migration service, compatibility path, queue, or reconciliation mechanism. The hand-built experiment regression was deleted and replaced with canonical vault source records through real projection, browser-replica retention, goal, selection, and experiment-analysis paths. The proof includes newer unitless and older unitful BUN, TSH, MCH, and MCHC records, an alias-anchored BUN baseline older than 365 days, runtime/browser agreement, and the v17 unchanged-manifest rebuild.

## Exact-head CI regression disposition

The canonical-key remediation exposed one valid metric class that has a catalog definition but no biomarker mapping. Active experiments using `biomarker:sleep-efficiency` were incorrectly rejected after assessment re-resolved only through the biomarker map. The current head keeps canonicalization and definition resolution in one owner-bound helper, preserves metric-slug fallback, and adds a direct regression alongside the three formerly failing CLI files.

## ReviewGPT round 3 disposition

Round 3 independently found the same catalog-only metric regression exposed by exact-head CLI coverage. The finding was accepted: canonical biomarker identity and metric-definition fallback now resolve together, so definitions without a dedicated biomarker-map owner remain valid experiment outcomes. Focused proof covers `body-weight`, `sleep-efficiency`, and `sleep-score`, retains canonical BUN behavior, and reruns the three affected CLI files through real experiment start/edit paths.

## ReviewGPT round 4 result

The trusted exact-head correction audit returned `ROUND_OUTCOME: PASS` with zero qualifying findings after validating the packaged PR lineage, full diff, cumulative delta, immediate remediation delta, and the experiment admission/readiness correction. Model verification recorded `gpt-5-6-pro`. Earlier same-round attempts were discarded as invalid because their invocation omitted the guarded PR-context packaging environment; they did not count and made no code changes.

## First-review change shape

| --- | ---: | ---: |
| Source | 91 | 13 |
| Tests | 198 | 4 |
| Docs | 60 | 0 |
| Config/tooling | 0 | 0 |
| Generated/other | 0 | 0 |

ReviewGPT first-reviewed head: ae2cf30bc74fe92abf5a380467c7cbc4729d9014

## ReviewGPT current head

| Head | Source + | Source - | Tests + | Tests - | Docs + | Docs - |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| `343fbf7f453fd8e477464c030a5362b02d1d5dd7` | 208 | 56 | 634 | 6 | 60 | 0 |

The current head includes the recorded owner-bound redesign, production-path regression replacement, and metric-slug fallback correction described above.


rocketman-21:
ReviewGPT round 2 retrospective decision

Original requirement: collapse explicit BUN, TSH, MCH, and MCHC aliases into one longitudinal lab history while keeping generic urea and BUN/creatinine ratio separate, putting device metrics first, and closing lab groups by default.

Shape comparison: the immutable first-reviewed head was source +91/-13, tests +198/-4, docs +60/-0. The current reviewed head is source +110/-17, tests +490/-6, docs +60/-0. Review-driven growth from the first head is source +19/-4 and tests +298/-8. The source growth added the v18 projection rebuild, the scalar test-result missing-unit guard, and explicit catalog biomarker aliases. It added no new state owner, queue, migration service, or compatibility path.

Repeated mechanism: scalarMetricPoint now leaves canonical fields null for a unitless lab, but health-metrics selection and series still retain independent raw/display-unit fallbacks. Browser replica experiment-anchor retention also compares raw biomarker keys after metric points have already been canonicalized. Those retained owners can recreate invented comparability or miss old alias-anchored points.

Decision: redesign and continue in this PR because goals, experiments, charts, and lab detail all consume the same shared metric catalog and must not disagree about these aliases. MetricPoint canonicalValue/canonicalUnit will be the single comparability boundary for catalog metrics; selection and series will stop falling back to raw values when a canonical unit is defined, while raw values remain available only to provenance/lab-history presentation. Experiment anchor biomarker keys will be canonicalized once through the existing catalog before retention matching. This removes competing authority without adding state or reconciliation machinery. The v18 rebuild remains. Regression proof will use canonical vault records through the real projection/browser-replica builders for newer unitless plus older unitful BUN, TSH, MCH, and MCHC, goal/experiment non-invention, runtime/browser agreement, and alias anchors older than the 365-day lookback.

rocketman-21:
Retrospective evidence refinement: the focused compatibility suite showed that an explicitly reported but non-canonical unit must remain a typed value (for example, a unit the catalog cannot convert). The continuation decision is therefore implemented as one shared rule: use the canonical value/unit pair when available; otherwise use a finite raw value only when the source also reported a non-null raw unit; otherwise the point is not selectable or series-comparable. No consumer may attach definition.displayUnit to an untyped numeric value. This preserves existing unsupported-unit warnings and raw provenance while eliminating the unitless fallback mechanism identified in round 2.

PR #839, opened 2026-07-22
Restore experiment daily trends without rewriting saved outcomes

rocketman-21:
## Why this PR exists

Completed experiment reports regressed to one flat window-average point even when the encrypted browser replica still contained daily measurements. The same closeout path could also analyze fresh evidence before resolving an already referenced write-once result. This PR restores the daily series and makes saved-outcome retries reference-first without adding a migration, provenance ledger, or second state owner.

Completed experiment reports show every available baseline and intervention day again. The affected private fixture now projects 19 points for every data-bearing metric instead of a single flat average. Hover dates stay on the correct calendar day in non-UTC time zones.

## User experience

- Current v2 outcomes render their immutable saved daily-point snapshots and saved window averages.
- Legacy v1 outcomes remain unchanged and point-free. The results read model may pair their saved averages with current browser-safe daily measurements selected strictly by the outcome's saved biomarker identities and saved windows.
- Unsupported or genuinely data-less metrics keep the existing aggregate-only fallback.
- Copy remains neutral about saved/current provenance, and completed reports suppress mutable pre-run history.

## Invariants

- A valid referenced outcome artifact is write-once; repeated closeout or delivery work resolves and returns that exact artifact before attempting fresh analysis.
- Missing, mismatched, or misaddressed references fail closed.
- Terminal closeout may advance an active or on-time-completed interim reference once to a distinct final artifact without overwriting either artifact; an early-stopped run never advances.
- A later run-plan edit cannot move an immutable interim's saved terminal horizon earlier.
- Final-results automation proceeds only when the write-once owner returns a completed canonical outcome; an active interim is a deterministic pre-provider skip.
- Existing completed status and `endedOn` metadata remain authoritative during final outcome publication.
- v2 daily snapshots remain frozen and always win over conflicting live rows.
- v1 artifacts are never upgraded, rewritten, or represented as containing historical daily evidence they did not save.
- Legacy display rows cannot be widened by post-save run plans, analysis selectors, session fields, anchors, extra biomarkers, or out-of-window data.
- Web remains compatible with both v1 and v2 while Cloudflare is held until the corrected Web reader is live.

## Review findings resolved

- **Unsafe legacy saved-time reconstruction — accepted, redesigned:** removed legacy outcome upgrading and the duplicated reconstruction/provenance machinery. v1 stays immutable and aggregate-only in storage and replica; results may overlay current browser-safe rows bounded by the artifact's own biomarker identities and windows. Focused production-path proof confirms 19 displayed points for every data-bearing metric, while v2 tests prove saved snapshots beat conflicting live data.
- **Analysis before reference resolution — accepted:** added a core reference resolver and call it inside the canonical write lock before analysis. Ordinary title, status, run-plan, analysis-plan, anchor, and session-field edits return the exact referenced artifact; only an interim result reaching its terminal horizon is eligible for one distinct final artifact.
- **On-time stopped interim finalization — accepted and reproduced:** the terminal decision incorrectly required the current run to remain active, so a run stopped exactly at its intervention end could return its interim as final. One core-owned pure predicate now governs both use-case preflight and core write behavior: active or on-time-completed runs advance once, early-stopped runs remain immutable, and completed status plus `endedOn` are preserved.
- **Mutable plan could reclassify an early stop — accepted and reproduced:** shortening the current run plan after an early stop could make the stop appear on time. The same core predicate now uses the later of the current plan end and the immutable interim's saved intervention end. A supported edit-path regression proves analysis is not invoked and the exact interim bytes and reference remain unchanged.
- **Active interim could authorize a final-results message — accepted and reproduced:** the writer correctly returned the unchanged active interim, but assistant automation treated any successful write as terminal publication. The existing final-results precondition now continues only when the returned canonical outcome status is `completed`. A real-writer regression covers interim save, early stop, supported plan shortening, precondition skip, and exact artifact/reference preservation; the existing cron boundary proof confirms a skip never admits the provider.
- **Frontend provenance and date review — accepted:** replaced ambiguous source wording with neutral copy and made trend-day arithmetic explicitly UTC-stable. A New York regression proves day 1 remains the saved calendar date.

- Core experiment closeout owns reference validation, terminal advancement, and write-once artifact semantics.
- Vault use cases resolve the current canonical experiment revision and referenced outcome inside the canonical write lock before invoking the analyzer.
- Browser replica construction parses legacy outcomes unchanged; results projection owns the narrow display-only overlay.
- Assistant maintenance no longer refreshes linked outcomes; final-results delivery trusts the canonical returned outcome status rather than mutable plan dates.
- The trend tooltip formats date-only values without local-time-zone drift.

## Change-shape breakdown

| --- | ---: | ---: |
| Source | 280 | 336 |
| Tests / fixtures | 975 | 361 |
| Docs | 3 | 3 |
| Config / tooling | 0 | 0 |
| Generated / other | 0 | 0 |
| **Total** | **1,258** | **700** |

ReviewGPT first-reviewed head: 2f4dce9c39e6bcc435924f551d9cbb999aee861e

| Review round | Current head | Source added | Source deleted |
| --- | --- | ---: | ---: |
| 1 | `2f4dce9c39e6bcc435924f551d9cbb999aee861e` | 162 | 309 |
| 2 | `4be8733b39cc3d11db6298f0e66ec698daa6b891` | 256 | 334 |
| 3 | `f017cf132aea5345e7fccfffeb6c15c174e15b2e` | 266 | 335 |
| 4 | `da08e150b669de010703ae1dd5a1d51ee5b72655` | 273 | 335 |
| 5 | `b361d041c38807072c1f78954147f14ae0806e9a` | 280 | 336 |

## Review retrospective

The original requirement remains one indivisible reader-and-writer invariant: show daily evidence while preserving write-once outcomes and never announce a final review from an active interim. Current authored-source churn is 616 lines, well below the 2,000-line threshold. Review remediation centralized terminal policy in one core-owned predicate and added one assistant-boundary status branch; it added no state owner, lifecycle, queue, migration, or repair path. The explicit decision is to continue this PR because splitting would separate the compatible reader, writer invariant, and user-facing publication gate that must agree.

Deploy the corrected Vercel Web reader first, confirm the production health endpoint and private report, then immediately deploy the Cloudflare writer. New Web reads both v1 and v2. Cloudflare remains paused until Web is confirmed because old Web cannot read newly written v2 outcomes. After Cloudflare deploys, smoke the private report and monitor outcome parse failures.

## Verification

- `pnpm test:diff packages/assistant-engine packages/vault-usecases packages/core`
- `pnpm verify:acceptance`
- Full Web suite: 489 files / 6,126 tests passed
- Assistant-engine suite: 171 files / 2,580 tests passed
- Query owner suite: 575 tests passed; vault-usecase suite: 210 tests passed; core suite: 731 tests passed
- Cloudflare verification: 1,852 tests passed
- Focused terminal-outcome proof: active and stopped-on-time advancement, unchanged and edited-plan early-stop immutability, completed metadata preservation, interim byte immutability, retry idempotency, and final-results pre-provider skip for a canonical active interim
- Required coverage-write and frontend-review passes; independent fallback frontend review returned `NO FINDINGS`
- Private export replica proof: 19 points for every data-bearing metric; no private values persisted
- `git diff --check` and scoped privacy scan


rocketman-21:
ReviewGPT round-3 anomaly retrospective: the original requirement is to restore daily experiment trends while preserving write-once outcome artifacts and a single closeout owner. The immutable first-reviewed source shape was +162/-309; the current source shape is +266/-335. Review remediation added 104 source lines and deleted 26, principally to remove unsafe legacy reconstruction, make reference resolution precede analysis, and centralize terminal interim advancement. The repeated closeout mechanism is now collapsed into one core-owned pure predicate used by both vault preflight and core write; no new state owner, lifecycle, queue, migration, or repair path was added. Current authored-source churn is 601 lines, well below the 2,000-line trigger. Decision: explicitly continue this indivisible PR because splitting or reverting would separate the reader compatibility fix from the writer invariant it depends on, while the current design is smaller than the first attempted remediation and has production-path regression proof. The first round-3 tool attempt completed in 369 seconds and was rejected as untrusted by the mandatory 10-minute guard; it is a tooling retry only and does not advance the substantive round.

rocketman-21:
ReviewGPT round 4 finding accepted and reproduced. The canonical writer preserved the active interim exactly, but the assistant precondition treated successful resolution as terminal publication. The pushed remediation adds one status branch at the existing final-results boundary: only a returned completed outcome may continue. A real-writer regression now proves the shortened early-stop path skips while bytes and reference remain exact; the existing cron boundary proof confirms the provider is not invoked. Canonical diff verification is green; acceptance and final review are running against this head.

PR #840, opened 2026-07-22
fix(runtime): resume completed assistant asks

rocketman-21:
## Why

A joined-group Ask could finish successfully and reach the originating personal
runtime, yet remain pending with zero attempts. The same incident showed about
2 minutes 20 seconds of end-to-end group-read latency: roughly 2 minutes 13
seconds waiting behind the dirty runtime's normal idle-checkpoint window and
about 7 seconds of actual group execution.

## User-visible goal and UX

1. The member asks private Murph to consult a joined group and immediately gets
   an accepted acknowledgement.
2. The existing separate, read-only group process starts promptly even when the
   group runtime is dirty. This does **not** publish, advance, or shorten the
   normal idle checkpoint.
3. When the answer returns, it automatically wakes private Murph; the member
   does not need to send another message.
4. The group answer is untrusted context for the existing output-only
   continuation. Private Murph composes a natural response and queues it through
   the ordinary idempotent outbox. The raw group answer is not sent verbatim
   automatically.
5. If an Ask completion occurred before already-pending personal input, the Ask
   continuation owns the next assistant pass and the personal input remains
   queued for the following pass. A later completion cannot overtake earlier
   personal input.

## How it works

- Joined-group request admission reuses the existing pre-checkpoint-safe mailbox
  prefix. One shared import context carries the joined-group-only restriction
  through every import in that pass, including pre-assistant, follow-up, and
  active-turn conversation/system imports. The distinct consented-member flow
  remains checkpoint-gated.
- Completion ordering reuses the existing foreground-causal mailbox lane. A
  fresh turn derives its cutoff from the oldest `occurredAt` in the bounded
  accepted-input batch already owned by that pass. A pass without fresh input
  reads only the existing complete pending-input index. Missing, incomplete, or
  invalid evidence fails closed without foreground backfill or compaction.
- Pre-checkpoint causal-only passes admit only the runtime-control and
  pending-effects work they are designed for; Ask completions wait for the
  ordinary cutoff-aware pass.
- Auto-reply preparation stabilizes the existing foreground watcher and flushes
  already-imported member-channel updates locally before dispatch. It does not
  refetch the remote system prefix or convert unrelated checkpoint-gated work
  into a current-reply delivery barrier. Provider entry re-reads the existing
  local automation-state authority.
- The existing continuation and outbox retain natural composition, stable
  idempotency, delivery, and retry ownership.
- A metadata-only warning is derived when a completion's first claim is at least
  60 seconds late. No alert state is persisted.

## Invariants

- The mailbox remains durable transport, not an Ask-specific lifecycle
  coordinator.
- The detached Ask controller remains the sole group-read process owner.
- The ordinary idle/shutdown checkpoint schedule is unchanged.
- Only `joined_group` requests may use the narrow dirty-window fast path;
  `consented_member` requests retain ordinary checkpoint timing and
  reviewed-exact response policy.
- A joined-group answer is model context for private Murph, never implicit
  exact-send copy.
- Ordering ends at durable intent creation. An unrelated cross-turn carrier
  retry does not freeze newer personal replies.
- No schema, queue, scheduler, polling loop, receipt state machine, dependency,
  or persisted state was added.

- Every pre-checkpoint importer, including the active-turn foreground loop, can
  observe a new system row. The joined-group target restriction must therefore
  travel with the shared pass context instead of wrapping one import call.
- Personal input occurrence and receipt are distinct timestamps. Only
  occurrence defines causal order.
- A complete pending-input index may retain a terminal stale first entry until
  background compaction. Fresh turns therefore use their bounded pass-owned
  accepted-input batch; only no-fresh-input passes use the complete-index
  fallback.
- The pending-input index has a separate background maintenance owner.
  Foreground reply admission may inspect existing evidence but must not repair
  it.
- Causal-only passes intentionally avoid pending-input resolution, so they
  cannot safely admit Ask completions.
- Remote system-mailbox catch-up is not a current-reply authority boundary. An
  update already imported into the local mailbox is selected ahead of queued
  Ask work and flushed before delivery; a remote update behind a non-admitted
  Ask remains transport work until the ordinary checkpoint. The current reply
  deliberately continues from last-known-good local authority instead of
  waiting up to the idle-checkpoint window.

## Review implementation-direction retrospective

The immutable first-reviewed head is
`07ba1255d27e494b01c744d69af30899009c7dd1`. It changed 117/29 authored-source
lines across two files, 346/10 test lines across three files, and 111/2
docs/plan lines. Config/tooling and generated/other were both 0/0.

The current patch at `8e2d2cadaed345ecb786df93082352bb20469e80`
changes 281/157 authored-source lines across six files, 846/182 test lines
across six files, and 439/2 docs/plan lines. Config/tooling and generated/other
remain 0/0. Relative to the immutable first-reviewed patch, review remediation
grew authored source by 164 additions and 128 deletions, moving source churn
from 146 to 438 lines; tests grew by 500 additions and 172 deletions.

Rounds 1-5 grew an increasingly complex retained-row/coordinator design to 597
lines of authored-source churn. The completed retrospective chose deletion and
redesign around existing owners. The current patch removes the coordinator,
retained-row lanes, global cross-turn delivery barrier, duplicated private wake
rules, Ask-specific lifecycle machinery, and the later remote pre-auto-reply
catch-up barrier. It retains only shared mailbox import context, existing
foreground-causal selection, pass-owned accepted-input evidence, the read-only
complete-index fallback, local channel authority, and the ordinary continuation
and outbox. Continuation in this direction is explicitly justified because it
preserves the user outcome with no new state, queue, scheduler, lifecycle, or
owner and remains substantially smaller than the rejected coordinator design.

The raw `since-first-reviewed-head.diff` also contains base-history merges made
after the immutable first-reviewed commit. In the invalid first round-10
attempt, 234 paths in that raw Git-history delta were absent from both the
current PR diff and `changed-files.txt`; those paths are base movement, not
review remediation. They do not reset or replace the immutable baseline. The
patch-shape comparison above uses the original reviewed patch and the current
PR patch. Round 12 reviewed the clean round-11 remediation delta from
`7c3c53426211aaf0e0e08d852c214a3e967eeabd` to
`97d8aa58309dd95ebcd737858f8fcdb391c7e176` and passed with zero findings.
Round 13 reviews the normal current-main merge plus its single manual conflict
resolution. Runtime source merged automatically; the documentation-index
resolution preserves both PR #840's Assistant Ask contract and main's
stale-fence diagnostics contract without changing executable behavior.

## ReviewGPT history

| Round | What it found | Accepted / rejected | Complexity impact |
| --- | --- | --- | --- |
| 1 | Generic causal selection could release personal work without owning the exact older Ask; dirty-window import scope was too broad. | Accepted. | Added a causal barrier and bounded prefix; meaningful complexity. |
| 2 | One invalid older completion could let another be overtaken; private wake logic duplicated outbox semantics. | Accepted. | Added full-set convergence while deleting duplicate wake logic; net complexity increased. |
| 3 | No-input maintenance could consume the occurrence anchor before delivery became terminal. | Accepted. | Added retained-row/no-input machinery; substantial complexity. |
| 4 | Two attempts were invalid; the valid retrospective identified split selection ownership as the repeated mechanism. | No invalid finding accepted; retrospective accepted. | Replaced split consumers with an exclusive coordinator; largest complexity increase. |
| 5 | First attempt invalid; valid review found an abandoned-history collision under the stable delivery key. | Accepted. | Added one narrow active-intent predicate. |
| 6 | No qualifying finding. | PASS. | None. |
| 7 | Deletion redesign lost occurrence ordering and widened dirty-window admission; proposed global delivery blocking would stall live chat. | Accepted timestamp and target-scope gaps; rejected global cross-turn terminal blocking. | Small derived timestamp selection and target-kind guard; no new state. |
| 8 | Target restriction covered only one import, receipt time could misorder input, causal-only passes admitted completions without a cutoff, and foreground lookup performed background maintenance. | Accepted all three grouped findings. | Shared context, narrower allowlists, and read-only lookup; no new owner or state. |
| 9 | Active-turn system imports dropped the target restriction; a terminal stale index head could displace the fresh pass cutoff. | Accepted both. | Propagated existing context and added a bounded read-only batch lookup; no new owner or state. |
| 10 | The pre-auto-reply system catch-up still dropped the pass-wide target restriction. | Accepted. | Moved only the target-kind fallback to the shared import boundary and deleted duplicated active-loop propagation; net source deletion. |
| 11 | The shared target fallback made a checkpoint-gated consented-member Ask become a current-reply delivery barrier, delaying an unrelated private reply until idle. | Accepted after a requirement-level retrospective. | Deleted the remote catch-up/barrier and retained existing local authority; 74 production-source lines deleted net. |
| 12 | No qualifying finding. | PASS. | Confirmed the round-11 correction as a net deletion with coherent local channel authority. |
| 13 | No qualifying finding in the current-main merge or manual documentation-index resolution. | PASS. | Confirmed both contracts were preserved and executable behavior was unchanged. |

The first round-10 attempt was `INVALID`, not a substantive round: the PR body
still described the prior head and did not explain unrelated base-history paths
in the cumulative Git delta. The corrected retry kept round number 10, found the
pre-auto-reply scope gap, and that valid finding was reproduced and fixed.

## Verification

- ReviewGPT round 11: one valid `REVIEW_INDUCED` finding on
  `7c3c53426211aaf0e0e08d852c214a3e967eeabd`; the requirement decision and
  deletion correction landed in `97d8aa58309dd95ebcd737858f8fcdb391c7e176`.
- The overlap regression failed before the fix with a deferred consented-member
  Ask causing `auto-reply.blocked`; after the fix the current reply reaches the
  ordinary delivery path.
- Focused proof: 4 files / 616 passed; Assistant Runtime typecheck passed.
- Pristine canonical `pnpm test:diff`: passed, including Assistant Runtime 76
  files / 1,797 passed and 2 skipped, plus Cloudflare 1,856 passed.
- Full pristine `pnpm verify:acceptance`: passed, including Web 6,126 passed and
  150 skipped, Assistant Engine 2,600 passed and 5 skipped, Cloudflare 1,856
  passed, worker verification, package checks, and production builds.
- `git diff --check`, scoped parent review, and privacy inspection passed.
- ReviewGPT round 12 returned `ROUND_OUTCOME: PASS` with zero findings on
  `97d8aa58309dd95ebcd737858f8fcdb391c7e176`; requested-model evidence was
  valid.
- The current-main merge required one manual conflict resolution in
  `agent-docs/index.md`. Canonical docs-scoped `pnpm test:diff` passed, the
  resulting merge tree is conflict-free, and runtime source merged without
  manual edits.
- ReviewGPT round 13 returned `ROUND_OUTCOME: PASS` with zero findings on
  `8e2d2cadaed345ecb786df93082352bb20469e80`; requested-model evidence was
  valid.
- Exact-head GitHub CI on `8e2d2cadaed345ecb786df93082352bb20469e80`
  is fully green, including release coverage/build/typecheck, hosted runner
  bundle, every hosted E2E shard and aggregate gate, viewport, hygiene, and
  frontend-design proof.

## Change breakdown

| --- | ---: | ---: |
| Source | 281 | 157 |
| Tests | 846 | 182 |
| Docs/plans | 439 | 2 |
| Config/tooling | 0 | 0 |
| Generated/other | 0 | 0 |

## Deployment

Runtime-only behavior change. No schema migration, Web producer change,
Temporal command-order change, Cloudflare API change, or tandem Vercel
deployment is required. Deploy the updated hosted runner normally and verify
one live joined-group Ask wakes private Murph without an early idle snapshot.

ReviewGPT first-reviewed head: 07ba1255d27e494b01c744d69af30899009c7dd1


rocketman-21:

Original requirement
- A completed Assistant Ask older than accepted personal input must be resolved first.
- Retries must keep the existing stable delivery key, ambiguous non-idempotent provider work must never be resent automatically, and accepted personal input must not be stranded.
- The existing mailbox and outbox remain the only durable owners. The costly idle workspace snapshot stays on its existing idle or shutdown schedule.

Shape comparison
- The first-reviewed head admitted Ask completion through a generic foreground-causal mailbox allowlist.
- The current head correctly narrowed the dirty-window reader to a leading joined-group Ask prefix, but the ordering remediation added a feature-specific lifecycle coordinator: retained recording rows, an outbox-only fallback scan, private terminality and wake rules, and one-item cleanup.
- That remediation added 528 lines and deleted 46 lines of authored runtime source. Round 2 proved that the duplicated owner logic can skip a second older completion and can block forever behind an outbox state that deliberately has no automatic retry.

Decision: complexity collapse and continue this PR
- Keep the mailbox row only as the original occurrence and ordering anchor.
- Keep the outbox as the sole owner of delivery state, retry timing, replay, and terminality.
- Replace one-item release with a bounded discharge loop: remove invalid, already-terminal, or safely parked older obligations and immediately check the next older completion before admitting personal input.
- Export and reuse the existing exact-intent outbox wake fact. Delete the private Ask wake calculation and delete the outbox-only compatibility scan. A non-idempotent confirmation-pending intent remains parked and unchanged in the outbox, but its mailbox ordering anchor is discharged so later accepted input can proceed.
- Preserve the existing outbox confirmation grace and stale-reconciliation clock for a real sending intent with no nextAttemptAt.
- Add no queue, scheduler, state machine, durable owner, or snapshot trigger.

Recovery guarantee
- No duplicate irreversible provider send.
- No personal reply overtakes an automatically actionable older Ask completion.
- No indefinite foreground block behind a deliberately parked ambiguous send.
- No synthetic immediate-wake loop.
- Multiple older completions are checked as a complete prerequisite set.

Required focused proof
- Two older completions where the first creates no intent and the second remains active.
- Restart with a terminal first intent and retained first mailbox row, followed by an active second completion.
- A real non-idempotent confirmation-pending Ask: no resend, no synthetic wake, mailbox anchor discharged, later input handled.
- A real non-idempotent sending Ask without nextAttemptAt: owner grace preserved and no tight loop.

This continues the same indivisible correction because the narrower design deletes duplicated coordination and leaves ownership with the existing mailbox and outbox rather than introducing another mechanism.

rocketman-21:
ReviewGPT round 3: 1 accepted High finding.

The ordinary no-personal-input system-mailbox path could materialize a completed Ask, create its exact outbox intent, and remove the mailbox occurrence anchor. A retryable or in-flight intent could then survive restart without the prerequisite that blocks later personal input.

Disposition: accepted and reproduced on the reviewed head. The focused regression fails because the ordinary preparation omits retained-row handling.

Correction in progress: reuse the existing retained exact-intent mailbox lane for due completion work even before personal input exists. Keep the mailbox as occurrence/order owner, the outbox as delivery/wake owner, and do not restore the outbox-only scan or change the idle workspace snapshot schedule.

PR #851, opened 2026-07-22
Archive closed integration ingest shards during idle shutdown

rocketman-21:
## Why this exists

PR #475 taught the integration-ingest layer to read and amend archived monthly shards, but hosted live vaults still kept every hydrated month as raw NDJSON indefinitely. That left closed historical months carrying unnecessary byte and file-processing overhead through every checkpoint and restore.

## User-visible goal and UX

Automatically archive closed UTC-month integration-ingest shards during true idle maintenance. There is no user action or visible format transition: current and future months remain raw and appendable, historical imports continue to work through transparent archive amendment, and a wake or shutdown request aborts compaction promptly.

The maintenance pass is capped at 30 seconds and fails open so checkpointing and Codex compaction still proceed. If a process stops after publishing the gzip but before deleting its identical raw source, startup repairs the exact duplicate pair before the vault is served.

## Invariants

- Only closed UTC months are eligible; current and future shards remain raw.
- Archive bytes are deterministic gzip level 6.
- A raw shard is deleted only after bounded decompression, NDJSON validation, receipt/integrity validation, and source-identity revalidation.
- Raw and gzip representations cannot silently diverge. Exact interrupted pairs are repaired; conflicting pairs block startup with a typed error.
- Symlinks and special files are rejected at the shard boundary.
- Normal gzip reads, append inspection, amendment, rollback, and novelty selection use bounded streaming rather than whole-archive buffering.
- The gzip row parser owns decompression through EOF. Malformed, oversized, and downstream semantic row failures cannot abandon a still-running decompressor or escape as a late unhandled process error.
- Idle work is abortable, time-bounded, and does not prevent the normal checkpoint path on failure.
- Recovery mutations require a checkpoint before runtime service begins.

- Core atomic file publication now supports streamed preparation for replace and exclusive-create cases.
- Integration-ingest receipt, append/resume, amendment, rollback, exact-ID inspection, typed reads, event listing, and novelty-selection paths share the compressed representation safely.
- Hosted runtime startup repairs exact interrupted archive transitions after legacy-format migration.
- Hosted idle maintenance invokes archival only when no pending work remains.
- Vault layout, hosted runtime protocol, reliability, and workspace file-count contracts document the new lifecycle.

## ReviewGPT remediation

- Round 1 found that a valid matching row could return before gzip trailer validation. A child process reproduced a delayed `INTEGRATION_INGEST_ARCHIVE_INVALID` terminating the process; the initial correction removed that early return.
- Round 2 found the same ownership gap on unsafe and parser-error outcomes. The required retrospective chose EOF-before-result as the requirement and prohibited new lifecycle machinery.
- Round 3 proved the caller opt-in correction was too narrow: production device import first traverses exact-ID inspection, and append planning plus ordinary typed reads also share the parser. The child-process test reproduced the real `importDeviceBatch` crash. The retrospective was updated before round 4 to move ownership into the shared parser and delete caller opt-in behavior.
- Round 4 remediation makes malformed and oversized gzip row errors provisional until clean EOF. If a downstream consumer rejects a semantic row, parser finalization drains the same bounded iterator before preserving that typed error. A terminal decompression error remains owned and propagates normally. Raw JSONL and already-buffered ZIP behavior are unchanged.
- Child-process coverage now exercises complete, integrity-invalid, malformed, device-import, and ordinary typed-read outcomes against intact and delayed-trailer-failure archives. It reproduced both the device-import parser crash and a downstream semantic-read crash before the correction; every case exits normally afterward.

## Validation

- `pnpm --filter @murphai/core exec vitest run` — 752/752 passed on the round 4 source.
- `pnpm --filter @murphai/core typecheck` — passed on the round 4 source.
- Corrupted-gzip child-process matrices — reproduced the pre-fix novelty, production device-import, and typed-read process exits, then passed after remediation.
- Assistant-runtime idle/workspace focused suite — 257/257 passed on the first-reviewed patch.
- Full assistant-runtime package suite — 1,791 passed, 2 skipped on the round 3 source before its local diff run was stopped for the accepted round 3 finding.
- `pnpm verify:acceptance` — the round 4 isolated run passed the product/build lanes and stopped on an unrelated setup-wizard selection flake under parallel coverage; the focused wizard file passed 6/6 immediately afterward.
- `git diff --check` and privacy scans — passed.
- `pnpm test:diff packages/core/src/integration-ingests.ts packages/core/test/integration-ingests.test.ts` — the round 4 isolated run passed every affected typecheck, the complete core suite (752/752), and the other reached package suites; the dispatcher stopped on an unrelated vault-usecases test whose ignored generated Health Commons fixture is absent from a clean testbox.
- ReviewGPT correction-verification round 4 — `ROUND_OUTCOME: PASS`, no qualifying findings, with verified `gpt-5-6-pro` model metadata.
- GitHub Actions — all 24 reported checks passed on the rebased head, including runtime checkpoint durability and both hosted aggregate gates.

## Change shape

ReviewGPT first-reviewed head: 8536343c53b47ac91ad27644e7cdbd9217bdbf5e

| Round | Head | Source | Tests | Docs/process | Config/tooling | Generated/other |
| --- | --- | ---: | ---: | ---: | ---: | ---: |
| 1 | `8536343c53b47ac91ad27644e7cdbd9217bdbf5e` | +1151/-168 | +483/-2 | +85/-3 | +0/-0 | +0/-0 |
| 2 | `a6a8505e1e8a83c2200a188fca610d41c0fe3505` | +1151/-171 | +601/-3 | +85/-3 | +0/-0 | +0/-0 |
| 3 | `18e3fcdfe0031d91273fac6623efbbcde24353c6` | +1178/-177 | +642/-4 | +85/-3 | +0/-0 | +0/-0 |
| 4 | `da0452a592dfcc0b26b20995da17f4612e2e364b` | +1224/-198 | +803/-4 | +85/-3 | +0/-0 | +0/-0 |

After the zero-finding round, the four commits were rebased cleanly onto the then-current `main`. `git range-diff` marked every commit exactly equivalent (`=`); the rebased head is `6b17595db7d47d00e95b41014bb11e9e8b95ab72`, and the full GitHub Actions matrix passed on that head. ReviewGPT was not rerun because the reviewed patch did not change.

## Deployment skew

The core and assistant-runtime changes ship together in the Cloudflare runner artifact. There is no Vercel, database-schema, or public API dependency. Because the archived representation relies on the transparent gzip reader/amender introduced by #475, the rollback floor is any runner version that includes #475. Deploy the Cloudflare runner normally; no tandem Vercel deployment is required.


rocketman-21:
ReviewGPT anomaly retrospective (before round 3)

Original requirement: every archived-gzip novelty result is provisional until the bounded reader reaches EOF or reports its terminal decompression error; damaged history must fail open without terminating the hosted process.

Shape: first-reviewed authored source was +1,151/-168 (1,319 churn); the current head is +1,151/-171 (1,322 churn). Round 1 remediation deleted three source lines and added focused regression coverage; it introduced no owner or mechanism. The repeated finding is the same stream-ownership gap: complete now drains, but unsafe and parser-error outcomes can still publish before EOF.

Decision: continue this PR with a deletion/reordering correction at the existing parser/reader boundary. Record unsafe, malformed-row, and oversized-row outcomes while continuing to drain the same bounded iterator; publish either the fail-open or deduplicated result only after normal EOF, or fail open after its caught terminal error. Add no manager, state machine, queue, reconciliation path, or second reader owner. Prove intact and delayed trailer failures after complete and early malformed/integrity-invalid rows in child processes, all exiting normally with the intended selection.

rocketman-21:
ReviewGPT anomaly retrospective update (before round 4)

Round 3 showed that the prior continuation decision chose too narrow a boundary: the novelty caller drained safely, but exact-ID inspection, append planning, and ordinary archived reads still used the same gzip parser in throwing mode. This is the same repeated stream-ownership mechanism, and the next audit would be round 4.

Current shape at the round-3 head is +1,178/-177 authored source (1,355 churn), versus +1,151/-168 (1,319 churn) at the immutable first-reviewed head. Review remediation has added 27 source lines and 9 source deletions; the growth is localized but the repeated finding proves the opt-in callback is the wrong ownership boundary.

Updated decision: redesign by deleting caller opt-in behavior. For gzip only, the existing parser retains the first malformed/oversized-row error, consumes the bounded stream through EOF, then throws that retained typed error after a clean EOF; a terminal gzip error supersedes it and propagates normally. Raw JSONL and already-buffered ZIP behavior stay immediate. Integrity-invalid records continue to accumulate unsafe state in their existing callers. No new owner, state machine, queue, repair pass, compatibility path, or reconciliation mechanism. Production-path child-process proof will cover importDeviceBatch and an ordinary typed archive read in addition to the novelty matrix.

PR #857, opened 2026-07-22
Fix phone-call result conversation continuity

rocketman-21:
## Superseded design note (read first)

This PR originally folded a completed phone-call result into the initiating conversation as internal context, surfaced only on the member's next message. ReviewGPT rounds 1–7 reviewed that context-only design; its retrospective and verification history are retained below as record. At the repo owner's direction the delivery was then **redesigned to be proactive**: on completion Murph composes the result in its own voice and messages the member directly, reusing the existing `assistant.notification.requested` machinery. Because delivery now resolves the member's messaging route fresh at completion time, the context path, the origin-session durability requirement, the pre-provider workspace checkpoint (and its mid-pass dormancy hazard), and the associated causal-eligibility plumbing were all removed. The sections immediately below describe the current, proactive design; that is the intent contract for this head.

## Why

A completed phone call must reach the member. The prior context-only path recorded the result silently and relied on the member happening to message again, and it required a pre-provider durability checkpoint to keep the exact initiating session alive — which introduced a mid-pass `next_wake_at = null` dormancy risk on crash. Proactive delivery reaches the member directly and needs none of that machinery.

## User-visible goal and flow

1. A member asks Murph to place a call from a direct conversation.
2. Murph starts the call; the call row records the initiating session id (used only for request-key idempotency).
3. When the call completes, Web appends an `assistant.notification.requested` mailbox event.
4. Murph composes the result in its own voice and proactively messages the member's resolved messaging route — or skips if there is nothing meaningful to report (allow-send-or-skip). Provider/callee text is treated as untrusted.

## Invariants

- Web remains the durable owner of the encrypted call brief and result.
- Result delivery resolves the member's messaging route at completion time, exactly like every other proactive notification; it does not require the initiating session to be durably present in any snapshot.
- Delivery is idempotent on `phone-call-result:${callId}` (notification `deliveryIdempotencyKey`); a repeated webhook sends no second message and runs no second turn.
- Murph may skip a non-meaningful result; when it sends, provider/callee text stays bounded, quoted, untrusted, and non-authoritative.
- `originSessionId` is retained only for phone-call request-key idempotency and the shared user-action scope; it no longer governs delivery.
- Result encryption-at-rest and Retell authority/idempotency hardening are unchanged.
- Genuine detached assistant notifications keep their existing output-only behavior.

- **Reuse, not new machinery:** completion emits the existing `assistant.notification.requested` wake. No new mailbox kind, runtime consumer, pre-provider checkpoint, or conversation-context module remains. The `phone-call.resulted` wake kind, the `record-phone-call-result-context` pre-planning route and its now-unreachable causal-eligibility plumbing (`maxCausalSeq`, the pending-item `causalSeq` field, `acceptedAssistantInputCausalSeq`), and the conversation-context module were deleted. The live member-preferences `preferenceCausalSeq` ordering is untouched.
- **Dormancy hazard removed:** deleting the pre-provider checkpoint removes the mid-pass `next_wake_at = null` swallowed-turn risk it carried.
- **Phone-call persistence and idempotency:** additive nullable `origin_session_id` remains; a duplicate request key from a different session is still rejected. Legacy rows without an origin session still deliver (delivery does not depend on it).
- **Delivery route:** a member without a resolvable messaging route cannot receive the result — the same requirement as any proactive notification.
- **Deployment boundary:** no new mailbox kind, so there is no result-path consumer-first window; the `create_phone_call` start schema still requires `originSessionId`, so a runner-first window fails those starts closed at the old Web endpoint.

1. Apply the additive nullable `origin_session_id` migration (feeds only request idempotency, so legacy rows without it still deliver).
2. Because completion reuses the existing notification wake, there is no new mailbox kind and no result-path consumer-first rollout.
3. The `create_phone_call` start schema requires `originSessionId`: deploy Web and the runner together, or keep the runner-first window short (a new runner sends the field to an old Web start endpoint that rejects it).

## Verification (current, proactive design)

- Web phone-call result tests pass (notification wake shape, `allow_send_or_skip` policy, `phone-call-result:${callId}` dedupe token, member-route resolution, untrusted-text fence, result summary in instructions); the encryption/authority/replay tests are unchanged and pass.
- Assistant Runtime and Assistant Engine typechecks pass; the affected mailbox-routing, mailbox-state, workspace-runner, workspace-assistant-phase, and workspace-entrypoint suites pass, as do the hosted-execution parser/kind suites.
- The hosted-local Retell call-result roundtrip E2E was rewritten to assert proactive delivery — a single composed message sent to the member's route with the untrusted-text frame and no follow-up user turn, plus idempotent replay — and passes locally.
- The dead causal-ordering plumbing removal typechecks and passes the affected runtime suites.
- Docs updated to the proactive design: `apps/cloudflare/DEPLOY.md`, `apps/web/README.md`, `packages/assistant-runtime/README.md`, and `agent-docs/references/hosted-runtime-protocol.md`.

## Change shape

Classification rule: test paths and `*.test.*` are tests/fixtures; `*.md` is docs; scripts/manifests/workflows are config/tooling; remaining authored code, Prisma schema, and migration SQL are source; binary/generated files are generated/other. No binary files are included.

### First-reviewed baseline (immutable)

| --- | ---: | ---: |
| Source | 476 | 77 |
| Tests/fixtures | 601 | 88 |
| Docs | 142 | 4 |
| Config/tooling | 0 | 0 |
| Generated/other | 0 | 0 |
| **Total** | **1,219** | **169** |

ReviewGPT first-reviewed head: 606ce441664021a6dddcd14c6cba7ac7c5023ce2

### Current head

| Head | Full SHA |
| --- | --- |
| Current | `6c0ce66f2fa24f646b62478708b1e07a418ee17a` |

| --- | ---: | ---: |
| Source | 157 | 88 |
| Tests/fixtures | 588 | 289 |
| Docs | 801 | 6 |
| Config/tooling | 0 | 0 |
| Generated/other | 0 | 0 |
| **Total** | **1,546** | **383** |

The net source contribution shrank versus mid-development because the proactive redesign deleted the context path, the pre-provider checkpoint, and the causal-eligibility plumbing that the earlier context-only design had added.

## Review retrospective (superseded context-only design)

The rounds below reviewed the earlier context-only design and are retained as history. They do not describe the current proactive delivery.

ReviewGPT round 2 found that a generic causal field had coupled phone-result and preference ordering, and that permanent invalid origins retried indefinitely. The correction restored independent causal owners and settled only permanent invalid-origin events.

ReviewGPT round 3 found the newly created direct-session crash window before the ordinary idle snapshot; the first correction put a checkpoint inside `create_phone_call`.

ReviewGPT round 4 proved that mid-tool correction could deadlock and moved durability to the pre-provider quiescent owner.

ReviewGPT round 5 found that deterministic welcome output could create the session before the first user reply; the correction keyed durability on the session's physical presence in the restored snapshot.

ReviewGPT round 6 found the direct-session checkpoint advanced only the outer committed workspace; the correction returned the accepted workspace/version to the runner session.

ReviewGPT round 7 returned `PASS` with no findings on head `806f4bb3411ebecf6059e333261c54d5d6a0278f`.

All of the context-append, pre-provider checkpoint, and causal-frontier machinery those rounds shaped has since been removed by the proactive redesign described above.


rocketman-21:
Round 6 retrospective and continuation decision

ReviewGPT found one accepted high-severity original-PR issue in the round-6 correction scope: the direct-session full checkpoint advanced the outer runtime's committed workspace, but not the still-active checkpoint request session. A canonical metadata write in the same provider pass could therefore submit the pass-start snapshot at the newly accepted version and durably roll the workspace back past the origin session.

The repeated mechanism is divergent ownership of the latest accepted workspace during a mid-pass durability boundary. The previous wake correction was necessary but incomplete: it correctly preserved invocation-local wake/status ownership while leaving the runner-local checkpoint owner stale.

Requirement-level decision: continue with a narrow owner-boundary correction. The accepted direct-session checkpoint will be returned to the existing active checkpoint request session, which will remain the single source for later runner-local checkpoint workspace/version metadata. The outer runtime will continue to own invocation-local wake and status state. This adds no new durable state, queue, state machine, lifecycle, compatibility path, or reconciliation mechanism.

Required proof before the next review round:

- reproduce the stale snapshot-reference rollback through the ambient canonical-write port;
- prove the canonical checkpoint retains the newly published direct-session snapshot;
- crash before the ordinary dirty checkpoint, restore from the accepted workspace, and record the phone result on the exact origin session;
- keep the projected-wake regression and mutation proof passing.

This remains necessary to the indivisible user outcome: the call must not start until its exact origin session is restart-safe, and its result must be available in that same conversation after a process loss without an unsolicited result message.

PR #883, opened 2026-07-23
Add safe OC-to-ENAM R2 migration tooling

rocketman-21:
## Why this PR exists

The hosted `BUNDLES` R2 buckets are fixed in Oceania while the hosted runtime is in Eastern North America, so every cold workspace restore crosses the Pacific. R2 bucket location is immutable; moving the data requires new ENAM buckets and a controlled copy and cutover.

This PR adds temporary operator tooling and a fail-closed runbook for that move. It does **not** create buckets, mutate production data, change GitHub variables, deploy Workers, or read/write production credentials.

Members keep every workspace checkpoint and restore path, and after cutover the runner reads storage in the same region where it runs.

The visible cost is one booked maintenance window. Ordinary message, workspace, automation, and upload writers are fenced. Account deletion is explicitly declined before passkey approval or challenge consumption because, while both buckets can become active, accepting deletion against only one would make rollback republish data. The existing message says maintenance is in progress, nothing was started, and to try again afterwards; it makes no duration promise.

No production layout or component behavior changes in this remediation. The existing synthetic design-catalog study continues to show the maintenance state.

## Architecture

- Keep the application on its existing single `BUNDLES` binding and matching presign bucket. There is no dual read, fallback, data-model migration, new dependency, or persistent migration state.
- Run the whole copy inside one external write fence. There is no live pre-seed.
- Close new account-deletion admission with a 100-percent maintenance deployment and Vercel Skew Protection Threshold set to that exact deployment. Wait the current absolute Node.js Function maximum of 1,800 seconds so every already-running predecessor has retired.
- Treat Vercel runtime logs as diagnostic only. No log query, cursor, request count, status, or flush delay authorizes destination creation.
- Before the destination exists, run a source-only ownership gate. It uses the same automatically paginated AWS CLI inventory primitive as the migration for two stable source reads and obtains each current `hosted_member.id` with its left-joined canonical `hosted_workspace.snapshot_ref` through one Keychain-backed read-only `murph-prod-psql-ro` query. It derives deterministic `hsn_` namespace IDs and exact supported legacy checkpoint keys in memory, then blocks unknown, unowned, unreferenced, malformed, or missing placement.
- Only after the ownership gate passes, create the unbound ENAM destination and use a new pair-scoped key for the existing copy/parity command.
- Do not assume the source is immutable. Scheduled cleanup may delete source objects, so the migration converges using a bounded destination-only `--prune <count>` path and re-proves exact parity.
- Delete the temporary owner gate, migration command, maintenance control, and runbook after OC retirement.

## Invariants the PR preserves

- The tool never deletes from OC and never has a high-level sync path.
- Destination deletion can name only the ENAM destination, requires `--apply`, the exact pair marker, an exact operator count, and a fresh source-absence proof.
- Credentials remain in allowlisted child-process environments; they are never command arguments, logs, generated config, or durable files.
- The source-only owner gate fails closed on an empty source, unstable inventories, malformed or incomplete CLI output, query failure, an absent owner, a malformed canonical snapshot reference, a missing canonical checkpoint object, or ambiguous/unreferenced placement. It prints counts only—never member IDs, namespace IDs, snapshot references, object keys, or the database URL.
- Source and destination differ, destination location is `enam`, lifecycle rules match the canonical file, and lifecycle-managed staging prefixes are empty before copying.
- Every migration proof ends with two stable inventories, the exact pair marker, exact key/size/ETag/storage-class parity, and zero destination-only objects.
- Account deletion is declined only while the maintenance flag is set, before sensitive authorization is spent, and is restored through a fresh deployment on every exit.
- The external fence remains active through copy, variable readback, the immediate Cloudflare deploy, parity proofs, and both controlled cold-restore canaries.

- **Account deletion routes.** The challenge and delete boundaries keep the existing maintenance guard. The delete guard remains before challenge consumption and the route pins `maxDuration = 300`. Round 11 removes the review-induced entry/terminal runtime markers and all log-based migration authority; normal deletion behavior is otherwise unchanged.
- **Canonical ownership.** The temporary operator command reads each current `hosted_member.id` and its left-joined canonical `hosted_workspace.snapshot_ref` through one existing Keychain-backed read-only helper call. The runtime's existing parser extracts exact supported legacy full/base/hot/delta keys; current v2 references must match the member and derived HSN namespace. The database URL and R2 credentials are stripped from the wrong child environment, and query/output failures stop the operation.
- **HostedUserRunner cleanup alarms.** They can still remove source objects during the window. That is why parity is mutation-aware and the only delete is a bounded destination prune.
- **Lifecycle-managed staging prefixes.** `hosted-email/messages/` and `hosted-meal-photos/images/` must be empty before destination creation so the move neither resets retention nor loses a pending import.
- **Vercel skew.** A 100-percent rollout alone is insufficient because skew identifiers can address older deployments. The runbook sets the threshold to the exact maintenance deployment and uses the platform’s absolute predecessor lifetime, not observability inference.
- **Rollback.** Before the first durable ENAM checkpoint, rollback is all three captured OC variables plus one immediate deploy. After the first durable ENAM checkpoint, the operation is forward-only and OC is a bounded retirement copy.
- **Cloudflare rollout.** The three bucket variables switch together in one immediate deployment. Gradual rollout is prohibited because warm containers on different variables would read and write different regions.

- **Prompt:** not applicable; no prompt, instruction-stack, or tool-description change.
- **Frontend:** the current remediation removes server-side log instrumentation only. Existing redacted desktop and mobile design-catalog evidence still proves the maintenance copy and containment with synthetic data.
- **Coverage:** applicable. Canonical diff and acceptance verification both pass on the exact current patch.

## Change shape

Classification: authored source is production/operator TypeScript and TSX; tests are under test paths; docs/process are the runbook and completed plans; config/tooling is the package command. Counts are the three-dot `origin/main...HEAD` diff.

| --- | ---: | ---: |
| Source | 1,612 | 0 |
| Tests/fixtures | 1,401 | 2 |
| Docs/process | 1,063 | 0 |
| Config/tooling | 2 | 0 |
| Generated/other | 0 | 0 |
| **Total** | **4,078** | **2** |

| Review candidate | Head | Files | Source added | Total added | Total deleted |
| --- | --- | ---: | ---: | ---: | ---: |
| Immutable first-reviewed baseline | `100d35947727270a82e0b5504321c5c6c8429af0` | 6 | 1,041 | 2,254 | 0 |
| Current candidate | `e1051675d823a963190d14a99370cef7323d2831` | 18 | 1,612 | 4,078 | 2 |

Current authored-source additions are 571 lines above the first-reviewed baseline. The increase is temporary migration safety tooling, the bounded maintenance guard, and the existing design study—not a new runtime state owner, queue, lease, reconciliation service, compatibility path, or dependency.

## Design proof

- Design page: [account-deletion maintenance study](https://www.withmurph.ai/design?tab=sections#account-deletion-maintenance)
- Desktop evidence: `.artifacts/review-gpt/pr-883-account-deletion-maintenance-desktop.png` at 1,440px.
- Mobile evidence: `.artifacts/review-gpt/pr-883-account-deletion-maintenance-mobile.png` at 390px.

Both captures use synthetic data. They prove the exact copy and containment, not authenticated-flow focus, close-affordance, or constrained-height scroll behavior. Public image hosting remains blocked by the absence of a local least-privilege Images Write credential; no binary evidence is committed.

The Web maintenance deployment must reach 100 percent and become the Skew Protection Threshold before the 1,800-second predecessor drain begins. Clearing the flag also requires a fresh production deployment; an instant rollback behind the threshold is not the release path.

The Cloudflare cutover then switches all three bucket variables in one immediate deployment, followed by exact-version direct-R2 smoke, two parity proofs, and two controlled cold restores. The second restore must use a new Worker version and cold container. Keep OC and the transition credential until the first durable ENAM checkpoint is proven and the one-hour signed-GET window has expired.

## Verification

- Focused Cloudflare migration suite: 54/54 passed, covering current HSN ownership, v2 member-owner mismatch, legacy full/layered/working references, unreferenced and deleted-member legacy residue, missing canonical checkpoint objects, malformed canonical data, unstable inventory, automatic pagination failure, owner-query failure, and identifier-free output.
- Focused Web delete-route suite: 6/6 passed.
- `pnpm --dir apps/cloudflare typecheck` and `pnpm --dir apps/web typecheck:prepared` passed.
- Owner-gate CLI help rendered; every runbook Bash block passed `bash -n`; stale marker/log-query scans were clean.
- Canonical `pnpm test:diff` passed the affected Cloudflare lane: typecheck, 107 Node test files with 1,982 passing tests, and 2 Worker tests.
- Canonical `pnpm verify:acceptance` passed all workspace typechecks, coverage, app verification/builds, package boundaries, documentation gardening, artifact guards, 519 Web test files with 6,643 passing tests, Cloudflare's 1,982 Node tests, and 2 Worker tests.
- `git diff --check` and the direct-identifier privacy scan passed.
- Current read-only Vercel metadata confirms Skew Protection is enabled, Fluid Compute is enabled, and the ordinary Function default is 300 seconds.

## Review findings resolved

ReviewGPT first-reviewed head: 100d35947727270a82e0b5504321c5c6c8429af0

Rounds 1–7 progressively removed the live pre-seed, immutable-source assumption, elapsed-age drains, failure-shape rollback inference, expiring member timing promises, and unsafe abandonment authority. They established the single fenced copy, bounded destination prune, explicit account-deletion deferral, destination-only abandonment credential, and honest design-evidence scope. Round 7 passed the then-current head.

Rounds 8–10 found and fixed review-induced flaws in the account-deletion drain: ambient destructive authority, non-terminal cleanup markers, split-scan races, missing admission/terminal joins, and a pause-before-marker race. The final Round 10 mechanism used paired Vercel runtime markers and Skew Protection.

Round 11 accepted one review-induced High and required a retrospective: Vercel logs document filtering, limits, and retention, but not complete enumeration or bounded indexing, so an empty or partial result could incorrectly authorize copying. The retrospective removes runtime logs from correctness instead of adding another cursor, delay, or poll. The replacement uses the platform-owned admission close plus absolute predecessor retirement, then joins the stable frozen R2 source to canonical Postgres ownership before the destination exists. A prior partial deletion with a removed member row and surviving object now blocks mechanically; a completed deletion leaves no object to copy. This adds no persisted state and leaves the production deletion flow unchanged.

Round 12 accepted one review-induced High: namespace-only ownership incorrectly rejected supported current legacy full, layered base-plus-hot, and working base-plus-delta checkpoints. The correction keeps the same two stable source inventories and one source-only query, left-joins the current member's canonical snapshot reference, reuses the established runtime parser to derive exact legacy keys, requires every canonical checkpoint object to exist, and still blocks deleted-member, unreferenced, malformed, or ambiguous legacy placement. Current v2 references must match their member and derived HSN namespace. No state, owner, queue, compatibility path, or production flow was added.

Round 13 passed the exact remediation head with no qualifying findings. It verified the canonical join cardinality, existing parser/key-boundary reuse, exact legacy reference admission, v2 ownership checks, frozen-source checkpoint presence proof, identifier-free failure boundary, and unchanged deletion-only drift handling. The review's sole body discrepancy was corrected here: focused coverage includes a v2 member-owner mismatch case, not a separate object-key namespace-mismatch fixture.

Round 11 reviewed head: `7a30bcbb1ec27ddcd027f67cb6ee2a83226a7936`  
Round 12 reviewed head: `50398cc44010572bf95531de732425c1128a9367`  
Current exact remediation head: `e1051675d823a963190d14a99370cef7323d2831`  
Round 13 exact-head verification passed with no qualifying findings; all 28 CI checks are green on the same exact head.


rocketman-21:
## Requirement-level retrospective — round 1 `RETROSPECTIVE_REQUIRED`

**Original requirement.** Move hosted `BUNDLES` storage from OC to ENAM so the runtime and its storage are co-located, without losing a checkpoint or breaking restore, and without adding permanent runtime complexity.

**Finding, accepted unchanged.** A copy that runs while production is live cannot converge. Ordinary snapshot and bundle cleanup deletes a source object after that object has already been copied; the delete-free destination rule then rejects the entire destination, so one routine cleanup event forced another whole-bucket copy. Three requirements collided: a live pre-seed to keep the pause short, no destination deletion, and zero destination-only objects.

**Shapes.** First-reviewed head `100d359477`: 6 files, 2,254 additions, 0 deletions. Current head: the same 6 files, net −69 lines against that baseline (313 added, 382 deleted). There is no review-driven growth to attribute and no repeated mechanism; this is round 1 and the correction is a net deletion.

**Decision: redesign by removing the live window, not by reconciling the destination.**

The bounded-pause requirement is the one that yields. The whole copy now runs inside the cutover write fence, so the source cannot move under it and a destination-only object means the fence failed rather than that the algorithm needs a repair path. Aborting is free at any point before the first ENAM checkpoint: nothing at the source changed and no configuration moved, so recovery is to delete the destination — which only ever held copies — and run again.

What that deleted rather than added:

- the `seed`/`final` two-phase command surface, replaced by one command that always requires `--source-frozen` and is a read-only mirror proof without `--apply`;
- the temporary Vercel WAF account-deletion block and its raw `403` product contract;
- the 24-hour ownership lease and the abandonment escalation built on it;
- the live incremental-seed retry path and its separate convergence checks;
- the lifecycle-prefix exclusion and the "must be naturally empty before cutover" booking blocker; staged objects are now carried forward instead of dropped.

**Rejected alternatives.** Destination reconciliation with provenance re-adds the delete authority the preliminary pass already removed, and still needs the account-deletion block plus an application-owned maintenance state. Fencing deletion inside the runtime puts one-time migration machinery in production source. An R2 Bucket Lock on the source is a retention lock on a live production bucket that cannot be shortened, so it would block member deletion requests for its duration. Cloudflare exposes no native R2 replication, snapshot, or bucket-clone primitive.

**Cost accepted.** Member-visible downtime grows from the ten-minute drain plus deploy to that plus the whole-bucket copy at roughly 77 GiB across roughly 61,000 objects. Window sizing is now an explicit runbook step driven by `wrangler r2 bucket info` volume and the timed preview-rehearsal rate, and an overrun is a rebooking rather than a data risk.

**Second round-1 finding, account deletion.** Also resolved by deletion. The block is gone, the authorized success path stays enabled throughout, and the runbook states why each phase is correct: before the fence no ENAM copy exists; inside the fence a deletion mutates the frozen source and fails the proof closed; between the proof and the switch the post-deploy proof catches it while rollback is still free; after cutover deletion targets ENAM and OC is destroyed whole at retirement.

**Proof of the chosen direction.** The mid-fence deletion case is committed regression coverage at the tool boundary (`fails closed when ordinary cleanup deletes an already-copied source object`), and the mandatory real-R2 preview rehearsal in runbook section 2 remains the production-faithful proof of the copy path, metadata equality, and both cold restores.

rocketman-21:
## Round 2 dispositions — `ROUND_OUTCOME: FINDINGS`

Both findings accepted; both were review-induced by the round-1 correction. Fixed at `547ef68c5e`.

**High — the correction does not actually freeze the source. Accepted.**

Correct, and the mechanism was narrower than it looked. Every other deleter — bundle-transition GC, replaced legacy snapshot cleanup — runs only on a write path and was already fenced. The one unfenced owner is the `HostedUserRunner` orphan-cleanup alarm, which fires on schedule with no invocation.

The fix uses that owner's own existing constant rather than a new control. `WORKSPACE_SNAPSHOT_ORPHAN_CLEANUP_MIN_AGE_MS` is 65 minutes, and a candidate is only recorded when a checkpoint replaces a previous snapshot. Once writers are fenced no candidate can be created, so the newest one that can exist becomes eligible 65 minutes after the last possible write. The fence now holds for that full period plus margin before the copy starts, and the runbook requires two equal `wrangler r2 bucket info` object counts across the drain as quiescence evidence. Nothing can become newly eligible during the copy, so this is a freeze rather than an acknowledgement. The fixed pre-copy cost is now over 75 minutes and section 1 sizes the window accordingly.

**High — the rollback branch could republish an accepted account deletion. Accepted, and it was the more serious half.**

The defect was the runbook's reflexive "if either proof fails, restore the OC variables", not the availability of deletion. The two failure shapes need opposite responses and the tool already reports them distinguishably:

- `Destination has N unexpected object(s)` — ENAM holds objects the source no longer has. Roll back: OC has already lost them and abandoning ENAM destroys the extra copies.
- `Destination is not a complete copy of the frozen source` — ENAM lacks objects OC still has. Do not roll back; restoring OC is what would republish the deletion. Establish the cause from the production logs first: an accepted `POST /api/settings/privacy/delete` in that interval means ENAM is correct and the operation continues forward with prompt OC retirement, and explicitly must not re-run `--apply`, which would copy the deleted member's objects back into the active bucket. No deletion means the copy is genuinely incomplete and `--apply` is re-run under the still-intact fence.

This keeps the authorized deletion path enabled, which `docs/contracts/00-invariants.md` § Product-Critical Flow Preservation requires, without adding a dual-delete binding, a queue, or lifecycle state. `reports opposite recovery directions distinguishably` is committed regression coverage, because the recovery contract now depends on those two messages never collapsing into one.

**High — copying lifecycle-managed staging objects resets retention. Accepted in full.**

The prior shape was the smaller and safer one and is restored: the migration now refuses to start while anything remains under `hosted-email/messages/` or `hosted-meal-photos/images/`, in both `--apply` and the read-only gate, before it creates or copies any destination object. The objects are neither copied nor dropped; the existing import, post-checkpoint cleanup, account-deletion, and lifecycle owners empty those prefixes and the window is rebooked if they cannot. `refuses to start while a staged raw email / staged meal photo object remains` and `blocks the read-only gate while a lifecycle-managed source object remains` cover both entry points and assert zero mutations.

**Not adopted: the four end-to-end deletion proofs.** Deletion at each cutover boundary is an operator-run production sequence against real buckets, not something this repo has a lane to execute; the runbook owns it, and the real-R2 preview rehearsal in section 2 remains a mandatory prerequisite. The tool-boundary invariants those proofs would protect — recovery-direction distinguishability, staged-prefix refusal, and destination-only detection — are committed regression coverage instead.

**Not adopted: proving a real scheduled cleanup converges during the copy.** Under the drain that is now unreachable by construction: no candidate can be eligible during the copy, so there is no convergence case left to demonstrate. The drain itself is what the rehearsal proves.

Verification at `547ef68c5e`: `pnpm --dir apps/cloudflare typecheck` passed; `pnpm test:diff apps/cloudflare` passed with 1,892 Node tests; the focused migration suite passes 35/35.

rocketman-21:
## Round 3 retrospective and dispositions — `ROUND_OUTCOME: RETROSPECTIVE_REQUIRED`

Round 3 was right on all three points, and right that another tactical branch could not close them. Both requirement-level decisions it asked for were made and are implemented at `ab35d2eae8`.

**The frozen-source premise was wrong, and verifiably so.** `cleanupOrphanCandidates` retains the candidate and rethrows on a failed Web read or R2 delete, and `apps/cloudflare/test/user-runner-alarm.test.ts` pins that contract as `rethrows alarm cleanup failures so Cloudflare can retry`. The platform then retries on its own schedule, which no elapsed-age drain can bound. The 65-minute drain was a timing heuristic dressed as a proof, and it is deleted.

**Decision 1 — the source does not have to be immutable.** The migration now converges on the source as it actually is. When cleanup removes an object after it was copied, the run removes the stale destination copy rather than abandoning a 77 GiB destination. That delete is the only one the tool can issue and it is bounded four ways: it can name only the destination bucket, it runs only under `--apply`, it requires the exact pair marker, and it requires an operator-supplied `--prune <count>` that must equal what the read-only gate reported, after a second source read confirms every key is still absent. Without `--prune` the run still fails closed. Coverage: `prunes exactly the objects the source no longer has, then converges`, `never issues a delete against the source bucket`, `refuses a prune count that does not match the observed drift`, `refuses to prune a key the second source read still reports`, and `keeps the read-only gate incapable of deleting`.

This is the reviewer's "abandon the frozen-source premise and use a mutation-aware migration design", and it also removes the round-1 non-convergence mechanism at its root rather than by timing.

**Decision 2 — account deletion is deferred for the window.** Taken as the explicit product, privacy, and legal decision the review said it had to be, not as a technical fallback. While two buckets can each still become active, an accepted deletion cannot be honoured in both, and there is no durable record of a deleted member to repair from afterwards; deferring is the only option that never leaves a member's data recoverable in a bucket that later goes live. Statutory response windows are measured in days and this deferral is measured in hours.

`POST /api/settings/privacy/delete` returns `503` with `account_deletion_maintenance` while `HOSTED_ACCOUNT_DELETION_MAINTENANCE=1`. The check runs before the sensitive-action challenge is consumed, so the member keeps an unspent authorization and loses nothing, and the message names the exact return time. Coverage: `declines with a truthful message and spends nothing`, `stays available whenever the window flag is not exactly set`, and `shows the member the maintenance-window reason verbatim`, which asserts the rendered sentence and asserts the generic retry line is absent.

Because no deletion can be accepted while both buckets exist, rollback is unconditionally safe again. The failure-shape recovery table and its deployment-log correlation are deleted — the review was correct that error text and manual log correlation could not own that decision.

**The three concrete round-3 defects.**

1. *Rollback "on any failure" after both proofs pass.* Real, and now unreachable: no deletion can occur in that interval, so restoring OC cannot republish anything. Section 7 is a single unconditional rollback again.
2. *Mixed divergence collapsed into one message.* Real, and mine. `verifyFrozenMirror` threw on source-only failures before it looked at destination-only ones. It now collects both directions and reports them together, so a mixed state can never read as a pure one. The pure-shape test was rewritten as `reports both divergence directions instead of collapsing them`.
3. *Point-in-time observation.* Dissolved with the deferral: the state the proof observes can no longer be changed by a deletion between the proof and the deploy or rollback that follows it.

**Not adopted: the four end-to-end deletion proofs.** Three of the four states no longer exist — deletion during the copy, after the final proof, and after ENAM deployment are all declined by the window. The fourth, deletion during OC retirement, is the ordinary post-window path and is proven by clearing the flag in section 8 and requiring a real deletion to reach the application.

**Shape.** Authored source is 1,157 lines against 1,041 at the first-reviewed head. The growth is the 55-line maintenance module, the 4-line route call, and the prune path; the 65-minute drain, the recovery table, and the log-correlation branch were deleted. No new owner, queue, lease, lifecycle, compatibility path, migration, or reconciliation loop was added; the prune is a bounded operation inside the existing one-shot command, and the deferral is one env-gated guard with an exact removal condition in section 8.

Verification at `ab35d2eae8`: both app typechecks pass; `pnpm test:diff` passes with 6,219 web tests and 1,897 Cloudflare tests; the focused migration suite passes 40/40.

rocketman-21:
## Round-5 cap retrospective and requirement-level decisions

Round 5 returned `RETROSPECTIVE_REQUIRED` and this is also the five-round cap, so this record covers both. The repository owner has given an explicit decision to continue past the cap on this PR.

**The round-5 finding is accepted and was mine.** Section 4 set the maintenance flag and deployed it before section 5 created the ENAM destination. If any section-4 prerequisite failed, the abort left the flag set, and the abandonment procedure I added in round 4 was inapplicable because it requires a destination that did not exist yet. That is the round-4 mechanism repeating: a temporary control outliving its operation and indefinitely disabling a privacy-critical flow.

### Decision 1 — an exact return time is not required. Deleted.

The timestamp was the source of most of this control's surface and most of its failure modes: a second env var, a formatter, an expired-time fallback, an extension-deployment rule, and a sizing floor that existed only to keep the promise honest. A named time can expire while the window is genuinely still open, which is a worse lie than naming none.

Deleted: `HOSTED_ACCOUNT_DELETION_MAINTENANCE_UNTIL`, `formatMaintenanceWindow`, the elapsed-time branch, the extension rule, and the advertised-time floor in section 1. The message is now one constant ending "Please try again in a few hours."

### Decision 2 — one rule owns the window, and it is derived from a single condition.

The control is set only while an ENAM bucket holds copies that could still become live, and cleared the moment OC is sole authority again. The cutover owner owns both edges. Every exit follows from that one condition rather than from its own patched branch:

| Point in the operation | Control |
| --- | --- |
| Anything before the copy, including every section 4 abort | Never set |
| Copy, cutover, proofs, canary | Set |
| Any pre-commit abandonment or overrun | Cleared before writers reopen |
| Successful retirement | Cleared in section 8 |

Activation moved from section 4 to the last step of section 5, immediately before the first object is copied. That deletes the reported failure rather than adding a recovery path for it: an abort during the fence now happens while the control is still unset, so there is nothing to unwind.

Both guards are kept and both are necessary. The challenge guard is where the member finds out, before passkey approval and browser-vault teardown; the delete-route guard is the effect boundary, without which a direct request would bypass the window. They are 13 lines together.

### Decision 3 — the cross-app work stays in this PR, and it shrank.

Authored production source is now 1,152 lines against 1,041 at the first-reviewed head: 1,098 Cloudflare and 54 web, down from 1,176 at the previous head. The web side is a 41-line module and 13 lines across two guards.

Splitting it out would produce a PR that disables a privacy-critical flow with no visible reason, and an R2 migration that is unsafe to run without it. The deferral exists only because two buckets can each become active during this specific cutover, and it is deleted with the runbook in section 8. Keeping them together is what makes both reviewable.

### Decision 4 — governing retrospective and removal

The round-1 retrospective chose the frozen whole-bucket copy; round 3 replaced its frozen-source premise with mutation-aware convergence and added the deferral. This delta stays inside round 3's decisions: it adds no owner, state, or mechanism, and only removes surface from the control round 3 introduced. Removal is section 8 step 4 (clear the variable, prove both routes reach the application again) and step 5 (delete the module and both call sites with the runbook).

### Remaining known gap

There is still no browser capture of the rendered maintenance rejection. The proof is the initiation-path component test, which asserts nothing is started, plus the route tests. Capturing it would require running the hosted app in this worktree with isolated ports and database and uploading through Cloudflare Images; that is stated rather than papered over.

Verification at `148842260c`: both typechecks pass; `pnpm test:diff` passes with 6,223 web tests and 1,897 Cloudflare tests.

rocketman-21:
## Round 6 disposition — `ROUND_OUTCOME: FINDINGS`

Round 6 verified the round-5 correction: the stranded-control mechanism is resolved, the fix added no owner, state, or recovery branch, and it deleted 24 production lines. One finding remained, accepted in full and fixed at `460b50139c`.

**Material UX Failure — "in a few hours" is still an expiring promise. Accepted.**

Correct, and it undercut the round-5 claim. Deleting the timestamp removed the calendar time but left a relative one, and the window it describes is not a few hours: it runs from immediately before the copy until section 8 clears the control after OC retirement, which cannot happen sooner than an hour after cutover, is permitted as late as 24 hours, and is explicitly extended when a retirement check fails. A member could follow the instruction, return, and be declined again by the same advice that had already expired — on a privacy-critical action, where that costs trust and offers no truthful next step.

The fix is the reviewer's, and it is another deletion rather than another mechanism. The copy now ends `Please try again after maintenance.` — a condition that stays true for exactly as long as the window lasts, with no timestamp, formatter, duration, deployment rule, or lifecycle branch behind it.

Aligned with it: the runbook text that claimed the message named no return time now says plainly that it makes no timing promise and explains why, given the 24-hour retirement allowance and the extension path; the `Non-obvious affected surfaces` entry says the same.

Coverage follows the requested shape. `makes no timing promise that could expire while the window is open` asserts the exact sentence verbatim rather than the absence of a four-digit year, and asserts a second attempt inside the same window receives the identical guidance with still no challenge created. `declines an account-delete challenge before one is created` and `leaves every other sensitive action untouched` are unchanged, `declines deletion at initiation, before approval or vault teardown` still proves no deletion request, no vault teardown, and no reload, and `keeps the maintenance reason readable in the open dialog` asserts the same sentence renders in the open dialog.

**Rendered evidence gap, unchanged and stated rather than papered over.** There is still no browser capture of the open deletion dialog showing the rendered server message. The proof is the initiation-path component test plus the route tests. Capturing it needs the hosted app running in this worktree with isolated ports, database, and dist directory, plus a Cloudflare Images upload for the PR body.

**Verification at `460b50139c`.** Both typechecks pass. The focused web suites pass 24/24 and the Cloudflare migration suite passes 40/40. In the combined `pnpm test:diff` run, `health-commons-bryan-johnson-protocol.test.ts` failed once with `ENOENT` on a generated bundle artifact while the health-commons package was being rebuilt in the same invocation; the artifact is present and that suite passes 21/21 in isolation. No health-commons file is in this diff, so that is a build-versus-test race in the local lane rather than a change effect. PR CI is the authoritative check on it.

rocketman-21:
## Round 7 — `ROUND_OUTCOME: PASS`

The final gate passed on `2a1e52f373` with **no qualifying findings**. It verified each round-6 correction directly: the production message contains no duration, timestamp, or other claim that can go stale while OC retirement is delayed or extended; the challenge guard still runs before `createSensitiveActionChallenge` and the regression asserts the complete sentence verbatim, repeats the request under the same active window, and proves no challenge was created; the delete route keeps the shared guard before authorization is consumed; and the component path still surfaces the message in the open dialog before any teardown, invalidation, reload, or deletion request. It also confirmed the correction added no owner, lifecycle state, branch, or replacement timing mechanism, and that the base-merge conflict resolution preserves main's `dialogStep`, `exitReason`, and `exitNote` state positions alongside this PR's `dialogError` at positions matching the merged component's `useState` order, without altering production behaviour.

Evidence and scope validation passed in full: all review-context artifacts present and readable, `review-round.json` identifying correction round 7 with both ancestry checks true, and `changed-files.txt`, `pr.diff`, and the packaged source agreeing on the same 12 files.

**Round accounting.** Seven substantive rounds, past the five-round cap under the recorded cap retrospective and the repository owner's explicit continuation decision. Outcomes: round 1 `RETROSPECTIVE_REQUIRED`, round 2 `FINDINGS` (2 accepted), round 3 `RETROSPECTIVE_REQUIRED`, round 4 `FINDINGS` (2 accepted), round 5 `INVALID` then `RETROSPECTIVE_REQUIRED`, round 6 `FINDINGS` (1 accepted), round 7 `PASS`. Every accepted finding was fixed by deleting a mechanism rather than adding one; the two retrospectives replaced the live pre-seed with a fenced copy and then replaced the frozen-source premise with mutation-aware convergence plus a deferral. Tooling retries, which do not advance the counter: two failed draft stagings under lane contention before round 7 landed, and one `INVALID` caused by editing the PR body after the review package was sealed.

**Remaining gap, unchanged and stated.** There is no browser capture of the corrected sentence inside the open delete-account dialog, so this gate does not independently establish visual craft. The available evidence is the component source, the DOM-level component test that exercises the initiation path, and the route tests.

**CI.** 25 checks pass. `Murph Host Support` failed on this head in `assistant-engine/test/assistant-codex-runtime.test.ts`, which no file in this PR touches. The signature is a timeout cascade — three tests hit their 60s ceiling and the remainder fail in 3-6ms behind a poisoned warm process — and the same workflow passes on `main` and passed on every earlier head of this branch. Failed jobs have been re-run to confirm.

rocketman-21:
## Specialist assurance pass — 2 High accepted and fixed, 1 Medium narrowed

An extra preliminary specialist pass was run at the owner's request, after the final gate had already returned `PASS`. It was worth running: it found two high-severity holes that seven final-gate rounds did not, both in operator procedure rather than in the tool. Fixed at `bd8447c612`.

**High — abandonment could recursively delete the authoritative source. Accepted.**

The worst finding in this PR's history, and entirely mine. The abandonment procedure ran `aws s3 rm --recursive` using the pair-scoped migration credential, which is authorized for the source as well as the destination. A stale, unset, or transposed `DESTINATION_BUCKET` would have destroyed the ~77 GiB authoritative dataset, and the only guard I had written was an instruction to check the variable first. Shell-variable discipline is not an adequate guard for an unrecoverable outcome.

The step now mints a temporary destination-only key and proves it cannot even *read* the source before anything is deleted:

```bash
# Must fail with AccessDenied. If it succeeds, the credential is wrong: stop.
aws s3api list-objects-v2 --bucket "$SOURCE_BUCKET" --max-items 1 ...
```

Only after that read is denied does the recursive delete run, followed by a source object-count comparison against the count recorded in section 1. The catastrophic outcome is now outside the credential's authority instead of merely discouraged.

**High — deletions already in flight were never drained. Accepted.**

Also correct. Both guards run once, at request entry. A deletion accepted by the previous deployment continues through vendor revocation, database deletion, and Cloudflare object deletion, and nothing re-checks the flag partway through. If one is still running when the copy starts it can remove a member's objects from one side of the pair only, which is precisely the outcome the window exists to prevent, and it would make the section 7 rollback unsafe too. My tests proved new requests are rejected; they never proved admitted ones had finished.

Activation is now followed by a mandatory pre-copy drain: record when the maintenance deployment reached 100 percent, confirm from production logs that every `POST /api/settings/privacy/delete` started before that time has logged completion, and wait ten minutes past the last completion with no new admitted request. That is the same bound section 4 already uses for issued PUT URLs and is far above a single deletion's observed runtime. If anything has not completed, the copy does not start and it is handled as an incident.

Both fixes are runbook-only. No production code, no new owner, state, or mechanism.

**Medium — the catalog evidence reproduces the dialog rather than mounting it. Accepted as stated, not fixed by extraction.**

The finding is right that importing the message constant prevents copy drift only, and that markup can still diverge. I did not extract a shared dialog component, for a reason worth recording: the behaviours the finding is most concerned with — focus order, close affordance, constrained-height scrolling — come from the `Dialog` primitives wrapping the real component, so a catalog frame cannot establish them however it is built. Extraction would trade a real regression risk in a privacy-critical dialog for evidence that still would not cover those behaviours. The complete answer is capturing from a running `/settings`, which the owner deliberately deferred as disproportionate for a control that section 8 deletes.

Instead the study now states exactly what it proves and what it does not, in both its doc comment and its on-page description: the exact sentence and its containment at 1440px and 390px, not focus, close, or scroll behaviour.

**Verification at `bd8447c612`.** Both typechecks pass. `pnpm test:diff` passes with 6,455 web tests and 1,929 Cloudflare tests.

rocketman-21:
## Round 10 retrospective and disposition

**Finding:** accepted, High, review-induced. The pre-cursor/post-cursor split still allowed a request to pass the unset guard, pause before the late admission marker, miss the second query, and resume destructive deletion after destination creation.

**Why the mechanism repeated:** rounds 9 and 10 both tried to infer a continuous request lifecycle by dividing logs at an operator timestamp. Moving the cursor could not make an invocation that had not emitted its admission marker observable. Separately, a 100-percent Vercel rollout was not an authoritative close while Skew Protection could still address older deployments.

**Requirement-level decision:** delete the split scans. The existing Web request remains the sole deletion owner and now emits an identifier-free entry marker before every guard or await. It emits a paired safe-terminal marker on pre-deletion exits or only after the existing aggregate `cloudflare.deleted=true` result. False, throw, timeout, process exit, duplicate, or missing-log outcomes remain unmatched and block destination creation. The platform-owned admission close is the instant Vercel confirms the maintenance deployment at 100 percent **and** advances the Skew Protection Threshold to that exact deployment. A marker-bearing preparatory deployment is qualified with the same threshold primitive after one full route lifetime of observable quiet.

**Correlation decision:** current Vercel CLI JSON was inspected read-only. One request event exposes `id`, `deploymentId`, `requestPath`, `requestMethod`, and nested `logs`; the runbook validates the two messages directly inside that event. It no longer assumes a `requestId` field or performs separate marker queries.

**Architecture:** no durable flag, queue, lease, retry owner, reconciliation process, cross-runtime attempt identifier, or new dependency. The existing synchronous aggregate owner remains authoritative.

**Proof on checked head `7a30bcbb1ec27ddcd027f67cb6ee2a83226a7936`:** focused route suite 9/9; held-before-authentication requests released after activation for aggregate true and false; thrown deletion remains unmatched; `maxDuration=300` asserted; synthetic request-event gate accepts one exact pair and rejects an unmatched entry; 20 executable runbook Bash blocks pass `bash -n`; canonical `pnpm test:diff ...` and `pnpm verify:acceptance` pass.

rocketman-21:
Round 11 retrospective and disposition

Accepted the review-induced High at `7a30bcbb1ec27ddcd027f67cb6ee2a83226a7936`: Vercel runtime logs cannot prove complete enumeration, so the previous entry/terminal marker protocol could authorize destination creation from an empty or partial result.

Resolved at `50398cc44010572bf95531de732425c1128a9367` by removing all route markers and all Vercel-log-based correctness. The replacement is fail-closed and destination-preventing:

- deploy account-deletion maintenance to 100 percent and set the Skew Protection Threshold to that exact deployment;
- wait the current absolute Node.js Function maximum (1,800 seconds) so every predecessor has retired;
- before creating ENAM, run two stable, automatically paginated OC inventories and one complete Keychain-backed read-only `hosted_member.id` query;
- derive `hsn_` namespaces in memory and block every unknown/unowned object, unstable inventory, malformed output, pagination/query failure, or empty source;
- report counts only; never print member IDs, namespaces, object keys, credentials, or the database URL.

This deliberately adds no cursor, delay/poll protocol, queue, lease, durable flag, reconciliation owner, or production-state change. The existing account-deletion behavior is unchanged outside the maintenance guard. Focused tests, canonical `test:diff`, and full `verify:acceptance` pass. Round 12 exact-head ReviewGPT is starting alongside fresh CI.

rocketman-21:
## Round 12 finding and disposition

**Finding:** accepted, High, review-induced. The namespace-only owner gate at `50398cc44010572bf95531de732425c1128a9367` rejected supported current legacy workspace checkpoints even when the member's canonical `hosted_workspace.snapshot_ref` reached those exact objects.

**Resolution at `e1051675d823a963190d14a99370cef7323d2831`:**

- keep the same two stable, automatically paginated source inventories and one Keychain-backed read-only query;
- return each current member with its left-joined canonical snapshot reference as one JSON row;
- reuse the established hosted-execution parser to extract exact supported legacy full, base, hot, and delta keys;
- require current v2 references to match both the member and derived HSN namespace;
- require every canonical checkpoint object to exist in the frozen source;
- permit only exact currently referenced legacy keys; deleted-member, unreferenced, malformed, missing, and ambiguous placement still blocks;
- keep all derivation in memory and all output count-only.

This adds no persisted state, new owner, queue, retry protocol, compatibility path, or production-flow change. The focused migration suite passes 54/54, Cloudflare typecheck passes, canonical `pnpm test:diff` passes, and full `pnpm verify:acceptance` passes. Round 13 exact-head ReviewGPT is starting alongside fresh CI.

rocketman-21:
## Round 13 — `ROUND_OUTCOME: PASS`

ReviewGPT completed the exact-head correction-verification gate on `e1051675d823a963190d14a99370cef7323d2831` with **no qualifying findings**.

The gate verified the Round 12 remediation directly: one canonical current-member/snapshot-reference query, existing parser and bundle-key boundary reuse, exact supported legacy checkpoint admission, v2 member/HSN ownership checks, required frozen-source checkpoint presence, identifier-free failure output, and no new state or production flow.

It reported one PR-body discrepancy only: the focused suite has a v2 member-owner mismatch regression, not a separate object-key namespace-mismatch fixture. The body claim has been narrowed accordingly; no code change was required.

The response marker and SHA sidecar validate, the actual reviewer model is recorded as `gpt-5-6-pro`, and all 28 CI checks are green on the same exact head. PR #883 remains open and unmerged.

PR #896, opened 2026-07-23
Remove global Linq phone-line pool lock

rocketman-21:
## Why this PR exists

New Linq home-line assignment used one global transaction advisory lock. An unrelated member holding that lock could make activation fail immediately with a pool-busy error and force the user to retry even though assignable phone-line capacity still existed.

Activation and first-contact routing complete without failing solely because another member is choosing a shared phone line. Same-member decisions still converge on one durable home route. Signup-welcome capacity contention is handled inside the same request; ordinary thrown transient failures retain the existing bounded service retry path.

## User experience

- Entry points: hosted activation, inbound first contact, and an operator-requested home-line rehome.
- Immediate behavior: the server serializes only the affected member, reads current routing once under that ownership boundary, and either reuses the route or selects a healthy line.
- Timing and continuation: there is no new user wait or action. A signup-welcome claim retries once on the same line for a UTC-day rollover race, then tries another eligible line before the existing activation/outbox continuation proceeds.
- Terminal behavior: the member receives the existing welcome when a daily proactive slot is claimed. If every healthy line is at the hard proactive cap, activation still assigns a home line and preserves the member-initiated Text Murph path, but omits the proactive welcome.
- Recovery: genuine database/provider failures still surface through the existing bounded retry machinery. There is no synthetic shared-pool-busy failure and no user-facing “try again” branch.

Independent product-experience review: `NO FINDINGS`.

## Invariants

- One durable member row is the sole per-member owner for Linq routing,
  activation, invite, and mailbox writes; there is no separate per-member
  route advisory lock.
- Linq route ownership uses `FOR NO KEY UPDATE`: it still conflicts with
  activation and another route owner, while remaining compatible with mailbox
  foreign-key `KEY SHARE` locks from Linq, Telegram, and other channels.
- A routing mutation takes the member row before any shared participant-contact
  or chat lock, so every affected path follows one PostgreSQL lock order.
- Existing home and pending route authority is reused before any new selection.
- The per-line daily proactive cap remains a hard atomic row-owned gate and cannot overshoot.
- Active-member limits are balancing targets, not hard admission limits; small concurrent overshoot is accepted.
- Healthy/assignable line filtering, degraded-line fallback, privacy-sensitive chat binding, and ordinary genuine-failure retries remain intact.

- Signup welcome admission: the existing conditional line-row update is now the only atomic shared capacity gate. Focused unit tests cover same-line retry, next-line fallback, and hard-cap welcome suppression; a direct PostgreSQL test proves only one concurrent final-slot claim succeeds.
- Inbound first binding: an existing member initially classified as inactive
  first takes the member's route-owner row lock and rechecks access after any
  activation ahead of it commits. Reclassified and already-active messages
  then use that same owner before mailbox append.
- Cross-channel mailbox producers: Telegram updates its columns on the shared
  routing row before appending its mailbox item, while Linq takes the member
  route owner before updating that row. The route owner therefore uses `FOR NO
  KEY UPDATE`, which preserves same-member serialization without blocking the
  mailbox insert's parent-row `KEY SHARE`. A real PostgreSQL regression runs
  the production Telegram and Linq planners in both start orders for an
  already-active member and for a Linq inbound that first observes an
  uncommitted activation. It observes both the activation and routing-row
  waits, then proves activation and both planners commit with both mailbox
  items and both bindings intact.
- Participant-contact routing: pending-contact claims and route bindings take
  the member row before the shared participant-contact and chat locks. This
  prevents the same row/advisory inversion on identity-resolution paths.
- Operator rehome: this route-only operation does not send or open a proactive conversation, so it verifies target assignability under the member lock without claiming or checking proactive-send capacity. Focused ops tests cover active and proactive limit fields remaining non-blocking for manual rehome.
- Configuration and deployment: the existing active-member setting remains deploy-compatible as an advisory selection target. No schema, environment shape, dependency, or cross-runtime protocol changed.

- Prompt: not applicable; no prompt, tool description, or instruction assembly changed.
- Frontend: not applicable; no rendered UI or user-facing copy surface changed.
- Coverage: applicable. Exact-head `pnpm test:diff <touched paths>` passed,
  including 6,276 hosted-web tests, lint, development smoke, and the production
  build. Exact-head `pnpm verify:acceptance` also passed.
- Rendered evidence: not applicable.

Preliminary `completion-specialists` outcome: findings. Two proof gaps were accepted and fixed: the PostgreSQL test now observes the contender blocked on the owner before release, and a focused test covers every optimistic capacity claim losing. The rehome concern was resolved by deleting its obsolete assignment-count gate: manual rehome sends nothing, while the actual proactive-send counter remains the hard atomic owner. The returned test-only patch was downloaded from the exact review thread, inspected in full, passed `git apply --check`, and its two test changes were accepted and implemented.

## Direct proof and verification

- Focused post-review tests: 93 passed.
- Latest focused remediation tests: 263 passed.
- Hosted-web TypeScript check: passed.
- Local PostgreSQL final-slot concurrency proof: passed; an observer confirmed the contender was blocked on the owner, concurrent claims returned one success and one failure, and the persisted count stopped exactly at the limit.
- Local PostgreSQL deadlock proof: the production Telegram and Linq planners
  reproduced PostgreSQL's `deadlock detected` on the prior `FOR UPDATE` route
  owner, including the path where Linq first observes an inactive member while
  activation is uncommitted. With `FOR NO KEY UPDATE`, observer tests cover
  both channel start orders for already-active and reclassified members;
  activation and both planners commit, each event has exactly one mailbox
  item, and the shared route retains both bindings.
- A second real PostgreSQL regression reproduced the same inversion between a
  member-owned route write and a participant-contact claim. The member row now
  precedes that shared contact lock too. All ten direct PostgreSQL concurrency
  proofs pass, including activation versus route ownership in both orders and
  activation/reclassification composed with both cross-channel routing orders.

Final ReviewGPT round 1 found the opposite member-row/route-lock order. Round 2
verified that delta and found the same mechanism could recur through a real
mailbox foreign-key write, so it required a requirement-level retrospective.
The decision was recorded in the
[PR retrospective](https://github.com/cobuildwithus/murph/pull/896#issuecomment-5061099129):
delete the dedicated member-route advisory namespace and use the durable member
row as the only per-member routing owner, without retries, new state, or
lifecycle machinery. Round 3 found that the first row-lock implementation used
`FOR UPDATE`, which created a cross-channel back-edge through mailbox foreign
keys. The correction keeps the same one-row design but uses the minimum `FOR NO
KEY UPDATE` strength, with production-planner PostgreSQL proof. Round 4 reviewed
that remediation and found the inactive-member reclassification caller still
took `FOR UPDATE` before its access refresh. The final correction routes that
caller through the existing `FOR NO KEY UPDATE` owner and expands the real
PostgreSQL planner test to compose activation, reclassification, Telegram, and
Linq in both routing orders. Round 5 reviewed that exact delta and passed with
no findings.

## Change shape

Classification is path-based: runtime TypeScript under `apps/web/src` is source; `apps/web/test` is tests/fixtures; Markdown and the active execution plan/ledger are docs; config/tooling and generated/other are separate. There are no binary files.

ReviewGPT first-reviewed head: 2ba358286bd37e7074f390b15850e6e036ff8fe9

ReviewGPT round-3 head: d64f2a67d18d229081a93a942be82f1aa9deec21

ReviewGPT round-4 head: 20f9734724624e08609928add5fcf824a4416552

ReviewGPT round-5 head: 1b395e4b13e0a0183ed879bf0d219e33e0c1ac8d

| --- | ---: | ---: |
| Source | 132 | 309 |
| Tests/fixtures | 943 | 405 |
| Docs | 139 | 26 |
| Config/tooling | 0 | 0 |
| Generated/other | 0 | 0 |
| **Total** | **1,214** | **740** |

## Design proof

Not applicable; this PR does not change frontend UI.


rocketman-21:

### Original requirement

Remove the global shared phone-line pool lock and its synthetic busy failure so
unrelated members do not block one another or have to retry. Preserve one stable
home route per member, the atomic per-line proactive-send cap, and ordinary
bounded retries for genuine failures.

### First-reviewed versus current shape

- First-reviewed head: 85 source additions and 290 source deletions (375 lines
  of authored-source churn).
- Round-2 head: 99 source additions and 291 source deletions (390 lines of
  authored-source churn).
- Review-driven production growth: 14 additions and 1 deletion. The larger
  review-driven change is regression proof, not runtime machinery.

### Repeated mechanism

Round 1 found opposite ownership order between the member row and the
per-member route advisory lock. The first correction made initially inactive
traffic take the member row first, but an inactive request reclassified as
active could still hold that row while an ordinary active request held the route
lock and reached the mailbox item's immediate foreign key back to the member
row. The correction moved the same two-owner cycle instead of eliminating it.

### Requirement-level decision

Continue this PR with a complexity collapse: the existing `hosted_member` row
will be the single per-member routing owner. The dedicated `home-member`
advisory-lock namespace will be deleted. Route binding, activation, invite
issuance, reclassified first contact, ordinary active mailbox append, rehome,
and runtime route assertions will therefore serialize through the same durable
member row. Chat ownership remains a separate, narrower lock acquired after
the member owner where both are needed.

This adds no retry, queue, state machine, lifecycle state, compatibility path,
or new owner. It removes one lock concept and preserves unrelated-member
parallelism.

### Required proof before round 3

- Reproduce the current row/route/mailbox foreign-key cycle on local PostgreSQL
  using the real mailbox insert.
- Prove both activation-first and active-message-first schedules commit after
  the collapse.
- Keep the final-slot proactive-cap concurrency proof green.
- Run focused tests, typecheck, canonical diff verification, full acceptance,
  and exact-head CI before merge.

Decision: continue the same PR in this deletion-first direction. The immutable
first-reviewed baseline remains unchanged.

PR #917, opened 2026-07-24
fix(web): detect in-app browsers on the contact-card add flow

rocketman-21:
## Problem

A member who opens web onboarding inside a social app's in-app browser (X/Twitter, Instagram, Facebook, TikTok, LinkedIn, Snapchat, …) can tap **Add Murph to Contacts**, but the webview fetches the vCard (HTTP 200 server-side) and never hands it to the iOS contact importer. Nothing visibly happens, the picker has already optimistically advanced, and the download request logged nothing that distinguished the case — so the failure was silent and un-diagnosable.

Confirmed in production: the reporting member was in the X in-app browser. Reproduced end-to-end on the iOS Simulator — a real Safari session opens the native contact preview and imports cleanly; the served vCard bytes parse fine via `CNContactVCardSerialization`. The only failing variable is the webview.

## Change

New shared `apps/web/src/lib/in-app-browser.ts` → `detectInAppBrowser(ua)` returning `{ app, inAppBrowser, isIos }`:
- iOS in-app-browser rule: iOS WebKit UA with **no `Safari/` token** (every real iOS browser — Safari, CriOS, FxiOS — carries it; WKWebViews do not). This catches X, which announces no app token of its own.
- Explicit app tokens: Instagram, Facebook (`FBAN`/`FBAV`/`FB_IAB`), TikTok (`musical_ly`/`Bytedance`), LinkedIn, Snapchat, Line, Twitter, Google app (`GSA`).

Used in two places:
- **Picker** (`murph-contact-card-picker.tsx`): on a detected iOS webview the primary CTA becomes an `x-safari-https://` Safari-escape link with a short plain-language explainer; Android and real browsers are unchanged; the existing `onAddToContacts` callback still fires so onboarding advances identically.
- **Route** (`/api/murph-contact-card`): one structured log line per download carrying the `webview` flag, detected app token, avatar id, and member-id **suffix** only (existing sanitize/suffix logging conventions) — turns a future incident from an hour of forensics into a one-line query.

## Tests

Three new/updated files, 36 assertions: UA fixtures (iOS Safari, CriOS, X webview, Instagram, Facebook, TikTok, LinkedIn, Android Chrome negative), the CTA-swap logic, and the route log assertions.

- `pnpm --filter @murphai/hosted-web exec vitest run --config vitest.workspace.ts --no-coverage test/in-app-browser.test.ts test/murph-contact-card-picker.test.tsx test/murph-contact-card-route.test.ts` → 36 passed
- `pnpm --dir apps/web typecheck` → clean

## Notes

- No behavior change to the download response, and the native Linq share path / group vCard tool are untouched.
- The Safari-escape URL uses `window.location.host`, so it carries a non-default port in local dev; correct in production.

---
<a href="https://app.blacksmith.sh/cobuildwithus/codesmith/murph/pr/917"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787461143&installation_model_id=19880&pr_number=917&repository=cobuildwithus%2Fmurph&return_to=https%3A%2F%2Fgithub.com%2Fcobuildwithus%2Fmurph%2Fpull%2F917&signature=492e4d93daa34231e1cb80b53d1133b07696262e3028efd2f3b9ba4913e65fd1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

A member who opens onboarding inside an iOS in-app browser can add Murph's contact card and have onboarding continue, instead of tapping a button that silently does nothing.

## User experience

Irreducible user purpose: get Murph's number into the member's contacts, once, without the member having to understand why their browser is unusual.

- **Entry point.** The contact-card picker, reached from initial-visit onboarding, Settings contact-card customization, or the standalone add button in the join-invite flow.
- **Fewest necessary steps.** The in-app-browser branch changes one button's label and destination and adds one line of explanation. No extra screen, no extra choice, no confirmation step, and no second gesture.
- **Immediate feedback.** The button becomes "Opening Safari…" and is disabled; the avatar grid is frozen so a late response cannot launch a superseded selection.
- **Timing class.** Sub-second in the normal case. Longest normal wait is the 10s claim-issuance deadline; the launch acknowledgement window is 5s, chosen so a host "Open in Safari?" confirmation can still be tapped.
- **Asynchronous continuation owner.** The picker itself, via one `AbortController` per attempt. No durable store, queue, scheduler, or reconciliation path.
- **Final destination and audience.** Safari opens the member's own member-bound vCard; the audience is that member only.
- **Failure and recovery.** Issuance failure, timeout, dismissal, a synchronous launch throw, or an unacknowledged launch all leave the picker open with a legible retry. Only an acknowledged launch advances or closes the originating surface.
- **What happens next without a new inbound action.** On success the originating surface advances to persona setup or closes, so the member returns to the next step rather than to a finished task.
- **Local rendered evidence.** Desktop 1440x1000 and mobile 390x844 captures of the design-catalog CTA states, packaged for ReviewGPT and linked under Design proof. Gap stated honestly: those renders show the idle CTA states, not the pending/error/disabled-avatar states, and no device-level webview capture exists yet.

- A contact card is only ever served for the member who initiated the request; an unrelated Safari session must not be able to rebind it.
- The handoff claim is short-lived, signed, and fails closed on malformed, tampered, expired, or unknown-key-version input.
- No new secret, dependency, durable token store, or lifecycle owner is introduced; the existing contact-privacy keyring is reused.
- Logs carry a member-id suffix only, never a full member id or the claim itself.
- The non-webview download path is byte-for-byte unchanged.
- Completion is reported only for an acknowledged launch; backgrounding is not completion.

- **Prompt:** not applicable. No assistant prompt, skill, or model-facing text changes.
- **Frontend:** applicable. Rendered evidence packaged for ReviewGPT: `audit-packages/wv-desktop.png` (1440x1000) and `audit-packages/wv-mobile.png` (390x844), proving the default and in-app-browser CTA states from the `/design?tab=components` Contact Card Picker catalog entry.
- **Coverage:** applicable. Canonical command `pnpm --filter @murphai/hosted-web exec vitest run --config vitest.workspace.ts --no-coverage` — current outcome: 509 files / 6,534 tests passing in CI app verification.

## Change-shape breakdown

Classification rule: anything under `apps/web/test/` is tests; `.md` is docs; JSON/YAML/config files are config/tooling; everything else under `apps/web/src/` and `apps/web/app/` is authored source. No binary files and no generated code are in this diff (the design-proof screenshots are hosted externally and deliberately not committed).

| --- | --- | --- |
| Source | 669 | 54 |
| Tests/fixtures | 1498 | 21 |
| Docs | 0 | 0 |
| Config/tooling | 0 | 0 |
| Generated/other | 0 | 0 |
| **Total** | **2167** | **75** |

Authored-source churn is 723 lines; roughly two thirds of the diff is tests. The single largest source file is the 227-line stateless claim module, which review round 2 cleared on forgery, replay, rebinding, endpoint abuse, and key rotation.

Web-only; no Cloudflare Worker, runner bundle, runtime env, or persisted-state change. The claim is stateless and verified by the same deployment that issued it, so there is no cross-version token to strand. During a rolling deploy an old page bundle can still `POST` to the new route (session path unchanged) and a new page bundle can only reach a new route, so both directions are compatible. Per the contact-card ordering, deploy web before Cloudflare. `container_rollout=immediate` is not required.

## Design proof

The Contact Card Picker catalog entry now renders the new Safari-escape CTA state alongside the default.

- Design page: [/design?tab=components](https://www.withmurph.ai/design?tab=components) — "Contact Card Picker" section
- Desktop screenshot: Contact card picker CTA states, desktop
- Mobile screenshot: Contact card picker CTA states, mobile

- **Settings contact-card customization (`CustomizeMurphSettings`)** — This surface uses the shared `MurphContactCardPicker`, so iOS in-app browsers hit the same cookie-jar boundary as onboarding. The shared Safari handoff therefore applies here as well: Settings keeps its own instructions, issues a member- and avatar-bound handoff before leaving the webview, and stays open on issuance or launch failure so the member can retry. Regression proof: `apps/web/test/customize-murph-settings.test.tsx` exercises the real picker (not a mock) for the Settings copy, selected avatar, failure, and retry behavior; `apps/web/test/murph-contact-card-route.test.ts` proves the resulting handoff returns the initiating member's selected vCard with no Safari cookie and cannot be rebound by an unrelated Safari session.

ReviewGPT first-reviewed head: 335f6c3608704ddaf51fd5cff20cce89990386d2

### DIFF
--- a/apps/web/app/api/murph-contact-card/route.ts
+++ b/apps/web/app/api/murph-contact-card/route.ts
@@ -11,6 +11,11 @@ import {
 } from "@/src/lib/hosted-onboarding/linq-contact-card";
 import { readHostedMemberRoutingState } from "@/src/lib/hosted-onboarding/hosted-member-routing-store";
 import {
+  sanitizeHostedOnboardingStructuredLogDetails,
+  toHostedOnboardingLogIdSuffix,
+} from "@/src/lib/hosted-onboarding/logging";
+import { detectInAppBrowser } from "@/src/lib/in-app-browser";
+import {
   DEFAULT_MURPH_CONTACT_AVATAR_ID,
   findMurphContactAvatarOption,
 } from "@/src/lib/murph-contact-avatars";
@@ -28,6 +33,18 @@ export const GET = withJsonError(async (request: Request) => {
   // plus the active-access entitlement check, like the other member-bound
   // settings GET routes.
   const session = await requireActiveHostedAppSessionFromRequest(request);
+  const avatarId = new URL(request.url).searchParams.get("avatar")
+    ?? DEFAULT_MURPH_CONTACT_AVATAR_ID;
+  const avatar = findMurphContactAvatarOption(avatarId);
+  const browser = detectInAppBrowser(request.headers.get("user-agent") ?? "");
+  console.info("Hosted Murph contact-card request.", {
+    app: browser.app,
+    ...sanitizeHostedOnboardingStructuredLogDetails({
+      avatarId: avatar.id,
+      memberIdSuffix: toHostedOnboardingLogIdSuffix(session.member.id),
+      webview: browser.inAppBrowser,
+    }),
+  });
   const prisma = getPrisma();
 
   const routing = await readHostedMemberRoutingState({
@@ -47,10 +64,6 @@ export const GET = withJsonError(async (request: Request) => {
       retryable: true,
     });
   }
-
-  const avatarId = new URL(request.url).searchParams.get("avatar")
-    ?? DEFAULT_MURPH_CONTACT_AVATAR_ID;
-  const avatar = findMurphContactAvatarOption(avatarId);
 
   const [photo, backupPhoneNumber] = await Promise.all([
     avatar.src
--- a/apps/web/app/api/murph-contact-card/route.ts
+++ b/apps/web/app/api/murph-contact-card/route.ts
@@ -1,6 +1,18 @@
 import { requireActiveHostedAppSessionFromRequest } from "@/src/lib/hosted-onboarding/app-session";
-import { hostedOnboardingError } from "@/src/lib/hosted-onboarding/errors";
-import { withJsonError } from "@/src/lib/hosted-onboarding/http";
+import {
+  issueMurphContactCardHandoffClaim,
+  requireMurphContactCardHandoffClaim,
+} from "@/src/lib/hosted-onboarding/contact-card-handoff";
+import { assertHostedOnboardingMutationOrigin } from "@/src/lib/hosted-onboarding/csrf";
+import {
+  hostedOnboardingError,
+  isHostedOnboardingError,
+} from "@/src/lib/hosted-onboarding/errors";
+import {
+  jsonOk,
+  readJsonObject,
+  withJsonError,
+} from "@/src/lib/hosted-onboarding/http";
 import {
   buildMurphHostedLinqContactCardVcf,
   fetchMurphHostedLinqContactCardVcfPhoto,
@@ -14,12 +26,24 @@ import {
   sanitizeHostedOnboardingStructuredLogDetails,
   toHostedOnboardingLogIdSuffix,
 } from "@/src/lib/hosted-onboarding/logging";
+import { assertActiveHostedMemberAccessAllowed } from "@/src/lib/hosted-onboarding/member-access";
 import { detectInAppBrowser } from "@/src/lib/in-app-browser";
 import {
   DEFAULT_MURPH_CONTACT_AVATAR_ID,
   findMurphContactAvatarOption,
+  type MurphContactAvatarOption,
 } from "@/src/lib/murph-contact-avatars";
 import { getPrisma } from "@/src/lib/prisma";
+
+const MURPH_CONTACT_CARD_HANDOFF_BODY_LIMIT_BYTES = 512;
+
+type MurphContactCardAuthoritySource = "handoff" | "session";
+
+interface MurphContactCardAuthority {
+  avatar: MurphContactAvatarOption;
+  memberId: string;
+  source: MurphContactCardAuthoritySource;
+}
 
 /**
  * Downloadable Murph vCard with the member-chosen avatar embedded as the
@@ -29,26 +53,42 @@ import { getPrisma } from "@/src/lib/prisma";
  * headshot; the no-photo option omits `PHOTO` entirely.
  */
 export const GET = withJsonError(async (request: Request) => {
-  // Hosted app session (the same authority that rendered the picker UI),
-  // plus the active-access entitlement check, like the other member-bound
-  // settings GET routes.
-  const session = await requireActiveHostedAppSessionFromRequest(request);
-  const avatarId = new URL(request.url).searchParams.get("avatar")
-    ?? DEFAULT_MURPH_CONTACT_AVATAR_ID;
-  const avatar = findMurphContactAvatarOption(avatarId);
+  const url = new URL(request.url);
   const browser = detectInAppBrowser(request.headers.get("user-agent") ?? "");
-  console.info("Hosted Murph contact-card request.", {
-    app: browser.app,
-    ...sanitizeHostedOnboardingStructuredLogDetails({
-      avatarId: avatar.id,
-      memberIdSuffix: toHostedOnboardingLogIdSuffix(session.member.id),
-      webview: browser.inAppBrowser,
-    }),
-  });
-  const prisma = getPrisma();
+  const source: MurphContactCardAuthoritySource =
+    url.searchParams.has("handoff") ? "handoff" : "session";
 
+  let authority: MurphContactCardAuthority;
+  try {
+    authority = await requireMurphContactCardAuthority({ request, url });
+  } catch (error) {
+    logMurphContactCardRequest({
+      app: browser.app,
+      authOutcome: "rejected",
+      avatarId: source === "session"
+        ? resolveRequestedAvatar(url).id
+        : null,
+      errorCode: classifyMurphContactCardAuthorizationError(error),
+      memberId: null,
+      source,
+      webview: browser.inAppBrowser,
+    });
+    throw error;
+  }
+
+  logMurphContactCardRequest({
+    app: browser.app,
+    authOutcome: "authorized",
+    avatarId: authority.avatar.id,
+    errorCode: null,
+    memberId: authority.memberId,
+    source: authority.source,
+    webview: browser.inAppBrowser,
+  });
+
+  const prisma = getPrisma();
   const routing = await readHostedMemberRoutingState({
-    memberId: session.member.id,
+    memberId: authority.memberId,
     prisma,
   });
   // Same phone authority as the invite/home surfaces that show the CTA: a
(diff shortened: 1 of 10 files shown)


rocketman-21:
## Anomaly retrospective (round 2 `RETROSPECTIVE_REQUIRED`)

Trigger: the same underlying mechanism produced an accepted finding in the previous round — the cross-runtime handoff has no event that truthfully owns completion.

**Original requirement.** A member who opens onboarding inside an iOS in-app browser must be able to add Murph's contact card. The in-app WebKit view fetches the vCard but never hands it to the contact importer, so the normal add is a silent no-op.

**Shape comparison.** First-reviewed head (`335f6c3608`): one direct `x-safari-https://` anchor plus a synchronous parent callback — small, but it crossed the cookie boundary with no authority and claimed completion before anything was served. Current head (`eec3d589f5`): a session- and CSRF-guarded POST, a key-versioned signed-claim module, dual GET authority resolution, and local pending/error state. Authored-source churn: 605 lines for the PR, 515 of it review-driven remediation.

**Repeated mechanism.** Round 1 accepted "advances before a card is served." The remediation removed the premature `onAddToContacts` call and did not replace it, so the successful iOS path now calls neither `onAddToContacts` nor `onOpenChange(false)`. The defect moved from *completes too early* to *never completes*: a member who imports the card in Safari and returns finds the picker still open, and can only proceed by tapping "Skip for now" on a task they finished. `apps/web/test/customize-murph-settings.test.tsx` asserted that open drawer, encoding the broken journey as expected behavior.

**Decision: continue, with the client handoff shrunk — not redesigned into new machinery.**

The stateless authority bridge is retained. It addresses a real WKWebView/Safari cookie separation, and round 2 explicitly found no qualifying forgery, replay, rebinding, endpoint-abuse, or key-rotation issue against it. No durable token store, state machine, queue, scheduler, or reconciliation path will be added.

Requirement-level decisions:

1. **Completion contract.** Successful claim issuance plus Safari launch is completion; the picker then fires `onAddToContacts` and advances/closes exactly as the non-webview path does on click. The source webview cannot observe an OS-level contact import — but neither can today's shipped non-webview path, so this adds no new dishonesty and no extra tap. Members who abandon the import can re-add from Settings.
2. **Single owner.** The picker and its existing parent-callback boundary own initiation, cancellation, and completion for all three callers. The caller inventory is onboarding initial visit, Settings contact-card customization, **and `MurphAddToContactsButton` as used by `join-invite-stage-server.tsx`** — the third surface was missing from the disclosure and is added.
3. **Timing and recovery.** Issuance is bounded by an explicit timeout and aborted on dismissal. The avatar is frozen for the duration of an in-flight issuance, so a late resolution cannot open Safari with a superseded selection or after the member cancelled. Issuance failure, timeout, and abort leave the picker open with a legible retry; only a launched handoff advances.
4. **State deletion.** Any local state that does not correspond to an observable, truthful event is removed rather than kept to make the async code look complete.

Proof obligations: cookie-empty Safari serves the initiating member's selected card; an unrelated ambient Safari session cannot rebind it; each of the three callers reaches its intended next state on success without a mislabeled "Skip"; and pending-cancellation, timeout, selection-change, and issuance-denial paths are covered.

rocketman-21:
## Retrospective addendum (round 3 findings)

Round 3 returned two review-induced findings against `a1d8cdc9c1`. Both accepted; both verified locally before acceptance.

1. **Completion still isn't proven.** `window.location.assign()` returns `void`; returning is not an acknowledgement that a custom-scheme navigation was accepted. A host webview can suppress the scheme, or the member can decline its confirmation prompt, without any synchronous throw — and the picker would still fire the completion callback. The focused tests cannot see this, because the harness replaces `location.assign` with a `vi.fn()`, so "the mock was called" was standing in for "Safari launched."
2. **`AbortSignal.any` excludes older iOS WebKit.** It reached Safari only in 17.4. On an older build the first tap throws before `fetch` starts, the picker shows "Check your connection," and every retry fails identically — a false message, since the network is not the problem. The remediation introduced this dependency; the previously reviewed head did not require it.

**Decision: the existing retrospective direction stands unchanged.** No new implementation direction is being chosen, so this is an addendum rather than a fresh retrospective. The completion contract remains exactly what was decided — successful launch is completion, with no confirmation step and no extra tap. What changes is only the *fidelity of the signal* used to detect that launch: an attempt-scoped, one-shot source-document lifecycle acknowledgement (`pagehide`/`visibilitychange` to hidden) replaces the assumption that `assign()` returning implies success. No acknowledgement inside a short launch window leaves the picker open in its existing retryable error state.

Finding 2's correction is a net deletion: the already-owned `AbortController` absorbs both dismissal and deadline via a `setTimeout`, removing the composite-signal and both static-combinator dependencies. No new state, owner, or abstraction.

One open question is being sent with the fix rather than assumed: whether awaiting the claim POST consumes the tap's transient activation and causes the subsequent scheme navigation to be rejected in the target webviews. If so, claim preparation is reordered so the final tap performs the navigation directly — not solved by adding a confirmation screen or another durable owner.

rocketman-21:
## Round 4 disposition and residual risk

Round 4's finding is accepted and fixed at `3a4dd2784b`: `visibilitychange:hidden` fires for any backgrounding, so an app switch, screen lock, or interruption inside the launch window was being read as a successful Safari launch. It is deleted as a success signal. Only `pagehide` — the document actually unloading — now completes the handoff, which is both the more discriminating signal and one fewer listener and branch.

Two new tests cover the directions the review asked for: dispatching `visibilitychange` and `blur` after a normally-returning `assign` must not invoke any parent callback and must fall through to the real timeout and retry state; and an acknowledged launch invokes the parent callback exactly once even if the acknowledgement is delivered twice.

**Residual risk, stated rather than closed.** Two limits remain and neither is claimed as resolved:

1. A host that leaves its "Open in Safari?" prompt pending past the 5s window will show "Couldn't open Safari" even though a later acceptance still opens the valid handoff URL. The claim remains valid for its 5-minute TTL, so the member's retry succeeds; the cost is a misleading message on return, not a broken card.
2. `pagehide` is chosen on documented WebKit semantics — it tracks unload rather than backgrounding — but this has not been measured on a real social-app webview. That device smoke is owed and listed below, not claimed as done.

**Owed before this can be considered proven on device:** immediate acceptance, deliberate acceptance after the 5s boundary, decline, app switch, screen lock, return to the originating surface, and final vCard delivery — each in a supported in-app browser. If measurement shows `pagehide` does not distinguish acceptance from backgrounding in these hosts, the client-only acknowledgement cannot truthfully implement the settled completion contract and the requirement goes back to a retrospective rather than another heuristic.

PR #923, opened 2026-07-24
Add sleep-stage and every-workout group challenge shares

rocketman-21:
## Why this PR exists

Group challenges could not safely score deep sleep, REM sleep, or time-of-day workout rules from consented shared data. The old workout summary exposed only daily count/minutes, while earlier versions of this PR could reopen a settled date when a member changed time zones, erase valid history after date-line travel, or publish a false settled zero at a UTC source-window boundary.

## User-visible goal and experience

Members can grant separate permissions for deep sleep, REM sleep, every workout by day, and their current time-zone name. Group Murph can then score sleep-stage and rules such as “any workout after 6 PM” without disclosing providers, routes, locations, or absolute workout timestamps.

Workout days carry a producer-owned `calendarClosedThroughDate`. A date is scoreable only after it has ended in UTC-12 and the ordinary projection refresh has published the new watermark. A stale snapshot stays pending; the reader never advances it from its own clock. Time zone remains optional human-readable context and is not settlement authority or location proof.

A model-sized shared read keeps whole member rows only. If the result cannot fit, it returns `status="partial"` plus `omittedParticipantIds`; omitted IDs are still current members and cannot be treated as departed, scored, diagnosed, or permission-missing. Partial results are never complete standings.

## Invariants

- Exact-scope consent: existing sleep/workout grants never widen implicitly.
- `workouts.v0` remains a dense seven-date projection; an empty settled day is an observed zero, while an absent day is unobserved.
- Event time zone, with validated vault fallback, owns workout date/local clock. Mutable profile time zone cannot reopen closure.
- The source read, UTC-12 output window, 500-row cap, and final seven-date filter remain fail-closed and aligned.
- Malformed or unbounded workout data fails the projection closed; a valid event outside the current producer window is deferred without erasing valid dates.
- First published settled challenge results are not silently rewritten by later imports.
- No scheduler, queue, table, compatibility owner, or provider dependency was added.

- The complete Web response keeps its existing projection array and authority fields. At the model boundary, every retained projection is keyed by exact scope and its `grantStatus`/`dataStatus` pair is collapsed to `not_granted`, `missing`, or `available`. Non-workout record arrays remain intact. Only `workouts.v0` additionally compacts repeated day, kind, time-semantics, and completion-watermark fields.
- If that normalized model result exceeds the size limit, only whole member rows are omitted and the result becomes typed `partial`. Encrypted storage and Web-returned truth remain complete.
- Workout delivery records replace per-record `provisional` with one consistent `calendarClosedThroughDate`. Delivery parsing rejects mixed watermarks, and source revisions include the watermark.
- The workout source cutoff is derived from the same fixed UTC-12 producer date as the output window and starts one UTC date before the oldest emitted date. This admits positive-offset boundary workouts before event-zone filtering while preserving the 500-row cap and final seven-date filter.
- The group join disclosure says time zone is optional context, not score-date or location authority. The production join policy and design study render the same shared copy constant.
- Deep/REM retain their existing member-local `provisional` behavior; other challenge scopes retain their established date semantics.

## Review and verification

The one-shot preliminary `completion-specialists` pass applied prompt and coverage lenses. Its UTC prefilter and undefined-cutoff findings were accepted and fixed. The required local product-experience review then found and verified corrections for typed partial results, refresh-timing language, sender attribution, and consent copy; its final re-review returned no findings.

Final ReviewGPT round 11 found and resolved two issues:

1. Between 00:00 and 11:59 UTC, the source cutoff could advance while the UTC-12 output window did not, dropping the oldest positive-offset workout and publishing a false settled zero. The cutoff now uses the same UTC-12 calendar. A full reader regression proves the July 2 Tokyo workout and July 7 watermark remain identical immediately before/after UTC midnight, then proves the window advances at 12:00 UTC without leaking July 2.
2. The PR described model compaction as workouts-only although every scope uses the keyed three-state model envelope. Source comments, durable docs, this disclosure, and test names now state the uniform envelope accurately. A mixed workouts/daily-metric/device-status read proves `available`, `missing`, and `not_granted`, intact non-workout records, workout-only day/kind compaction, and a bounded result.

Canonical remediation proof:

- `pnpm test:diff packages/assistant-runtime packages/assistant-engine` passed all guards and affected typechecks;
- assistant-engine 2,674 tests;
- assistant-runtime 1,884 tests;
- assistant-cli 128, assistantd 40, CLI 1,083, setup CLI 124;
- Cloudflare Node 1,898 and Workers runtime 2;
- focused boundary and mixed-envelope files passed 80 and 61 tests.

The prior pushed head also passed Web’s 6,554 tests, lint with warnings only, cold dev smoke, production build, and all 27 required PR checks. Round-12 CI reruns against the remediation head.

## Change shape

| --- | ---: | ---: |
| Source | 1,079 | 120 |
| Tests/fixtures | 2,927 | 159 |
| Docs | 281 | 2 |
| Config/tooling | 0 | 0 |
| Generated/other | 0 | 0 |
| **Total** | **4,287** | **281** |

ReviewGPT first-reviewed head: a969e95c21df8192ea593d9d45c3d733122c044f

## Design proof

- Study: `/design?tab=sections#group-join`
- Desktop: `audit-packages/pr-923-group-join-desktop.png`
- Mobile: `audit-packages/pr-923-group-join-mobile.png`
- The study renders the real join component with synthetic props, no live requests, and inert controls. It proves the Time zone permission copy in desktop and mobile layouts.

## Deployment concerns

Deploy Web first, then the Cloudflare runner immediately, and keep the interval short. The new workout wire and old workout wire intentionally reject each other: new Web rejects legacy workout `provisional`, while old Web rejects the new watermark. During the bounded interval, workout sharing may be temporarily unavailable but existing conversation, consent, and other projection flows remain intact.

Use immediate runner rollout rather than gradual warm-container skew. After both sides converge, trigger/observe one ordinary projection refresh, verify a workout delivery is accepted, and verify `read_shared` returns the watermark-backed model view. No migration or backfill is required.


rocketman-21:
## Round-2 retrospective (RETROSPECTIVE_REQUIRED)

Recorded before further remediation, per `pr-reviewgpt-loop.md`. Not a merge
verdict; a requirement-level decision on the completed-date mechanism.

### The confirmed defect

`workout-latest-start-days.v0` (and the sleep-stage kinds) date each record in
the member's own event/vault timezone, but the skill decides whether that date
is *complete* by comparing it to "today" in one group-level `scoringTimeZone`.
When the group schedule timezone is ahead of a member's, the group can cross
midnight and finalize the member's still-open local day — announcing a ruling
while the member still has hours to exercise; a later same-day workout then
invalidates a result already shown to the group. Members ahead of the group get
completed days held pending, so the "scored next dispatch" promise is untrue in
ordinary remote/travel groups. Verified in
`packages/assistant-engine/skills/group-challenge/SKILL.md` (completed-date
contract) against `readProjectableWorkoutLocalStart` in
`vault-share-projection.ts`. Classified ORIGINAL_PR: the mechanism existed at
the first-reviewed head; round 2 narrowed its scope but kept the wrong clock.

### Requirement decisions

1. **Completion is decided in the member's own timezone — the same clock that
   dated the record — never the group schedule's.** A date is final only once
   that member's local clock has advanced past it.

2. **Owner: the Vault Share projection producer.** It is the only actor that
   holds the member timezone at production time and can decide completeness
   *without disclosing the timezone* (which the projection deliberately hides).
   No new state owner, scheduler, or reconciliation loop. The daily-metric and
   sleep/workout producers already emit local-dated records over a rolling
   window; completeness is a property of that same production step, not a
   second group-side computation.

3. **Delete the single-group-`scoringTimeZone` completion comparison from the
   skill.** The skill scores the settled records it receives and treats a
   producer-marked provisional (member-current-local-day) record as pending. It
   never re-derives completion from a group clock. This is a net deletion of the
   fragile mechanism and returns to the existing
   `group-challenge-data-diagnostics.md` contract, which already says settled
   standings exclude a current-local-day value and same-day values appear only
   in clearly labeled provisional/live standings.

4. **Truthful pending/delivery timing.** A member's current-local-day value is
   provisional and settles on the first run after that member's own local day
   closes — stated without promising a specific "next dispatch," and correct for
   members both ahead of and behind the group and across DST/travel.

5. **Invariants preserved:** exact grants, no timezone disclosure, no widened
   wire envelope beyond an in-`data` completeness signal if one is needed,
   one-record-per-date, seven-record window, 500-row fail-closed, frozen legacy
   capability list, unchanged behavior for existing scopes.

### Direction chosen

Delete-and-relocate, not another skill patch: remove the group-timezone
completion rule and let the producer own member-local completeness. This is the
minimal-complexity direction (removes a concept and a persisted
`scoringTimeZone` field rather than adding compatibility machinery) and fixes
the bug at its root — the record's own clock decides its completeness.

Change-shape note: this is expected to be net-neutral-to-negative on authored
source (deletes the group-timezone rule and its persisted field, adds a bounded
producer completeness check). Churn stays far below the retrospective size
thresholds; the trigger here was mechanism seriousness, not size.

rocketman-21:
## Round-3 retrospective (RETROSPECTIVE_REQUIRED)

Recorded before further remediation, per `pr-reviewgpt-loop.md`. Round 3 returned
four Material UX Failures, all REVIEW_INDUCED by the `workouts.v0` redesign. I
verified each against the code. This is the second consecutive retrospective, and
findings 1 and 2 are the *same completion mechanism* that produced the round-2
retrospective. That repetition — not size — is the trigger, and it means another
tactical patch on that mechanism is the wrong move.

### Verified evidence

1. **Single-calendar completion.** `selectWorkoutsCalendarTimeZone` picks the
   timestamp-latest candidate's zone as one calendar for the member's whole
   window, then synthesizes empty days and finality against it. A traveling
   member can get a settled "no late workout" that a delayed same-day import
   later reverses, or lose the entire projection when a valid candidate falls
   outside the selected calendar. This is the round-2 mechanism relocated, not
   removed.
2. **No refresh owner for finality.** Snapshots are republished only by idle
   member-runtime work or a one-time grant wake. Nothing advances a provisional
   record when the member's local day closes, so a closed date can stay pending
   indefinitely; and "source complete" means only "the bounded query returned
   ≤500 rows", so late provider delivery can reverse a published settled-empty
   zero. The 7-day window also is not re-trimmed, so an idle grant can expose
   records older than the disclosed window.
3. **Generic kinds wipe the projection.** `resolveAdherenceObservationActivityKind`
   deliberately rejects generic tokens (`workout`, `activity`, `session`,
   `other`, `unknown`) — confirmed in `experiment-adherence.ts:114-132`. A WHOOP
   workout with an unusable sport name canonicalizes to `activityType: "workout"`,
   is rejected, and the selector returns `[]` for the *entire* projection,
   removing unrelated valid workouts.
4. **Aggregate budget.** Confirmed `ASSISTANT_HOSTED_GROUP_SHARED_READ_MAX_RESULT_CODE_UNITS
   = 48_000` and `HOSTED_RUNTIME_GROUP_CHAT_PARTICIPANTS_MAX = 32`. 25 members
   with one workout/day serialize to 49,857 units and the whole read fails
   closed as `group_shared_result_too_large`. The per-member 16 KiB ingress
   test proved a different boundary at a different layer.

### Requirement restated

Unchanged: a group challenge scores every consented member's workouts, including
"did anyone work out after time T", without exposing raw history. The owner
explicitly chose all workouts with start time, duration, and type.

### Measured cause of the size failure

The payload is mostly repetition, not data. Per day-record: **221 units, of which
58 is the actual workout**. `"timeSemantics":"canonical-event-zone-or-vault-zone.v0"`
costs 56 units on *every* record, and the same date is serialized three times
(`data.date`, `recordKey`, `occurredAt`) for another 65.

So the budget failure is self-inflicted format overhead, not the owner's
requirement. No product guarantee needs to yield yet.

### Decisions

1. **Delete empty-day synthesis and producer-side settled claims.** Emit only
   observed workout days. A missing date means unobserved — the existing,
   already-documented semantics. This removes the mechanism behind findings 1
   and 2 rather than patching it, and removes the empty-record cost in finding 4.
2. **Provisional becomes per-record, not per-member.** Each emitted record
   derives its date *and* its provisional flag from that record's own
   event-else-vault timezone. No member-wide calendar is selected, so there is no
   single-calendar reversal or whole-projection wipe. A traveling member's
   records are simply each correct in their own zone.
3. **Remove format redundancy.** `timeSemantics` is a constant of the projection
   kind, not a per-record fact — it must not be repeated per record. The date
   must not be serialized three times. This is deletion, and it is what buys the
   aggregate budget back.
4. **A generic provider workout is a workout.** Adherence-classification
   semantics do not belong at the disclosure boundary. Use a truthful generic
   kind; never let one unclassifiable row invalidate other valid workouts.
5. **Prove the aggregate, not just the member.** Coverage must serialize the real
   model-facing result at the 32-participant cap with the mandatory
   `device-sync-status.v0` scope.

Direction is **net deletion**: remove empty-day synthesis, remove the member-wide
calendar, remove repeated constants. No new scheduler, watermark, paging, or
reconciliation owner. If after this the aggregate still does not fit 32 members,
the product tradeoff (group size, window, or detail) returns to the owner with
real numbers — but it is not justified while two-thirds of the payload is
redundancy.

rocketman-21:
## Cap retrospective (round 5 of 5) + continuation decision

Round 5 returned `RETROSPECTIVE_REQUIRED`. The task owner has explicitly
authorized continuing past the five-round cap, so this records the required
requirement-level reconsideration before round 6.

### What round 5 found, and it is correct

My round-4 remediation was **incomplete, and my test lied about it.**

I narrowed constant-marker hoisting to the `workouts.v0` `timeSemantics`
constant, but left `groupSharedModelRecord` applied to *every* projection, so
`activity-days.v0`, `steps-days.v0`, and `workout-days.v0` still lose
model-facing `recordKey` and `occurredAt`. Their shape is still changed by a PR
that is not about them.

Worse, I named the replacement test "leaves an existing daily scope's model
shape unchanged" while asserting only `metricSemantics` placement via partial
`toMatchObject`. It never asserts `recordKey`/`occurredAt` survive, so it passes
while the property in its own title is false. That is the same defect class I
have been rejecting in returned patches all PR: a green test certifying
something it does not check.

### First-reviewed vs current shape

At `a969e95c21` the adapter passed shared projections through to the model
untouched. The current head adds a model-facing representation owner
(`groupSharedModelProjection` / `groupSharedModelRecord`). That is genuinely
broader than "add three scopes", and the PR body does not disclose it.

### Decisions

1. **Scope the representation change to `workouts.v0` only.** Every other
   projection returns to byte-identical pass-through. This deletes the
   cross-scope behaviour rather than documenting it, and makes the round-4
   Purpose Drift correction actually effective.
2. **Adopt a date-keyed map for `workouts.v0`, with readable field names.**
   Round 5 confirms the map is acceptable and is not objectionable coupling —
   ISO dates are that projection's real day-keyed identity. I am rejecting my
   own single-letter `{s,m,k}` proposal: it buys bytes by moving complexity into
   prompt explanation, and the measurements show it is unnecessary.
3. **Yield the *unconditional* every-workout guarantee; keep every-workout
   semantics inside a declared supported envelope.** No representation can
   reconcile one 48,000-unit read with every permitted 200-member roster; only
   pagination or continuations could, and that is a materially larger
   architecture than this PR should introduce. Measured with readable names and
   a date-keyed map: 32 members at 2 workouts/day = 35,120 units and 40 members
   = 44,288, when device diagnostics accompany only members whose scoring data
   is missing. That last point is spec-aligned — the diagnostics contract
   already states device status cannot override present metric evidence, so
   shipping it for members who have data is pure waste.
4. **Fix the test to assert what its title claims**, including `recordKey` and
   `occurredAt` preservation on untouched scopes.
5. **The 200-member ceiling is out of scope.** `steps-days.v0` already exceeds
   the same budget at 60 members in production today, and at 200 members the
   device-status scope alone exceeds it before any health data. That is a
   pre-existing read-contract limit affecting shipped scopes and belongs in its
   own change, not another compaction hunk here.

### Direction

Net deletion for existing scopes (pass-through restored), one scoped
representation for the new one, an explicit supported envelope instead of an
unprovable universal promise. No pagination, continuation, scheduler, or new
state owner.

Remaining open from round 3 and not yet implemented: single-calendar completion
reversal, snapshot refresh ownership, and generic provider workout kinds wiping
a member's projection. These are tracked and unresolved; this retrospective does
not claim otherwise.

rocketman-21:
## Round-7 retrospective — recommendation: split the PR

Round 7 returned `RETROSPECTIVE_REQUIRED` with three REVIEW_INDUCED findings.
All three are verified against the code. This retrospective recommends
**splitting**, and the evidence for that is the pattern rather than any single
finding.

### The three findings are real

1. **My two-read edit was incomplete.** I changed two locations in
   `group-challenge/SKILL.md` but left the daily-loop owner at line 366 and the
   mandatory companion `group-chat/SKILL.md` at line 364 still instructing the
   combined read. Worse, `assistant-skill-assets.test.ts` now asserts the old
   combined sentence at line 947 *and* my new scoring-only sentence at 951-953:
   the regression contract pins the contradiction in place.
2. **The two-read flow breaks consent recovery.** `observedNotGrantedScopeKeys`
   in `workspace-assistant-phase.ts` is cleared on every read, so a second
   diagnostic read erases the first read's proof that someone needs the scoring
   permission. The one allowed `post_join_offer` is then rejected as
   `scheduled_group_permission_offer_unavailable`, and the affected member never
   receives the consent card. My own correction contract said a failed
   diagnostic read must not discard a consequence of the verified scoring read;
   it does.
3. **The declared envelope is wrong for a third time.** I measured with
   `running` (7 chars). The compact view repeats `kind` per workout and the
   producer accepts up to 80 characters; the repository's own fixtures contain
   `functional-strength-training`. At 32 members and two workouts/day that is
   ~48,620 units — refused. A 32-person strength challenge fails inside the
   envelope I declared one round ago.

### Why this is a design bind, not a queue of bugs

Findings 1 and 2 are in direct tension. One combined read keeps the
permission-offer evidence intact but exceeds the size ceiling. Two reads fit the
ceiling but erase that evidence. Fixing either by itself reintroduces the other,
and fixing both together means taking ownership of read sequencing away from two
mandatory co-owner skills and reworking the runtime's evidence gate — new
ownership, in a PR whose stated purpose was to add three share scopes.

### The pattern

The same result-size mechanism has produced findings in **rounds 3, 4, 5, 6 and
7**. My envelope claim has been wrong three separate times, each for a different
reason: a device-diagnostics conditioning that existed nowhere in the code, then
workout-count-only measurement, now variable-width activity kinds. Every fix
uncovers another payload dimension. Authored-source churn has gone from 448
lines at the first-reviewed head to 835.

Meanwhile `deep-sleep-days.v0` and `rem-sleep-days.v0` have drawn **zero
findings** since round 1. They are small fixed-width daily facts that fit the
existing contract, and they are finished.

### Recommendation

**Split.** Land the two sleep scopes from this PR and move `workouts.v0` to its
own change designed against the constraints we now understand: the 48,000-unit
whole-response ceiling, variable-width activity kinds as a budget dimension,
`group-chat` as a co-owner of read sequencing, and the runtime's per-read
evidence gate.

That is not abandoning the workout feature. It is refusing to keep patching a
representation whose envelope has been disproven three times, inside a PR that
already carries finished, reviewable work. The three round-3 correctness
findings (single-calendar reversal, refresh ownership, generic kinds emptying a
projection) also belong to the workout scope and would travel with it.

Awaiting the owner's decision; not proceeding with another tactical correction
in the meantime.

rocketman-21:
## Adjudicated design for the two remaining blockers

Round 9 returned **zero new findings**; the only things blocking `PASS` are the
two original correctness blockers. I attempted the calendar fix three ways and
each collided with a different requirement already encoded in the suite, so I
put the decision to a design adjudication rather than keep guessing. Recording
the outcome here so it is durable and implementable directly.

### Decision: fixed UTC−12 completion horizon

Completion is decided by one fixed calendar, never by candidate data or mutable
profile metadata:

- Compute the current date in `Etc/GMT+12`, the last zone to leave any date.
- Emit the dense seven-date window ending on that date, keeping empty arrays so
  **provable zeros survive**.
- The fixed calendar's current date is open; every earlier emitted date is
  closed. A date `D` closes at `D + 1 day, 12:00Z`.
- Canonical event timezone (with the existing vault fallback) still decides each
  workout's own date and `startLocalMs`. It no longer decides settlement.
- Defer valid event-local dates later than the window rather than failing the
  projection or moving the window.
- Keep failing the whole projection on genuinely malformed timing/duration/array
  data. A malformed workout must never become an empty-day zero.
- `selectWorkoutsCalendarTimeZone` disappears entirely.

**Why the other two models lose.** Anchoring to the vault timezone can settle a
travelling member's day early, so a later workout still changes a published
result. Dropping empty days removes provable zeros, a deliberate capability
guarded by `scores a settled empty array as observed zero but keeps a missing
date unobserved`. Promptness is the least damaging sacrifice because it is
recoverable and legible, whereas the other two silently produce wrong or
unanswerable standings.

**Accepted cost:** a member in UTC+14 sees results settle up to 26 hours after
their own midnight, and only on the next ordinary runtime wake. The
eastern-member expectation in
`settles a member-ahead date even while a group timezone is still on it` must
change, and that is a correction rather than test-bending: it currently asserts
settlement derived from an inferred current timezone that the system does not
actually know.

### Same mechanism answers the refresh blocker

Replace transient per-day `provisional` flags with one ordered watermark,
`calendarClosedThroughDate`, on the projection. Consumer rule:

```
target > calendarClosedThroughDate            -> pending, snapshot not complete through that date
target <= watermark and days[target] exists   -> score it; [] is an observed zero
target <= watermark and days[target] absent   -> missing / unobserved
```

The reader must never advance the watermark from its own clock; only a newly
produced projection may move it. This is honest about what metadata can do: it
prevents stale data from being *trusted*, but it cannot make stale data fresh.
Guaranteed settlement at the boundary would need a wake owner, so the accepted
cost is safe pending status until the next ordinary wake — which is exactly the
"no scheduler" constraint this PR is under.

One ordered watermark is also simpler than asking the model to interpret a list
of transient flags.

### Implementation state

Producer, parser and watermark were implemented against this design and
typecheck clean, but eight existing tests encode the previous per-candidate
model and the model-view/skill consumer rule still needs migrating. Rather than
leave a half-migrated contract in the branch, that work is reverted and the
branch remains green at `86313e4096`. The design above is settled and ready to
implement in one pass.

Tests requiring migration, per the adjudication: the per-day provisional
assertions become watermark assertions; the diagnostics skill tests must
separate workout completion from event-local representation and use the
watermark; `keeps established Steps challenges on their existing date behavior`
stays unchanged as the guard that this affects only `workouts.v0`; and a new
stale-snapshot regression must prove a prior open date stays pending across a
boundary with no producer run, then becomes a scoreable zero after a refresh.

rocketman-21:
## Adjudicated design for the two remaining blockers

Round 9 returned **zero new findings**; the only things blocking `PASS` are the
two original correctness blockers. I attempted the calendar fix three ways and
each collided with a different requirement already encoded in the suite, so I
put the decision to a design adjudication rather than keep guessing. Recording
the outcome here so it is durable and implementable directly.

### Decision: fixed UTC−12 completion horizon

Completion is decided by one fixed calendar, never by candidate data or mutable
profile metadata:

- Compute the current date in `Etc/GMT+12`, the last zone to leave any date.
- Emit the dense seven-date window ending on that date, keeping empty arrays so
  **provable zeros survive**.
- The fixed calendar's current date is open; every earlier emitted date is
  closed. A date `D` closes at `D + 1 day, 12:00Z`.
- Canonical event timezone (with the existing vault fallback) still decides each
  workout's own date and `startLocalMs`. It no longer decides settlement.
- Defer valid event-local dates later than the window rather than failing the
  projection or moving the window.
- Keep failing the whole projection on genuinely malformed timing/duration/array
  data. A malformed workout must never become an empty-day zero.
- `selectWorkoutsCalendarTimeZone` disappears entirely.

**Why the other two models lose.** Anchoring to the vault timezone can settle a
travelling member's day early, so a later workout still changes a published
result. Dropping empty days removes provable zeros, a deliberate capability
guarded by `scores a settled empty array as observed zero but keeps a missing
date unobserved`. Promptness is the least damaging sacrifice because it is
recoverable and legible, whereas the other two silently produce wrong or
unanswerable standings.

**Accepted cost:** a member in UTC+14 sees results settle up to 26 hours after
their own midnight, and only on the next ordinary runtime wake. The
eastern-member expectation in
`settles a member-ahead date even while a group timezone is still on it` must
change, and that is a correction rather than test-bending: it currently asserts
settlement derived from an inferred current timezone that the system does not
actually know.

### Same mechanism answers the refresh blocker

Replace transient per-day `provisional` flags with one ordered watermark,
`calendarClosedThroughDate`, on the projection. Consumer rule:

```
target > calendarClosedThroughDate            -> pending, snapshot not complete through that date
target <= watermark and days[target] exists   -> score it; [] is an observed zero
target <= watermark and days[target] absent   -> missing / unobserved
```

The reader must never advance the watermark from its own clock; only a newly
produced projection may move it. This is honest about what metadata can do: it
prevents stale data from being *trusted*, but it cannot make stale data fresh.
Guaranteed settlement at the boundary would need a wake owner, so the accepted
cost is safe pending status until the next ordinary wake — which is exactly the
"no scheduler" constraint this PR is under.

One ordered watermark is also simpler than asking the model to interpret a list
of transient flags.

### Implementation state

Producer, parser and watermark were implemented against this design and
typecheck clean, but eight existing tests encode the previous per-candidate
model and the model-view/skill consumer rule still needs migrating. Rather than
leave a half-migrated contract in the branch, that work is reverted and the
branch remains green at `86313e4096`. The design above is settled and ready to
implement in one pass.

Tests requiring migration, per the adjudication: the per-day provisional
assertions become watermark assertions; the diagnostics skill tests must
separate workout completion from event-local representation and use the
watermark; `keeps established Steps challenges on their existing date behavior`
stays unchanged as the guard that this affects only `workouts.v0`; and a new
stale-snapshot regression must prove a prior open date stays pending across a
boundary with no producer run, then becomes a scoreable zero after a refresh.

---

### Implementation addendum (measured)

The producer, parser, model view and skill were implemented against the design
above and all typecheck clean. The remaining work is migrating eight tests in
`packages/assistant-runtime/test/vault-share-projection.test.ts`, and it is a
semantic migration rather than a rename.

Concretely: those tests fix `nowMs` at `2026-07-04T00:00:00.000Z`. Under
`Etc/GMT+12` that is `2026-07-03`, so the watermark is `2026-07-02` and the
suite's primary data day `2026-07-03` moves from settled to pending. Most of
their core assertions invert. That is the intended consequence of choosing the
fixed horizon — an eastern date settles later — but each test needs its
expectation restated deliberately, not bulk-edited, or the suite stops proving
anything.

The sleep scopes keep `data.provisional`; only `workouts.v0` moves to the
watermark, so the sleep assertions at lines ~508-659 stay as they are.

rocketman-21:
### Migration refinement (measured, correcting my earlier estimate)

My earlier note said the eight-test migration was a semantic inversion. Having
run it and inspected the actual diffs, that was **too pessimistic** — for most
of them the change is purely additive.

Representative failure, `scores a settled empty array as observed zero but keeps
a missing date unobserved`:

```
  {
+   "calendarClosedThroughDate": "2026-07-03",
    "date": "2026-07-03",
    "timeSemantics": "canonical-event-zone-or-vault-zone.v0",
    "workouts": [],
  }
```

`date` and `workouts` are unchanged and the test's intent — a settled empty day
is an observed zero — is preserved. Only the new watermark field needs adding.

The remaining non-additive edits are small and localised: three `provisional:
true` expectations (around lines 977, 1124, 1188) and one
`.not.toHaveProperty("provisional")` (around line 1047) become watermark
comparisons, and the shared `scoreSettledWorkoutsDate` helper should encode the
consumer rule directly:

```ts
const watermark = records
  .map((candidate) => candidate.data.calendarClosedThroughDate)
  .find((value): value is string => typeof value === 'string')
if (watermark === undefined || date > watermark) return { status: 'pending' }
```

Branch remains green at `86313e4096`; the source implementation for the
adjudicated design typechecks clean and is described in the comment above.

rocketman-21:
### Correction to my earlier migration estimate

I posted a note saying the eight-test migration was "mostly additive". **That
was wrong** and I am correcting it rather than leaving a misleading estimate.

I generalised from one unrepresentative test. With the source change applied,
six tests fail because they assert **settlement on each member's own timezone**,
which the adjudicated fixed Etc/GMT+12 horizon deliberately replaces. Examples:

- `settles a member-ahead date even while a group timezone is still on it`
- `uses the event timezone for travel finality without disclosing a timezone`
- `keeps a member behind group midnight pending and settles a later threshold flip`
- `keeps both repeated-DST-hour workouts and settles only after the local date closes`

Those encode the *old* completion model. Each expectation has to be restated
deliberately against the new one — it is genuine semantic work, not a field
rename, and my first estimate was the accurate one.

Two things that are established and worth keeping:

1. **Scope the edits to the `selectProjectableWorkoutsDays` describe block only**
   (starts line 895). A broad regex over the file strips `provisional` from the
   sleep tests too; sleep scopes keep `data.provisional` and are unchanged by
   this design. I made exactly that mistake.
2. Shifting only that block's clock from `2026-07-04T00:00:00.000Z` to
   `2026-07-05T00:00:00.000Z` keeps its fixtures on the settled side of the
   watermark (`2026-07-03`), which removes the incidental churn and leaves the
   real semantic edits visible.

Branch remains green at `86313e4096`. The loop stands at round 9,
`ROUND_OUTCOME: FINDINGS`, with these two blockers open.

rocketman-21:
### Round 10: declared-timezone anchor is unsafe as implemented

Round 10 returned `RETROSPECTIVE_REQUIRED`. Three findings, all verified, all
mine. Recording them so the state is not misread as working.

**1. The vault timezone is not the stable fact I claimed.** My comment called it
stable; it is stable against workout ingestion, which was the original bug, but
it is mutable profile metadata. Same clock instant, same workout rows: settle a
day while the zone is Asia/Tokyo, change the zone to America/Los_Angeles, and
that day reopens as `provisional`. A published-final day reverses with no new
data. This is the round-9 mechanism with a different mutable selector.

**2. A date-line-crossing workout wipes the whole snapshot.** Candidates are
admitted using each event zone, but the seven-date window is built from the
declared zone. A genuine Tokyo workout dated July 5 while the declared zone is
still on July 4 falls outside `windowDateSet` and hits `return []`, deleting all
seven days rather than deferring one. The member then reads as
granted-but-missing and can be routed into device diagnostics for data that
exists. The round-10 travel test does not cover it — its Tokyo workout is dated
July 4 — so this is a live bug on the branch.

**3. The disclosure is not wired into the journey.** The skill entry claims the
group can see the clock that settles a day, but the operative instructions still
say to request the scoring scope alone, still say never to disclose the
event/vault timezone, and the always-on group-health prompt does not list
`time-zone.v0`. That was stated intent, not behaviour.

### The structural point

Prompt settlement and irreversible finality cannot both be recomputed from
mutable present state. That needs a monotone closure fact — the
`calendarClosedThroughDate` watermark from the earlier adjudication, where a day
once closed stays closed whatever a later clock says. The watermark makes the
declared-timezone anchor safe rather than replacing it; on its own the anchor
trades one reversal for another.

### State

The owner-directed size work is unaffected and green at `26e3dfb78e`: 64,000
ceiling, explicit truncation naming omitted members, projections keyed by scope
with one `status`, and the earlier trims. Together those took 32 members at two
workouts a day from 48,714 to about 36,057 code units.

The timezone wiring should not ship as-is. Finding 2 is a data-loss bug, and
finding 1 means the anchor does not deliver the irreversibility it was adopted
for.

PR #948, opened 2026-07-25
Give group chats durable outbound addressability and enable group phone calls

rocketman-21:
## Why this PR exists

Murph could not place a phone call from a group chat. The blocker was not phone calls: **group chats had no durable outbound address at all.** `HostedThreadRoute` stored only blinded HMAC lookup keys, so Murph could recognise an inbound group message but had no way to originate anything into that room. Inbound worked only because the provider webhook carried the plaintext thread id in-band, and nothing retained it.

That made every group chat structurally reply-only. Any detached delivery — call results, scheduled automations, long-running jobs, retries — had nowhere to send. This PR fixes the conversation-addressability model and lets group phone calls fall out of it, rather than special-casing phone calls.

In a group chat, a member with their own activated Murph can ask Murph to make a call, and the result comes back **in that group**, not in someone's DM.

## User experience

- **Entry point:** a member asks in the group ("call the clinic and ask if they have anything Friday").
- **Immediate feedback:** the ordinary in-thread turn. No new screen, no new command, no extra confirmation step.
- **Timing class:** minutes. The call runs asynchronously; the longest normal wait is the call itself.
- **Async continuation owner:** the phone-call result notification, delivered through the generic notification-route resolver to the originating group thread.
- **Final destination and audience:** the same group room that asked. Everyone in the room sees the summary, which is the point — the room asked collectively.
- **Failure/recovery:** fail-closed and silent-failure-free. Admission resolves the delivery destination *before* dialing, so Murph cannot place a call whose result would have nowhere to land. A revoked or rebound route fails before provider work rather than delivering to a stale room.
- **What happens next without new inbound action:** the result arrives in the group on its own. No one has to poke Murph.
- **Deliberately unchanged:** 1:1 phone calls behave exactly as before.

Two group-specific restrictions, both server-owned:

- **Transfer is forced off.** A room-owned call must never bridge the live callee to one participant.
- **Only a member with their own activated Murph can trigger a call.** A drive-by participant cannot spend the room's pot or dial on the room's behalf.

- Blinded lookup keys stay one-way HMAC indexes and never become reversible; the new route material is a separate encrypted-at-rest field that is never used as an index.
- Decrypted route material must regenerate the stored blinded keys before use; mismatch, corruption, or absence fails closed.
- A container runtime never falls back to `ownerMemberId` or to `HostedMemberRouting`. A group result never lands in a personal DM.
- Group routes always emit `threadIsDirect: false`; directness is derived from runtime shape, never accepted from a caller. A false-to-true error would bind the event to the wrong audience partition and expose direct-only tools.
- Notification identifiers for a container must exactly match those produced by inbound import for the same container and thread, or detached delivery would open a parallel session instead of joining the room's.
- Exactly one route per container under the current product model; zero or multiple fails closed rather than picking one.
- Requester identity is derived from Web-owned mailbox evidence, never from a runner- or model-supplied value.
- Direct 1:1 phone-call behavior, group billing/usage attribution, dedupe keys, concurrency blockers, and Retell call ownership are unchanged.

- **`murph.create_phone_call` tool description changed** (one added sentence about group transfer). Tool descriptions feed `buildAssistantCodexContractFingerprint`, which hashes the full dynamic-tools array, so this **rotates every exposed session's provider thread on the first post-deploy turn**, with bounded committed-transcript replay. Necessary so the model does not promise a transfer it cannot get. Proof: `packages/assistant-engine/test/assistant-codex-turn-planning.test.ts`.
- **Both webhook providers (Linq and Telegram) now write encrypted route material on authenticated inbound.** Necessary because the route writer is the only place that legitimately holds the plaintext thread identity; this is also what repairs legacy rows without a backfill job. Proof: `apps/web/test/hosted-onboarding-linq-thread-route.test.ts`, `apps/web/test/hosted-onboarding-telegram-dispatch.test.ts`.
- **`apps/web/src/lib/phone-calls/notification-route.ts` deleted.** Its feature-specific derivation is replaced by the generic resolver so route facts stay in one owner. Proof: `apps/web/test/hosted-assistant-notification-destination.test.ts`.
- **`hosted-groups/participant-member.ts` extracted and `group-tool.ts` rewired to it.** The roster's `hasOwnMurph` and the new requester gate now resolve identity through one definition instead of two that could drift. Proof: `apps/web/test/phone-calls-group-requester.test.ts`.
- **`assistant.notification.requested` contract gained an optional external thread route authority**, propagated into `outboxExternalThreadRouteAuthority` so Web can reassert live route ownership immediately before provider egress. Proof: `packages/hosted-execution/test/parsers.test.ts`, `packages/assistant-runtime/test/hosted-runtime-events.test.ts`.
- **Emergency-number dialing — no production owner, guarantee pinned by tests.** An earlier revision of this PR added an emergency short-code catalog plus a Web-egress assertion. Final ReviewGPT rounds 1 and 2 established that machinery was unreachable: every emergency and crisis short code worldwide is two or three digits, the existing `^\+[1-9]\d{7,14}$` constraint requires eight to fifteen, and an invalid brief surfaces to the model only as the generic `invalid phone-call arguments`, so no emergency-specific explanation could ever reach the conversation. It is therefore fully deleted. The requirement — Murph cannot dispatch a call to an emergency number — is satisfied by the E.164 shape making such numbers unrepresentable, and is pinned by explicit regression tests asserting the brief schema cannot represent ~45 emergency/crisis codes or their country-code-prefixed forms while ordinary subscriber numbers still parse. Widening the phone format in future fails those tests. Net effect on this PR's scope: no commingled production behavior remains; only tests asserting pre-existing behavior. Proof: `packages/hosted-execution/test/phone-calls.test.ts`.

- **Prompt — applicable.** One sentence added to the `murph.create_phone_call` description, and it rotates the assistant contract fingerprint (see above).
- **Frontend — not applicable.** No `apps/web` UI surface is touched; there is no rendered state in this diff.
- **Coverage — applicable.** Canonical command `pnpm test:diff <changed source paths>`; current outcome **green, `EXIT=0`**, 513 test files passed / 11 skipped, 6482 tests passed. `pnpm typecheck:packages` green. CI on the reviewed head is green except two jobs that fail identically on `origin/main` HEAD `ae9c99ad49` (run 30176655913): `Release package coverage (cli)` and `Release checks (ubuntu)`. Both fail on `packages/cli/test/release-script-coverage-audit.test.ts > exposes only the package-backed review-gpt runner`, which compares the packaged `@cobuild/review-gpt@0.5.114` preset text against an expected string. This PR's diff contains no CLI release or ReviewGPT packaging path, so the failure is inherited upstream breakage rather than PR-caused. `pnpm typecheck:packages` green.

## Change-shape breakdown

Classification rule: a file counts as **tests** if its path contains a `test/` or `tests/` segment; as **config/tooling** if it is a Prisma migration or `schema.prisma`; otherwise **source**. No binary files. No generated code is committed in this diff.

| --- | --- | --- |
| Source | 1387 | 147 |
| Tests/fixtures | 2155 | 9 |
| Docs | 0 | 0 |
| Config/tooling | 3 | 0 |
| Generated/other | 0 | 0 |
| **Total** | **3545** | **156** |

Counts are from `git diff origin/main...HEAD --numstat` at head `c25d4e973ac1e45adbee46400a96deca9bea11a9`, across 40 files. An earlier revision of this body carried stale counts taken at the first commit; ReviewGPT round 3 flagged the drift and these are the corrected figures.

Test lines exceed source lines because the fail-closed matrix (missing, ambiguous, corrupt, mismatched, wrong-route, direct, self-authored, mixed-requester, unactivated) is enumerated explicitly rather than sampled.

## Design proof

Not applicable. This PR contains no user-facing frontend UI diff; no `/design` catalog surface changes.

This changes both sides of a deploy boundary.

- **Deploy together:** the `hosted-execution` contract, the assistant runtime/engine (runner bundle), and the Web admission change. A new Web service rejects an old group caller that omits the new provenance field, and an old strict Web schema would reject the new field. Direct 1:1 calls remain compatible in either mixed state, so the exposure is limited to group calls, which do not exist today.
- **Migration is safe alone.** `delivery_route_encrypted` is an additive nullable column with no backfill.
- **Warm containers:** a Worker deploy does not instantly replace active runner containers, so warm containers can stay on the old bundle during rollout. Because the group planning gate ships *in* the runner bundle, a warm old container simply does not offer group calls — it cannot emit a group call that new Web would reject. `container_rollout=immediate` is therefore not required, though it shortens the window.
- **Scale and reversibility:** group phone calls are a net-new capability with zero current volume, so the maximum realistically exposed operations during rollout is zero group calls plus ordinary direct calls, which are unaffected. Fully reversible by reverting; the added column can be left in place harmlessly.
- **Legacy rows:** the 20 existing thread containers have no encrypted route material. They stay reply-capable through ordinary inbound and self-repair on the next authenticated **owner** inbound. Repair covers absent material and, since the preliminary-review fix, non-empty ciphertext that is corrupt or whose embedded authority no longer regenerates the stored blinded keys: the owner refresh boundary now opens and validates existing ciphertext instead of trusting its presence. A non-owning delivering line still cannot rewrite the route. Until repaired, a row cannot originate detached delivery, and it never falls back to the owner's DM.
- **Convergence check:** confirm group inbound still routes normally after deploy, then confirm a group-originated call result lands in the group thread rather than a DM.

## Deliberately deferred

Extracting the requester-identity resolver into a generic action-policy layer. The gate is enforced at phone-call admission because that is the only action that needs it today; a shared policy tier would be speculative machinery until a second consumer exists.

---
<a href="https://app.blacksmith.sh/cobuildwithus/codesmith/murph/pr/948"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787598204&installation_model_id=19880&pr_number=948&repository=cobuildwithus%2Fmurph&return_to=https%3A%2F%2Fgithub.com%2Fcobuildwithus%2Fmurph%2Fpull%2F948&signature=fee467fc5cb7525f26375b1f6b0796abc908b1ab019b24373f6ccd0dfff2306f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Preliminary review resolution

One preliminary `completion-specialists` finding (coverage, medium) was accepted and fixed in `0dd4f9a6ae5b`: trusted inbound could not repair non-empty invalid delivery-route ciphertext, because the refresh predicate treated ciphertext *presence* as validity. The owner refresh boundary now decrypts and revalidates existing material, regenerating both blinded lookup keys, and reseals on absence, open failure, or mismatch. Proof: `apps/web/test/hosted-thread-delivery-route-repair.test.ts` plus owner-ingress cases in the Linq and Telegram route tests. Prompt and frontend lenses returned clean.

ReviewGPT first-reviewed head: 0dd4f9a6ae5b7573a4dae976a39ec2de596654bd


rocketman-21:
## ReviewGPT retrospective — repeated emergency-policy Complexity Collapse

Round 2 returned `ROUND_OUTCOME: RETROSPECTIVE_REQUIRED` because the round-1 Complexity Collapse mechanism repeated: emergency-specific machinery whose claimed observable behavior is precluded by the actual production path. This retrospective answers the requirement-level questions before any further review round.

**What went wrong, honestly.** Round 1 asked for the emergency subsystem to be deleted as unreachable. I accepted it only in part: I removed the duplicate Web-egress assertion and the speculative country-prefix stripping, but retained the catalog and refinement on the claim that reordering the check before the E.164 regex made it *reachable* by producing a named refusal reason. That claim was wrong, and round 2 disproved it with the production path: an emergency short code fails brief parsing, `executeMurphDynamicToolRequest` maps that to `invalid-phone-call-arguments`, and the model receives only the generic string `invalid phone-call arguments`. The safe validation digest deliberately omits Zod's custom message and is diagnostic-only. No emergency-specific explanation could reach the conversation. The remediation reduced two dead policy owners to one dead policy owner.

**Requirement decision.** The requirement is: *Murph cannot dispatch a call to an emergency or crisis number.* It is **not** that an emergency-specific explanation reaches the requester or the model. The original instruction was that Murph must never dial one; nothing in the product depends on the refusal being narrated differently from any other unusable number.

**Ownership decision.** The existing E.164 product boundary fully satisfies that requirement. Every emergency and crisis short code worldwide is two or three digits; `^\+[1-9]\d{7,14}$` requires eight to fifteen. Such a number is unrepresentable, so it cannot reach call admission, storage, or the provider. The separate emergency subsystem therefore ceases to be part of this PR: catalog, matcher, message, schema refinement, Web-egress assertion, and dedicated error code are all deleted. There is now **no** emergency policy owner in production.

**Durability, which was my only real objection to deletion.** I had argued that deleting it would leave the invariant as an accidental byproduct of a format regex that could evaporate if the regex were ever loosened. That concern is real but is answered by tests, not by unreachable production code. `packages/hosted-execution/test/phone-calls.test.ts` now asserts the brief schema **cannot represent** ~45 emergency/crisis short codes or their country-code-prefixed forms, and that ordinary subscriber numbers — including `+911234567890`, whose country code begins with an emergency code's digits — still parse. Widening the accepted format fails those tests loudly. Unreachable production code never could.

**Intent-contract alignment.** The PR body's *Non-obvious affected surfaces* entry claimed emergency numbers were blocked "both in the shared schema and unconditionally at the Web egress boundary". That was stale as of the round-1 remediation and is now corrected to describe the deleted owner and the test-pinned guarantee.

**Scope decision.** With the subsystem deleted, no commingled emergency *production* behavior remains in this PR — only tests asserting pre-existing E.164 behavior. The scope question that motivated splitting it out therefore dissolves rather than needing a split.

**Verification on head `3229f6959016`:** `pnpm typecheck:packages` green; full `pnpm test:diff` green, `EXIT=0`, 513 test files / 6482 tests passing.

PR #1012, opened 2026-07-27
Make hosted image generation non-blocking

rocketman-21:
## Why this PR exists

This replaces #968 with the smallest architecture needed for the actual product goal. Hosted image generation should not stall Murph's conversation while the provider and upload finish.

When a member asks Murph to generate an image, Murph can acknowledge that it started, keep handling conversation, and then wake immediately with the uploaded image URL when the work finishes.

## User experience

1. The member asks for an image through an existing conversation.
2. Murph starts provider generation/upload in the current hosted invocation and immediately continues the current turn.
3. Foreground conversation remains higher priority; a new inbound message can complete while the image is still running.
4. When upload finishes, one trusted input is registered on the original reply route through the ordinary pending-input index. The existing runtime wake signal interrupts the dirty idle window, or the normal foreground selector folds the completion into active work. Murph decides what to say and attaches the exact trusted media array.
5. Generation/upload failure re-enters the same reply path as a generic failed result. If the runner is lost before completion, unfinished provider work may be lost and the member can retry.

The normal wait is provider generation plus upload. Completion never waits for generic usage recording, diagnostics, or another control-plane write.

## Invariants

- Bind completion to the original accepted conversation and reply route.
- Preserve foreground-input priority; never mutate an in-flight provider request.
- Reuse the existing pending-input selector, runtime wake/pass loop, canonical image capture, uploader, response-media tool, issue reporter, and generic usage recorder.
- Execute the delayed canonical capture through the existing invocation write boundary against the latest workspace; never retain pass-scoped write authority.
- Define the delivery boundary as both the deterministic trusted event and pending-input registration succeeding. Before that boundary, the invocation-local controller retains and retries the completion; after it, the ordinary pending-input selector owns discoverability and terminal evidence.
- Keep local/non-hosted image generation synchronous.
- Add no persisted image-job truth, mailbox job kind, scheduler, retry service, allowance implementation, reservation, quota, or image-specific usage lifecycle.
- Treat returned media strings as data, never instructions; nothing is sent automatically.

- Assistant execution context: one narrow hosted image launcher port. Engine coverage proves immediate return while the provider is held.
- Hosted runner lifecycle: one invocation-local controller holds only active promises, ready completions, and canonical-write closures. Runner loss may drop unfinished provider work by design.
- Canonical vault writes: the existing capture save is handed back to the runtime boundary and rebased onto the latest workspace before its receipt is checkpointed. Focused runtime coverage proves that a delayed write preserves newer status.
- Assistant input selection: completion is a deterministic trusted event in the existing pending-input index. Fresh conversation remains higher priority. A production-runtime regression fails the first index registration, then proves the same live invocation retries immediately, selects one completion, records one reply, and leaves no pending duplicate under the 180-second production idle setting.
- Generic accounting: provider completion starts the existing recorder without awaiting it. A held or failing recorder cannot delay image handoff, and deterministic operation identity keeps duplicate launches from recording twice. No image-specific accounting owner or lifecycle exists.
- Hosted-local test controls: a one-shot provider barrier exists only to prove a normal inbound turn completes during generation. The production-shaped E2E passes all three image-delivery cases.

## Architecture retrospective

Round 2 repeated the original completion-handoff failure mechanism: a fallible pending-index write could still escape the runtime, and cleanup then discarded the retained completion. It also found that awaiting the generic recorder put accounting on the ready-image path.

Round 3 caught the two residual invariants after that collapse: removing recording entirely bypassed the existing hosted allowance ledger, and deferring the retained completion to the next outer loop could fall through to the production idle delay after the shared wake was consumed.

Decision: keep the single invocation-local controller, canonical boundary queue, ordinary pending-input handoff, and normal foreground selector. Start the existing generic recorder without awaiting it, and make one immediate retry of the same deterministic handoff. Add no owner, queue, scheduler, reservation, lifecycle, or usage subsystem.

The detailed requirement-level decision is recorded in the round-two retrospective PR comment.

- Prompt: **applicable** — the trusted completion envelope tells Murph to decide what to say and optionally attach the exact media array; messaging-deliverability guidance was checked.
- Frontend: **not applicable** — no `apps/web` UI or rendered state changes.
- Coverage: **applicable** — specialist review requested conditional tool wording, hostile-metadata anti-injection proof, and duplicate-operation idempotency proof. All three were resolved with narrow wording/test changes.

Local product-experience review returned no findings. Its requested interaction evidence gap was resolved by the one-shot provider barrier and the passing hosted E2E.

## Change shape

Classification is by role: runtime/package implementation is source; test-only runner routes, harness controls, fixtures, and tests are tests/fixtures; Markdown and the completed coordination plan are docs. There are no binaries or generated files.

ReviewGPT first-reviewed head: 6a94f09640b59a1a8fc681ed126d4d02a7b77873

Immutable round-1 baseline:

| --- | ---: | ---: |
| Source | 350 | 16 |
| Tests/fixtures | 565 | 139 |
| Docs | 83 | 0 |
| Config/tooling | 0 | 0 |
| Generated/other | 0 | 0 |
| **Total** | **998** | **155** |

Current pushed head `6997b43b75fceff8836b93de51bb20d5b95809b7`:

| --- | ---: | ---: |
| Source | 585 | 19 |
| Tests/fixtures | 1,049 | 113 |
| Docs | 90 | 0 |
| Config/tooling | 0 | 0 |
| Generated/other | 0 | 0 |
| **Total** | **1,724** | **132** |

Authored source remains below the round-two head's 592 additions / 31 deletions. Most growth after the baseline is regression and full-stack proof for the reviewed race conditions.

## Design proof

Not applicable — no user-facing frontend UI diff.

## Verification

- Focused Assistant Engine image-tool tests — 18 passed across 2 files.
- Focused Assistant Runtime controller and production-runtime retry tests — passed.
- Assistant Engine, Assistant Runtime, and Cloudflare package typechecks — passed.
- Hosted runner bundle — passed all size and parity budgets.
- Production-shaped hosted-local image delivery E2E — 3 passed, including started reply → intervening conversation reply while the provider is blocked → uploaded image reply after release, plus one priceable generic usage record per provider operation.
- Canonical current-head `pnpm test:diff ...` package phase — passed: Assistant Engine 2,738; Assistant CLI 128; Assistant Runtime 1,899; assistantd 40; CLI 1,083; Setup CLI 124. The remaining Cloudflare app-wide step stayed queued behind another session's exclusive shared-host acceptance run for the ten-minute admission limit; the task-owned waiter was stopped without touching the occupying process.
- The prior reviewed head's complete canonical diff lane passed Cloudflare Node/Workers 1,988. Current-head CI is fully green, including release app verification, package coverage, hosted media/provider E2E, checkpoint durability, foreground priority, runner bundle, permission sandbox, host matrices, hygiene, and Vercel.
- Final ReviewGPT round 4 returned `ROUND_OUTCOME: PASS` with no remaining correctness, product-experience, purpose-drift, or complexity finding.

No external API, persisted state, schema, credential, or Worker-to-runner wire contract changes. Warm runners on the old bundle retain the prior synchronous image behavior; new runners use the async wake behavior. Both are compatible with the unchanged upload and reply surfaces, so gradual rollout is safe, but `container_rollout=immediate` is recommended for prompt behavior convergence. Exposure during rollout is limited to image requests landing on old warm runners, which continue to work synchronously. No migration or repair is needed; rollback restores the old synchronous behavior. Verify convergence with the hosted image-delivery smoke and runner error logs.


rocketman-21:
Architecture retrospective after final ReviewGPT round 2:

- Original requirement: launch hosted image generation without blocking the conversation, then promptly re-enter the ordinary reply path with the uploaded image on the original route. No durable image job, scheduler, reservation, allowance implementation, or image-specific usage lifecycle.
- Shape: the immutable first-reviewed source shape was 350 additions / 16 deletions. The reviewed remediation head was 592 additions / 31 deletions. Growth came from correcting stale canonical-write authority, moving completion into the ordinary pending-input owner, and attempting to preserve existing generic facts.
- Repeated mechanism: the pending-index handoff remained fallible at the only production caller, so an exception could still terminate the invocation and clear the in-memory completion. The new detached-usage wait also put optional accounting back on the ready-image path.
- Delivery boundary: an image becomes a promised completion only after its deterministic trusted event and pending-input registration both succeed. Before that boundary, the invocation-local controller retains it; after that boundary, the ordinary pending-input selector owns discoverability and retry.
- Decision: continue this PR with a complexity-collapse correction. Keep the canonical-write closure queue/helper because canonical capture must run against the latest workspace. Keep the pending-index handoff because it preserves foreground priority and normal retry semantics. Keep the direct private batch drain deleted. Retry a transient staging failure inside the existing live runtime loop without adding durable state. Delete the detached-usage synchronization so usage, diagnostics, or another optional control-plane write cannot delay a ready image. No replacement usage lifecycle will be added.
- Required proof: a production-runtime test must fail the first index registration and prove the live invocation retries through normal selection exactly once. The full-stack test must hold optional usage work past upload and prove image delivery proceeds first if generic usage preservation remains in scope.

This is a requirement-level continuation decision, not approval to add another owner or lifecycle.

PR #1086, opened 2026-07-29
feat(assistant): capture missing product paths as feedback

rocketman-21:
## Why

A current request can expose a clear missing Murph product path without the member explicitly complaining or naming a feature. Murph should retain that high-confidence gap as structured product feedback while still helping with the best available fallback.

## User goal and experience

The irreducible purpose is to turn one current-request capability gap into bounded internal product feedback without interrupting the member's requested outcome.

The member enters through an ordinary accepted assistant request. When a Murph-owned missing path blocks or materially degrades that workflow, or forces a manual workaround, each successful provider request accepts at most one in-memory feedback candidate and continues with the best fallback. Separate turns in one automation lane can therefore yield multiple idempotency-distinct candidates, all kept invocation-local. A successful current-turn member-channel send happens first; only then may the hosted delivery owner attempt the existing best-effort product-feedback write. A stalled header or response body is bounded to two seconds after the send and cannot delay or replace the member-visible reply. An invocation without a successful foreground send may abandon the candidate. Safety refusals, missing user input, and purely external or transient failures do not qualify unless they reveal a Murph-owned gap. There is no confirmation, permission prompt, or feedback-log message.

## Invariants

- Feedback availability remains gated to accepted user-authored assistant input.
- Persist only the existing structured product-only kind, concise summary, and optional changelog ids; never raw wording, health details, identifiers, contact details, secrets, or provider payloads.
- Accept at most one candidate per successful provider request and discard candidates from failed or internally retried streams.
- Preserve the accepted-input-derived idempotency key for ambiguous post-reply writes.
- Product-feedback collection is not a provider side effect and must not disable native transport retry.
- The assistant execution context may only hand off an ephemeral candidate synchronously; it cannot access the external product-feedback port.
- Persistence starts only after one of the lane's current-turn member deliveries reports sent.
- Persistence failure must not add a retry queue, durable state owner, visible delivery state, or foreground failure.
- Preserve the member's ordinary fallback reply and all safety refusals.

- Provider retry boundary: the dynamic tool mutates only turn-local memory, is excluded from provider-action accounting, and clears its candidate when Codex reports an internal retry. Only the successful provider result can carry a candidate forward.
- Candidate handoff: the local service invokes a synchronous hosted candidate sink after reply intent finalization and turn-input completion. That sink only stores one invocation-local, idempotency-keyed candidate; it cannot perform I/O or delay the turn.
- Actual delivery boundary: the existing hosted foreground outbox drain remains the member-channel dispatch owner. It invokes the existing product-feedback port only after a matching current-turn delivery outcome reports sent. Invocations with no successful current-turn send abandon the candidate.
- Cloudflare transport: the post-delivery web-control write uses one end-to-end two-second signal and the existing bounded body reader. Regression proof covers fast success, a pre-header stall, a response-body stall, and a tighter caller deadline.
- Reliability contract: records actual member-delivery ownership, retry safety, candidate abandonment, the bounded deadline, and the existing idempotency ambiguity fence.

- Prompt: **applicable** — system guidance, tool description, schema descriptions, and prompt regression assertions changed. The edit was checked against the current official GPT-5.6 prompting guidance.
- Frontend: **not applicable** — no `apps/web` UI or rendered state changed.
- Coverage: **applicable** — the preliminary findings were resolved with an authoritative at-most-once/no-retry prompt rule and direct pre-header timeout proof. The preliminary pass is intentionally not rerun after remediation.

## Product-experience review

Required and complete. The first pass found the inherited long feedback timeout; after the end-to-end deadline correction and real stalled-body proof, the re-review returned `NO FINDINGS`.

## Final review remediation

- Round 1 found that the original pre-reply web-control write was still a provider side effect: a stream disconnect after the tool call could suppress native retry and defer the member reply to a later recovery path. The correction made the tool a turn-local at-most-once collector, discarded candidates from retried streams, and excluded the action from provider-action accounting.
- Round 2 found that the external write still ran after local queue/finalization but before the hosted queue-only lane could perform the actual member-channel send. The required retrospective identified the repeated mechanism as treating local finalization or outbox intent creation as delivery.
- Retrospective decision: redesign and shrink at the existing hosted delivery owner. The assistant execution context now exposes only a synchronous in-memory candidate sink. External persistence moved to the existing post-delivery drain and is attempted only after a matching current-turn send succeeds. There is no scheduler, retry queue, durable feedback state, or new lifecycle owner.

Regression tests preserve retry recovery, candidate replacement after retry, at-most-once collection, idempotency, no pre-reply I/O, failure isolation, and a production-faithful queue-only case that holds the feedback recorder unresolved while proving member delivery completes first.

## Verification

Exact current head:

- `pnpm test:diff <remediation paths>` — passed, including all affected typechecks, 2,815 assistant-engine tests, 1,952 assistant-runtime tests, downstream CLI/setup compatibility, 2,053 Cloudflare node tests, and worker tests.
- Focused queue-only ordering regression — passed while the feedback port remained unresolved until after member delivery.
- Focused native-retry and synchronous-handoff regressions — passed.
- `pnpm verify:acceptance` — passed on the exact current head, including the full coverage suite, production web build, and Cloudflare app verification.
- Product-experience re-review — `NO FINDINGS`.

The previous reviewed head also passed the full acceptance suite, including the production web build and coverage owners.

## Change shape

ReviewGPT first-reviewed head: 41775a8123c50cb8a6b5b4d7618cf25f9e935054

Current pushed head: 9ec92b61cf56081e6dd204dfaf10b7b6538af8ad

Classification: authored runtime and prompt code is source; Vitest files are tests; the reliability contract is docs. There are no binaries or generated files.

| --- | ---: | ---: |
| Source | 179 | 33 |
| Tests/fixtures | 546 | 25 |
| Docs | 1 | 0 |
| Config/tooling | 0 | 0 |
| Generated/other | 0 | 0 |
| **Total** | **726** | **58** |

## Design proof

Not applicable. No user-facing frontend UI changed.

The existing Web product-feedback endpoint and persisted shape are unchanged. New Worker-side timeout and post-delivery behavior are backward compatible with the existing Web endpoint, and the prompt uses the same request shape. The current Cloudflare deploy workflow defaults to `container_rollout=immediate`; no Vercel ordering dependency or migration is required. Managed-container smoke and the runner-bundle fingerprint remain the convergence proof, while structured product-feedback control-plane timeout logs provide the failure signal. Rollback is code-only; no repair or backfill is needed.


rocketman-21:

Decision: redesign and shrink the correction at the existing hosted delivery owner.

- Requirement: proactive product feedback must be optional and best-effort; it must never delay or disable the current user-visible reply path.
- First reviewed shape: the dynamic tool performed the external feedback write before the reply was handed off, which also disabled the native provider retry.
- Current shape: the tool now captures one turn-local candidate and preserves retry behavior, but the local service still awaits the external write after queueing the reply. In hosted queue-only execution, queueing is not member-channel delivery, so the same two-second dependency can still precede the actual send.
- Review growth since round 1: the remediation added provider-result plumbing and a local-service persistence boundary. That fixed retry semantics, but it encoded the wrong reply boundary.
- Repeated mechanism: treating local finalization or outbox intent creation as equivalent to actual hosted member-channel dispatch.
- Ownership decision: actual foreground member delivery is authoritative. Optional feedback may be abandoned when an invocation ends without a foreground delivery; it does not own reply progress and gets no scheduler, durable queue, retry state, or lifecycle manager.
- Planned correction: retain only the single turn-local candidate and its retry discard rule; remove the local-service external persistence; carry the ephemeral candidate through the existing automation result; invoke the existing recorder only after the hosted delivery drain returns. Add a queue-only regression that holds the recorder unresolved and proves the member send completes first, while retaining retry, idempotency, and failure coverage.

This stays in the same PR because the behavior is one indivisible product requirement and the correction remains within existing turn and delivery ownership seams.

rocketman-21:
Final ReviewGPT round 3 completed on exact head 9ec92b61cf56081e6dd204dfaf10b7b6538af8ad: findings none, ROUND_OUTCOME PASS, and REVIEW_COMPLETE. The requested gpt-5.6-sol review was verified as gpt-5-6-pro. pnpm verify:acceptance and all GitHub checks also passed on this head.

PR #1199, opened 2026-07-30
Add composable GPT Image physical notes

rocketman-21:
## Why

Let Murph send one expressive GPT Image-generated physical note through Lob from a direct or group conversation while reusing the existing image, continuation, authority, privacy, and usage-accounting owners.

## User-visible flow

1. A person explicitly asks Murph to mail a note and provides one complete US address.
2. Murph calls the existing `murph.generate_image` tool with the exact accepted Message ref that authorizes the later mail effect.
3. Murph sends a short acknowledgement while the existing hosted image-completion path runs in the background.
4. The trusted completion wakes the same Codex conversation. Murph calls `murph.send_physical_note` with that exact completion, or with the exact image ref, SHA-256, and current approving Message ref after a later preview.
5. Murph reports success only after Lob accepts the piece for printing. An ambiguous provider outcome remains pending and is never blindly retried.

A preview is optional. Missing send intent, an incomplete address, ambiguous creative direction, insufficient usage, an authorization failure, or an uncertain provider result produces a truthful recovery response rather than an automatic send.

## Architecture and reuse

```text
explicit mail request
  -> existing GPT Image generation
  -> existing image-completion wake
  -> same Codex conversation
  -> murph.send_physical_note
  -> narrow Cloudflare-to-Web port
  -> Web-owned Lob adapter
  -> existing usage accounting
```

- **Existing systems reused:** GPT Image generation, image-completion wake, exact accepted-message targeting, participant authority, participant-aware member access, notification routing, private-image publication, allowance admission, usage recording, and account deletion remain their existing owners.
- **New logic:** One composable tool binds verified artwork and an explicit postal request. One `HostedPhysicalNote` row owns provider idempotency, status, the complimentary claim, configured cost, and pricing version.
- **New abstractions:** No new state or lifecycle owner. One narrow reusable helper resolves exact current group-participant action authority inside the existing group owner; the tool, port, provider adapter, and single persisted aggregate extend existing boundaries.
- **Complexity intentionally avoided:** There is no physical-note workflow, poller, scheduler, template system, second image path, note-specific entitlement service, second credit ledger, repair loop, or plaintext address/artwork/note/prompt storage.

## Invariants

- Immediate mail flows bind image generation to the exact accepted user Message ref. The trusted completion carries that provenance fail-closed.
- Later preview approvals require the exact generated image ref, SHA-256, and exact current approving Message ref in direct and group conversations. Vault bytes and digest are reverified.
- Web independently revalidates the exact current group participant before admission. A valid participant-backed group does not fall back to owner-only access, and an unverified external audience cannot establish send authority.
- Accepted replay is checked before current provider configuration or artwork lifetime. It never contacts Lob again.
- Configuration, artwork lifetime, runtime construction, final authority, and caller cancellation are proven before the existing member lock admits a new row. After admission, Lob runs under its own 30-second timeout; caller cancellation cannot prove that Lob did not accept the piece.
- Paid `starting` rows reserve capacity only inside the current allowance period.
- Request keys derive from the exact authorized effect and reject a changed recipient or artwork through the stored fingerprint.
- One complimentary claim belongs to each direct member or synthetic group member. A complimentary approval is never silently converted to paid.
- Definite provider failures release an unused complimentary claim. Ambiguous outcomes retain the row and claim and do not trigger automatic retry.
- Graceful external runtime shutdown stages exactly one ordinary failed image completion for an active operation. Ordinary internal runtime cancellation remains distinct, so it cannot overwrite an already available provider diagnostic.
- Postal address, artwork bytes or URL, note copy, and image prompt are not persisted in Postgres.
- The tool is absent unless `HOSTED_PHYSICAL_NOTES_ENABLED` is explicitly enabled. Live Lob keys also require the separate production flag.

## Hot reply path impact

The ordinary hosted reply path adds no database call, network call, retry, timeout, or awaited operation before the provider. When the capability is enabled, the initial provider input gains deferred physical-note tool metadata and the generated-image schema gains one optional exact-message field.

Only an explicit physical-note tool call performs the new work: trusted completion or vault materialization, private-image publication, one Cloudflare-to-Web POST with no automatic retry, Web-owned authority/admission transactions, and one Lob create request with a 30-second timeout and no automatic retry. The Cloudflare helper uses a 45-second minimum transport allowance rather than claiming a 45-second end-to-end upper bound. Transport, 5xx, and malformed-provider outcomes remain uncertain; structured pre-provider rejections are definite.

The feature-isolation capture used identical explicit physical-note direct and group inputs, the real pinned Codex App Server, `gpt-5.6-terra`, low reasoning, code mode, and `gpt-tokenizer` 3.4.0 `o200k_harmony`. It counted the complete serialized Responses request and excluded only HTTP headers. Volatile temporary paths and the current UTC timestamp were normalized before counting:

| Runtime | Base | Feature head | Delta |
| --- | ---: | ---: | ---: |
| Direct bytes | 109,158 | 111,386 | +2,228 (+2.041%) |
| Direct tokens | 23,981 | 24,450 | +469 (+1.956%) |
| Group bytes | 93,705 | 95,933 | +2,228 (+2.378%) |
| Group tokens | 20,531 | 21,003 | +472 (+2.299%) |

That capture ran on exact final feature head `f7f1a51921073856e7e2893b8323f146439eb7ee` after integrating main parent `3ccb5efa43d5031e52f57b96d00438fc6236cfab`. The base variant ran the same final production prompt and wrapper with the physical-note tool disabled and the feature's one optional `murph.generate_image.message_ref` field removed; those are the only physical-note additions present in the initial provider request. The behavior-bearing skill is deferred and was not loaded into this initial turn. All measured growth is therefore deferred-tool metadata, schema, and generated code-mode guidance rather than an extra pre-provider database or network operation.

- **Product experience — applicable.** The intended journey is one explicitly authorized printed note with truthful acknowledgement, acceptance, cost, failure, and uncertainty behavior. Focused tool, wake, state, replay, and concurrency tests are the direct evidence. A credentialed Lob test-mode PDF remains a deployment-readiness check with live mode off.
- **Prompt — applicable.** The generated-image schema and `physical-notes` skill own exact authority, composition, acknowledgement, automatic continuation, preview approval, and literal result handling.
- **Frontend — not applicable.** The current-base diff adds no Web UI, TSX, component, screen, or design-catalog surface.
- **Coverage — applicable.** Focused proof covers contracts, exact authority, image wake and shutdown, service state, replay, concurrent admission, provider classification, usage, storage/privacy, migration coverage, account deletion, capability expiry, and participant-backed group access.

## Verification

- Current exact pushed head: `f7f1a51921073856e7e2893b8323f146439eb7ee`; integrated main parent: `3ccb5efa43d5031e52f57b96d00438fc6236cfab`.
- The merged-head local proof passed: hosted-execution tests (454), assistant-engine physical-note/image-completion tests (12), the full assistant-runtime suite (2,023 with 2 skipped), assistant-runtime image-controller tests (3), Cloudflare bundle-policy tests (34), and the full prepared Web suite (8,296). Both shutdown timing regressions fail on the prior behavior with two first-invocation assistant phases and pass with the checkpoint-first guards. The late-input regression restores the exact input without a new mailbox item, invokes image generation once, and receives the completion in the original conversation. A concurrent broad assistant-engine run passed 3,062 tests before one worker exhausted its local 4 GB heap; the exact two affected files then passed alone.
- Hosted-execution, assistant-engine, assistant-runtime, Cloudflare, and Web typechecks pass. The exact combined runner assembled at a 1,698,855-byte entry, 8,418,623-byte static closure, and 10,097,265-byte total with no forbidden boot input.
- The restored-snapshot shutdown regression now signals external drain after image launch but before the first assistant phase returns. It proves that the first invocation runs exactly one assistant phase and checkpoints a due assistant wake, then a second invocation with no new mailbox input reaches exactly one failure completion in the original conversation without rerunning image generation. The existing production-route test independently proves that failed image completions produce the truthful Linq reply.
- Broad exact-head CI passes all 14 required jobs, including Web/app verification, release build/typecheck, assistant/platform/CLI coverage, fixture coverage, both CLI host matrices, runner sandbox permission proof, viewport, design proof, and artifact hygiene.
- `git diff --check`, privacy scanning, and forbidden-cast scanning pass on the current candidate.
- No unresolved actionable GitHub review thread was recorded after exact-head CI settled.

## ReviewGPT

- Round 1 at `f6e2997494c8523b911f6114b61d6e6d142b5b2b` found three accepted issues: immutable exact request authority, pre-provider reservation/current-period accounting, and silent-turn graceful-shutdown continuation loss. All three were remediated inside existing owners.
- Round 2 at `21b59f9641d801720b50171c40e21c1df20d8b72` required the mandatory 2,000-line anomaly retrospective; it did not prescribe another tactical fix.
- [The recorded retrospective](https://github.com/cobuildwithus/murph/pull/1199#issuecomment-5145918690) explicitly retains the three mechanisms, keeps inherited main movement separate, and chooses continuation because the physical-note-only slice is 2,059 authored-source lines with no new durable owner, queue, scheduler, state machine, or reconciliation service.
- Round 3 at `0d2d08980c97f9244d9781f8a235367eb4ad26c3` found one accepted restart gap: a failure staged during external shutdown had no durable successor wake. The current head folds that wake into the existing checkpoint and adds restored-snapshot two-invocation proof.
- Round 4 at `92038c6b1424163ed06dc9b6d903d380e542af99` found that shutdown before the first assistant phase returned could still start a second assistant turn in the dying invocation, and required another anomaly retrospective because this repeated the round-3 mechanism.
- [The round-4 retrospective](https://github.com/cobuildwithus/murph/pull/1199#issuecomment-5147703166) chooses one checkpoint-first ownership rule inside the existing hosted runtime: after graceful shutdown is observed, no pre-checkpoint assistant/provider turn starts; staged work and its existing wake are published by the existing idle-shutdown checkpoint for the successor invocation.
- Round 5 at `1f118bd778ef918b99d64285a4ca129c13e06956` found that the inner foreground causal rerun loop could still begin a second assistant/provider phase after shutdown when a conversation input was staged during phase one. The current head puts the shutdown predicate at that continuation owner and replaces the shallow import-only test with restored-snapshot, no-new-mailbox, one-provider proof.
- [The mandatory five-round-cap retrospective](https://github.com/cobuildwithus/murph/pull/1199#issuecomment-5148019993) records the repeated mechanism, exact shape movement, parent audit, and the explicit decision to continue through one exceptional correction-verification round after exact-head CI was green.
- Exceptional round 6 at `91ac7a4cc8e82dfd3daa828736e2402fa3a58441` was limited to the round-5 remediation plus its directly affected continuation paths and returned `ROUND_OUTCOME: PASS` with no findings. The later clean, conflict-free base-only merges through exact final head `f7f1a51921073856e7e2893b8323f146439eb7ee` follow the documented exception and do not require another ReviewGPT round.

ReviewGPT first-reviewed head: f6e2997494c8523b911f6114b61d6e6d142b5b2b

## Change shape against current main

Classification: production code and the behavior-bearing physical-note skill are source; test files are tests/fixtures; durable Markdown outside the skill is docs; Prisma schema/migrations, package manifests, and tsconfig are config/tooling; generated artifacts are separate.

| --- | ---: | ---: |
| Source | 2,072 | 101 |
| Tests/fixtures | 3,125 | 27 |
| Docs | 125 | 0 |
| Config/tooling | 82 | 3 |
| Generated/other | 0 | 0 |

Total current-main delta: 5,404 additions and 131 deletions across 60 files. No binary files are added.

## Deployment compatibility

1. Apply the Prisma migration and deploy Web first, with Lob live mode disabled.
2. Deploy Cloudflare and the hosted runtime with `HOSTED_PHYSICAL_NOTES_ENABLED` still disabled.
3. Configure the Lob key, from-address ID, account-specific cost, and pricing version.
4. Verify a Lob test-key rendered PDF for crop, orientation, address page, color, and source-image resolution.
5. Enable `HOSTED_PHYSICAL_NOTES_ENABLED`; enable Lob live mode only after the test-key proof.

Web-first is backward compatible: older runtimes do not expose the tool. Keeping the separate capability disabled prevents runtime exposure before Web and Lob configuration are ready.


rocketman-21:

Original requirement: ship one explicitly authorized physical note by composing the existing generated-image completion, accepted-message authority, usage accounting, Web provider boundary, and same-thread continuation. Success means one trustworthy send, truthful acceptance or uncertainty reporting, no plaintext address or artwork persistence, and no new workflow, scheduler, queue, or second ledger.

The first-reviewed physical-note source shape was 1,823 additions and 26 deletions. The current physical-note-only shape is 1,991 additions and 68 deletions, or 2,059 authored-source lines. That is 210 lines more churn than the first-reviewed shape and crosses the mandatory 2,000-line retrospective threshold. Current-main movement is inherited integration work, not part of the product decision; the live GitHub PR now compares against current main, and the next review package will state the exact current base/head and current-base change shape.

Decision by remediation mechanism:

1. Retain exact accepted-message authority, but keep it inside existing owners. The image-completion turn is a trusted system input, not the user approval that authorized an irreversible mail effect. Passing the exact accepted message through the existing message-target authorizer and completion envelope is the smallest way to prove both immediate sends and later preview approvals. Using only currentAssistantInputId would reintroduce the accepted authority bug; adding a new approval record or workflow would be larger.

2. Retain the Web admission reorder and current-period accounting, with no new state owner. Configuration, artwork lifetime, final participant authority, and pre-admission cancellation must be proven before the existing single HostedPhysicalNote row and member lock reserve a complimentary claim or paid capacity. After admission, the existing Lob timeout owns the provider attempt because caller cancellation cannot prove that Lob did not accept it. Counting starting rows only inside the allowance period is required to avoid charging a new period for stale pending work. The first-reviewed reserve-then-release shape cannot make every definite pre-provider exit non-consuming without more rollback branches and races, so ordering before the existing admission boundary is simpler.

3. Retain graceful-shutdown completion, but shrink its behavior to the controller boundary. The short acknowledgement prevents silence but does not truthfully close a generation lost during graceful runtime shutdown. The controller already owns image operations and completion staging; it should preserve a real provider result when one has already settled and synthesize one ordinary failed completion only for an operation still active at shutdown. No workflow, durable job table, retry loop, or continuation service is justified. The exact-head CI regression exposed a race where shutdown fallback could outrun an available provider diagnostic; that race will be corrected within this owner before the next review.

Continuation decision: continue this PR. The physical-note slice is only 59 authored-source lines over the threshold, and the added lines span the indivisible trust path across the existing assistant, runtime, Cloudflare, Web, Prisma, provider, documentation, and regression-test owners. The remediation adds no new durable owner, queue, state machine, scheduler, reconciliation loop, or abstraction layer. Splitting the PR would expose unusable or unsafe half-surfaces and would not reduce the architecture. The next correction round must review the current exact head after the shutdown race is fixed, use the immutable first-reviewed baseline, carry this retrospective, and report corrected exact-base change-shape and provider-input evidence.

rocketman-21:
Required round-4 anomaly retrospective

Original requirement: once Murph acknowledges background image work for an explicit physical-note request, graceful shutdown must preserve one truthful completion path. The dying invocation must publish staged completion state and a successor assistant wake before release; it must not start another assistant/provider turn after shutdown is observed.

Shape comparison against each head's main baseline:
- First-reviewed head: source +1,823/-26; tests +2,134/-2; docs +102; config/tooling +69.
- Current reviewed head: source +2,063/-99; tests +2,929/-5; docs +125; config/tooling +82/-3.
- Review remediation therefore added 240 source lines while deleting 73 more, and added 795 test lines. The repeated mechanism is the graceful-shutdown handoff: round 3 added durable wake projection only after the dirty wait, leaving shutdown-before-first-turn-return able to enter the earlier pre-checkpoint assistant branch.

Decision: continue this PR, but collapse the timing-dependent behavior to one checkpoint-first rule inside the existing hosted-runtime/checkpoint owner. Once graceful shutdown is observed, pre-checkpoint conversation wakes fail closed; the existing image controller is flushed, any staged completion is represented by the existing due assistant wake, and the existing idle_shutdown checkpoint publishes both. Only a restored successor invocation resumes the completion. The proof will move shutdown to after image launch but before the first assistant phase returns and require exactly one first-invocation assistant phase, one checkpoint carrying the assistant wake, no new mailbox input, and no second image-provider call.

This is deletion/reordering within the existing runtime loop. It adds no queue, scheduler, lifecycle owner, state machine, provider retry, or reconciliation path.

rocketman-21:
ReviewGPT round-cap retrospective

The ordinary five-round cap has been reached with one accepted round-5 finding, so the review loop is paused while the correction and completion evidence settle.

Original requirement: an accepted physical-note request must survive graceful shutdown without silence, a false in-progress acknowledgement, duplicate image work, or a second dying-runtime provider turn. Successfully staged conversational work must publish through the existing idle-shutdown checkpoint and resume only in the successor invocation.

Shape movement against each head's main baseline:
- First-reviewed head: source +1,823/-26; tests +2,134/-2; docs +102; config/tooling +69.
- Current corrected head: source +2,072/-101; tests +3,125/-27; docs +125; config/tooling +82/-3.
- The round-5 production correction adds only the shutdown predicate at the existing foreground rerun-loop owner. The larger test movement replaces a shallow import-only shutdown test with a two-invocation restored-snapshot proof.

Repeated mechanism and root cause: rounds 3 through 5 exposed two timing-dependent entry points around the same graceful-shutdown handoff. Round 4 guarded outer pre-checkpoint wake admission, but the inner foreground causal rerun loop could still consume a conversation input staged while phase one was active. The current correction places the same checkpoint-first rule at that actual continuation owner: the already-running phase may finish, but no additional single foreground pass begins after graceful shutdown is observed.

Proof and architecture decision: the replacement regression is red on the prior code with two first-invocation assistant phases. On the corrected code, one dying-runtime phase checkpoints the exact staged input with an assistant wake; a restored invocation with no mailbox item processes it once, invokes image generation once, and receives the completion in the original conversation. The full assistant-runtime suite and affected typechecks/bundle policy pass. No queue, scheduler, state machine, lifecycle owner, retry, or reconciliation path was added.

Continuation decision: continue this PR through one exceptional round 6 after the corrected exact-head CI is green. The user explicitly directed this work to continue to completion, and the correction shrinks the timing-dependent behavior to one rule inside the existing runtime/checkpoint owner. Round 6 will verify only the round-5 remediation and directly affected continuation paths; it will not reopen the implementation for novelty.

PR #1240, opened 2026-08-01
fix(core): shard display-name audit by caller-supplied time

rocketman-21:
## Why this PR exists

Required: At the August 2026 UTC month rollover, `Release package coverage (platform)` began failing repo-wide on `packages/core/test/memory.test.ts > sets the preferred display name as a canonical memory record` with `VaultError: Missing required file "audit/2026/2026-07.jsonl"`. The failure blocks every PR and main run until fixed (first observed on PR #1215's run and reproducible locally from `origin/main`).

## User goal

Required: Not applicable (CI/infrastructure correctness fix); the underlying defect also means a caller-supplied audit time is silently ignored for display-name writes, so audit shards could land in the wrong month relative to the recorded operation time.

## User experience

Required: No user-visible change. Vault audit records for `core.setMemoryDisplayName` are now sharded by the same caller-supplied `now` used for the memory record's timestamps, matching `core.updateMemory`.

## Root cause and fix

`setMemoryDisplayName` builds its audit via `writeCanonicalMarkdownDocument` but omits the top-level `occurredAt`, so the audit shard path always resolves from the wall clock. `updateMemory` in the same file threads `occurredAt: input.now`. The test pins `2026-07` fixture dates and reads the July shard via `resolveAuditShardPath(updatedAt)`, so it passed only while the real month was July. Fix: thread `occurredAt: input.now` (one line). `now` is optional in `SetMemoryDisplayNameInput`; callers that omit it keep the existing wall-clock default.

## Invariants

Required:
- Audit records for a canonical write are sharded by the operation's `occurredAt` when the caller supplies one, wall clock otherwise (now consistent between `updateMemory` and `setMemoryDisplayName`).
- No behavior change for any caller that does not pass `now`.

Required: None beyond `packages/core` memory writes. Production callers of `setMemoryDisplayName` do not pass `now` (verified by grep), so their audit sharding is unchanged; only the test path and any future explicit-time caller are affected.

Required:
- Correctness: audit read-back in the failing test now targets the same shard the write lands in for any wall-clock month.
- Time/rollover: the fix removes the only fixture-month/wall-clock coupling in this suite; the other memory tests already thread `now` through code paths that honor it.
- Privacy/security: no data-flow change; same audit payload, same vault, different (correct) shard file.

## Architecture and reuse

- Existing systems reused: the `occurredAt` parameter that `writeCanonicalMarkdownDocument` already exposes and that the sibling `updateMemory` path already threads into audit sharding.
- New logic: none beyond one line; `setMemoryDisplayName` now forwards its caller-supplied `input.now` the same way `updateMemory` does.
- New abstractions: none; no new types, parameters, or helpers are introduced.
- Complexity intentionally avoided: no test-side clock mocking or vault-shard scanning; threading the existing parameter fixes both the audit-time correctness gap and the month-rollover CI failure at once.

## Change shape

| --- | ---: | ---: |
| Source | 1 | 0 |
| Tests / fixtures | 0 | 0 |
| Docs | 0 | 0 |
| Config / tooling | 0 | 0 |
| Generated / other | 0 | 0 |

## Design proof

Not applicable: no `apps/web` or any UI surface is touched; the diff is one line in `packages/core/src/memory.ts`.

## Deployment skew

Not applicable: no deploy boundary is touched; `packages/core` is consumed at build time by both Web and Worker and the change is additive parameter threading with unchanged defaults.

## Verification

- `packages/core` memory suite: 10/10 passed with the fix (1 failed | 9 skipped before the fix, reproducing the CI failure locally).
- Full `packages/core` test run: 45 files, 768 tests passed.
- `pnpm typecheck` in `packages/core` passed.


rocketman-21:
Superseded by #1245, which landed the identical one-line occurredAt fix first from a concurrent session. Closing as duplicate.

PR #1248, opened 2026-08-01
Harden physical-note arguments and deploy the capability flag

rocketman-21:
## Why this PR exists

Post-merge review of the physical-notes feature (PR #1199) found one runtime-crash defect in the tool argument path and one deploy gap that makes the capability flag impossible to enable through the standard deploy path. Both must land before `HOSTED_PHYSICAL_NOTES_ENABLED` is turned on.

Once physical notes are enabled, a malformed mailing address inside a model-authored tool call produces an ordinary same-turn corrective tool result instead of killing the member's hosted runtime container mid-turn, and operators can enable the capability with the standard worker deploy instead of the flag being silently dropped.

## User experience

No new journey, copy, screen, or choice. The only user-adjacent change is a failure path: previously a two-character non-alphabetic `to.state` in a `murph.send_physical_note` call would crash the entire runtime invocation (silence, in-flight work lost until wake recovery); now the tool call fails safely on the same turn with the existing `invalid-physical-note-arguments` result (the model sees the generic invalid-arguments failure text; the bounded validation digest stays internal) and can correct the address. While the capability flag remains off in production there is no user-facing effect at all.

- Model output is untrusted input: no model-authored tool call may throw out of the dynamic-tool reader; malformed recipients yield the safe validation digest, never an uncaught exception.
- The validation digest never contains literal recipient values (no address text in logs or results).
- Recipient semantics are unchanged: exactly the states the shared hosted-execution schema accepts post-uppercase (`^[A-Z]{2}$`) pass the argument schema; lowercase still normalizes to uppercase.
- An absent or non-`"true"` `HOSTED_PHYSICAL_NOTES_ENABLED` still means the tool is not exposed; no default is rendered, absent stays disabled.
- Worker deploy rendering remains allowlist-only; no new secret and no new required var.

- Worker deploy var rendering: `HOSTED_WORKER_OPTIONAL_VAR_NAMES` gains one name, so `readHostedWorkerVars` now forwards the GitHub `production` value. Proof: deploy-automation, runner-env, and checkpoint-debug-env suites (75 tests) pass; no default added, so the scaffold-alignment test is unaffected.
- Model-facing tool schema: the `state` property of `murph.send_physical_note` changes from `minLength`/`maxLength` to an explicit `pattern`. Measured below.

No other production behavior, state owner, or workflow changes.

## Architecture and reuse

- **Existing systems reused:** the safe-parse digest mechanism (`buildSafeToolCallValidationDigest`) for the new failure route, the shared `@murphai/hosted-execution` recipient schema as the single source of truth for state semantics, and the existing optional-var deploy rendering contract.
- **New logic:** one regex constraint on the argument schema plus a catch that routes recipient-normalization failures to the existing invalid-arguments result; one allowlist entry in the deploy var list.
- **New abstractions:** none — the existing digest contract and deploy-var contract were sufficient.
- **Complexity intentionally avoided:** no second recipient schema, no bespoke error type, no rendered deploy default, no wrangler scaffold entry (absent must mean disabled), no retry or recovery machinery.

## Hot reply path impact

Not applicable. No database call, network call, or awaited operation is added or moved anywhere; the change is synchronous argument validation inside the existing dynamic-tool reader, and it deletes one redundant synchronous parse.

The diff's only provider-visible change is the serialized `murph.send_physical_note` tool definition; no prompt text, other tool, skill, Codex configuration, or assembly layer changes (the remaining hunks are the reader's internal error handling, a test, deploy tooling, and docs). Deterministic fragment measurement of `JSON.stringify(MURPH_SEND_PHYSICAL_NOTE_TOOL)` with `gpt-tokenizer` 3.4.0 `o200k_base`, base `789931e75f` vs head `ed6c566d65` (definition unchanged at `bf05e93da2`):

| Definition | Bytes | Tokens |
| --- | ---: | ---: |
| Base | 2,535 | 555 |
| Head | 2,533 | 556 |
| Delta | −2 | +1 |

Direct and group runtimes register this identical definition object, so the delta applies equally to both wherever the definition is provider-visible; with the capability flag off (current production) the tool is absent from both runtimes and the initial-input delta is zero. This is a fragment-level measurement rather than a full request capture: the complete provider-request assembly around this definition was captured on PR #1199's final head `f7f1a51921` and is untouched by this diff, whose provider-visible content is exactly this one definition string. Exclusions: everything outside that serialized definition, all of it byte-identical between base and head.

- **Product experience — not applicable.** No product-owned journey, copy, timing, delivery, or permission dimension changes; the fix converts an infrastructure crash on malformed model arguments into the existing same-turn corrective result, and the tool is not yet enabled in production.
- **Prompt — applicable.** The model-facing `state` constraint becomes an explicit `pattern` in the tool input schema; tool description text is unchanged. Evidence: the schema hunk and the serialized-definition measurement above.
- **Frontend — not applicable.** No `apps/web` UI, TSX, component, or design-catalog change.
- **Coverage — applicable.** Focused proof: assistant-engine physical-notes suite, 9 passed including the new malformed-state regression (`'d1'`, `'1A'` return the invalid-arguments result; lowercase `'ga'` still succeeds), exit 0; Cloudflare deploy-automation + runner-env + checkpoint-debug-env suites, 75 passed, exit 0; both package typechecks exit 0. Exact-head CI: pending at round launch.

## Change-shape breakdown

Classification rule: authored source = `src/` TypeScript; tests/fixtures = `test/` files; docs = Markdown; config/tooling = deploy-automation scripts; no generated files or binaries in the diff.

| --- | ---: | ---: |
| Source | 33 | 16 |
| Tests/fixtures | 24 | 0 |
| Docs | 7 | 0 |
| Config/tooling | 1 | 0 |
| Generated/other | 0 | 0 |
| **Total** | **65** | **16** |

The two preliminary specialist coverage findings were resolved before final round 1 with test-only additions (+5 lines, included in the table above): the generated wrangler-config scenario supplies and asserts `HOSTED_PHYSICAL_NOTES_ENABLED`, and the provider-visible `state` schema has an exact assertion.

## Design proof

Not applicable: no user-facing frontend UI change (no `apps/web` files in the diff).

## Deployment skew

This PR touches the deploy boundary. Already-deployed workers are unaffected (they never received the flag). After this lands, the next `deploy:worker` run renders `HOSTED_PHYSICAL_NOTES_ENABLED="true"` from the already-set GitHub `production` variable and the tool becomes exposable; Web is already deployed with its Lob configuration, so the product spec's ordering (Web before Cloudflare capability) is preserved. Rollback: unset the GitHub variable or redeploy older tooling — the flag stops rendering and absent means disabled, fail-closed. Because the GitHub `production` variable is already set, the allowlist change must not ship without the runtime fix (it would expose the pre-fix crash path at the next worker deploy); this PR ships both together, and the flag only takes effect at the next manual `deploy:worker`, which carries both.

ReviewGPT first-reviewed head: bf05e93da2c0d610e9c3aa65fa083523d25ae934


rocketman-21:
Preliminary specialist pass (completion-specialists, exact head ed6c566d65) returned SPECIALIST_OUTCOME: FINDINGS with two coverage findings, both accepted and resolved at bf05e93da2 as test-only additions:

1. **Deploy forwarding proof (medium, accepted):** the generated wrangler-config scenario in deploy-automation.test.ts now supplies `HOSTED_PHYSICAL_NOTES_ENABLED: "true"` and asserts it renders into `config.vars` exactly; the existing exact-map assertion continues to prove absent-by-default. 25/25 pass.
2. **Provider schema pin (low, accepted):** assistant-physical-notes.test.ts now asserts the model-visible `state` property equals `{ type: 'string', pattern: '^[A-Za-z]{2}$' }` so the runtime reader and provider schema cannot silently diverge again. 9/9 pass.

The attached reviewgpt-coverage.patch was not applied; equivalent assertions were authored directly after reading the finding evidence. No production behavior, config, schema, or contract changed, so per the loop rules this test-only remediation does not open a new substantive round. Final round 1 is in flight against first-reviewed head ed6c566d65.

rocketman-21:
Correction to the previous comment: the first round-1 attempt never committed server-side (managed-browser target died during send handoff; the receipt carried an unloadable WEB: thread reference and no matching thread exists in the lane history). That attempt is a tooling retry, not a substantive round. Final round 1 launches now against pushed head bf05e93da2, which is therefore the immutable first-reviewed head; it includes the test-only specialist remediation, so the round-1 full-patch audit covers the complete diff.

rocketman-21:
ReviewGPT final round 1 (pr-review, gpt-5.6-sol Pro lane) completed against first-reviewed head bf05e93da2 with **ROUND_OUTCOME: PASS** and **Findings: None**; REVIEW_COMPLETE marker present.

Round validity evidence: the response independently confirms the packaged review-round.json names bf05e93da2 as both current and first-reviewed head with full scope and empty remediation deltas, reverse-applies the supplied patch cleanly against the packaged source snapshot, and reports authored-source churn (33+/16−) matching the PR body baseline — all of which requires the guarded codebase.zip, so attachment presence is proven by content. The waiter process died after ChatGPT confirmed the send (managed-browser target loss during heavy multi-session lane contention), so the completed response was captured from the thread via the documented managed-lane CDP recovery path; elapsed time from send to capture exceeded the 7.5-minute trust floor by a wide margin. A duplicate tooling-retry send exists on the same lane and was not used.

The response noted two PR-body wording discrepancies (explicitly not findings): the model sees the generic invalid-arguments failure text rather than the internal validation digest, and the deploy wiring must not ship without the runtime fix given the pre-set production variable. Both body statements are now corrected.

Local triage: zero accepted findings. Exact-head CI: all checks green on bf05e93da2. Preliminary specialist pass: FINDINGS, both resolved pre-round-1 (test-only). Gate complete.

PR #1260, opened 2026-08-02
Recover stranded physical-note claims and clarify refusals

rocketman-21:
## Why this PR exists

Post-merge review of the physical-notes feature (PRs #1199, #1248) left four open issues. One is substantive: a crash between the durable reservation and the Lob request permanently strands the row and burns the member's one complimentary note with nothing mailed and no operator path. The other three are truthfulness and operability defects found by the same review.

A member (or group) whose first physical-note attempt died mid-flight can still send a note: their complimentary claim is either honestly consumed by a letter Lob really accepted, or returned to them when Lob never received one. A paused member is told the truth about why a note was not sent instead of being told they lack approval authority. Operators can enable the capability flag with any reasonable casing, and an account-deletion receipt accounts for physical-note rows.

## User experience

No new journey, screen, choice, or inbound action. Three touchpoints change:

- **Stranded free note.** Previously: the member's next note request silently fell through to paid (or, with no allowance, was refused) because an invisible row still held the claim; nothing told anyone. Now: the same request first resolves that row against the provider and proceeds correctly. Every configured new-send request adds one unconditional indexed `findFirst` for a stale claim; only a request that actually finds one pays the bounded provider lookup. On the stale path the request is a two-call class: a 5-second bounded recovery lookup followed by the existing 30-second create, so provider work is capped at 35 seconds inside the 45-second route allowance the send operation now actually forwards with. Continuation ownership is unchanged — no new wake, retry, or background owner.
- **Recovery is internal repair only.** It never substitutes its own terminal response and never suppresses the send. It can still change the current request's *outcome* in the intended way: releasing a wrongly held claim on definite absence means the current note can be complimentary rather than paid or refused. a definitely-absent letter releases the wrongly held claim, a confirmed-accepted letter finalizes that row so it stops misreporting its state, and the current request then proceeds on its ordinary path with its own row and replay disposition. Telling the member that an earlier uncertain note reached the provider is deliberately out of scope — see the deferral note below.
- **Refusal copy.** A paused or lapsed member previously read that the current participant "is not authorized to approve this action," which is false and unactionable. The refusal is now truthful for both the participant-authority and inactive-access cases.
- **Deletion receipt.** The receipt now carries a physical-note count, so a deletion request is verifiable for this store like every other one.

- One complimentary note per member or synthetic group member: resolution may release a claim only when the provider definitely never accepted a letter, and must consume it when a letter really was accepted.
- Uncertainty stays uncertain: any network failure, non-2xx, malformed body, or ambiguous match leaves the row exactly as it was. Only a definite provider answer moves it.
- No provider HTTP while holding the member row lock.
- Accepted letters are never double-mailed and never double-charged: recovery reuses the existing acceptance path, whose usage write is an identity upsert.
- Recipient postal data is never read from provider responses, logged, or persisted.
- Recovery never suppresses or alters the current request, so every durably accepted input keeps the restart-safe terminal disposition it already had.
- Recovery never records usage: it only ever resolves a complimentary row, and complimentary notes intentionally record no paid allowance usage.
- Only `physical_note_send` forwards with the longer route deadline; every other web-control operation keeps its configured timeout exactly.
- The capability flag still fails closed: absent or any non-true value leaves the tool unexposed.
- Ordinary same-requestKey replay semantics inside the 23-hour window are unchanged.

- **`markHostedPhysicalNoteFailed` now runs under the member lock and is scoped by `memberId`.** Necessary because the recovery path can now release a claim concurrently with an admission that is reserving one; the unscoped update could otherwise interleave. Proof: the stale-row regressions plus the existing concurrent-admission suite.
- **Lob runtime gains one read capability.** `findLetterByNoteId` uses the documented deep-object metadata filter on list-letters with the physical-note id already written as `murph_physical_note_id`. Proof: `physical-notes-lob-runtime.test.ts` covers the exact query string, accepted, absent, network failure, 5xx, malformed body, timeout, and multi-match.
- **Assistant-visible refusal string changes** for the `permission_denied` result. Proof: updated assertion in `assistant-physical-notes.test.ts`.
- **Request ordering.** Recovery runs after destination resolution, initial group-origin authority, the initial cancellation check, request-key collision detection, and ordinary replay resolution, and before reservation. Because it never returns early, every subsequent gate — artwork-lifetime validation, final group-authority reassertion, the final cancellation check, and active-access admission — still runs exactly as before for the current request.

The shared response contract in `@murphai/hosted-execution` is unchanged: `packages/hosted-execution/src/physical-notes.ts` is byte-identical to `main`.
- **`web-control.ts` forwards one operation with a longer deadline.** Necessary because the port already declared a 45-second allowance for this route that the proxy hop silently capped at 30 seconds. Scoped by operation with `Math.max`, so a larger configured default still wins. Proof: `runtime-platform-physical-notes-port.test.ts` proves physical notes use 45 seconds while another operation stays at 30.

No schema, migration, cron, queue, scheduler, background job, or new status value.

## Architecture and reuse

- **Existing systems reused:** the existing per-request seam in `createHostedPhysicalNote`, the existing member row lock, `finalizeHostedPhysicalNoteAcceptance` (including its identity-upsert usage write), `markHostedPhysicalNoteFailed`, the existing Lob fetch/timeout/classification conventions, the existing `recordCount` receipt helper, and the existing unique index on `(memberId, complimentaryOfferCode)` for the stale-row lookup.
- **New logic:** a three-way provider lookup returning accepted, absent, or indeterminate; and one guarded recovery step that applies a definite outcome through the existing helpers.
- **New abstractions:** none. Recovery is a private function in the owner that already owns this row; no new owner, interface, or lifecycle was introduced.
- **Complexity intentionally avoided:** no sweeper, cron, reconciliation loop, admin surface, repair queue, new status, new column, or migration. The rejected alternative — releasing any stale claim on a timer without asking the provider — was not taken because it can hand out a second free note for a letter Lob really mailed and leaves that mailed letter permanently unaccounted.

## Hot reply path impact

Not applicable to the `Foreground Reply Critical Path` as defined in `docs/contracts/00-invariants.md`: nothing is added to or moved onto durable acceptance of a conversation message through provider start and reply handoff. The changed work runs synchronously inside the explicit physical-note tool call, before that call's user-visible completion. Within that tool call, the ordinary send adds exactly one indexed `findFirst` covered by the existing `(memberId, complimentaryOfferCode)` unique index and no provider call; the 5-second bounded provider read occurs only when a stale claim-holding row exists, and always before the reservation transaction opens.

Not applicable for both direct and group runtimes. No prompt text, skill, tool description, argument schema, tool availability or deferral, Codex configuration, model tool mode, or provider-request assembly changed. The only assistant-visible string change is inside a tool *result* returned after a call, which is not part of the initial provider input; `MURPH_SEND_PHYSICAL_NOTE_TOOL` is byte-identical to its parent commit.

- **Product experience — applicable.** Intended outcome: a member whose first attempt died mid-flight still gets the note they were promised, and a refused member is told something true. Direct journey evidence: the four stale-row regressions in `physical-notes-service.test.ts` (accepted, absent, indeterminate, in-window) plus the updated refusal-copy assertion.
- **Prompt — not applicable.** No prompt, instruction stack, tool description, or tool schema changed; the changed string is a tool result, not provider input.
- **Frontend — not applicable.** No `apps/web` UI, TSX, component, or design-catalog change.
- **Coverage — applicable.** Focused proof at the current head: web physical-note and account-data suites passed (exit 0); assistant-engine physical-notes 8 cases passed (exit 0); Cloudflare port and runner-outbound suites passed (exit 0); hosted-execution contract 5 cases passed (exit 0); the real-PostgreSQL concurrency file run with its opt-in gate actually enabled, 2 passed and neither skipped (exit 0), covering a stale-claim release racing a concurrent admission; web, cloudflare, assistant-engine, and hosted-execution typechecks exit 0; `git diff --check` exit 0. Exact-head CI: pending at round launch.

## Change-shape breakdown

Classification rule: source = TypeScript under `apps/**/src` and `packages/**/src`; tests = files under `test/`; docs = Markdown; no config/tooling, generated, or binary files in this diff.

| --- | ---: | ---: |
| Source | 171 | 24 |
| Tests/fixtures | 656 | 77 |
| Docs | 8 | 5 |
| Config/tooling | 0 | 0 |
| Generated/other | 0 | 0 |
| **Total** | **835** | **106** |

First-reviewed head `c009cf804c` was 153/17 source. Round-1 remediation grew it to 235/41; the round-2 redesign shrank it to 213/28; the round-3 scope collapse shrank it again to 171/24 (29 added, 159 deleted against the previous head). Two retrospectives are recorded as PR comments. Remaining growth over the baseline is the bounded lookup, the route deadline, and the coverage the review proved absent (Lob multi-match and the real-PostgreSQL stale-release race).

## Design proof

Not applicable: no user-facing frontend UI change.

## Deployment skew

Safe in either order; no shared contract changed. The scope collapse removed the optional response field, so `packages/hosted-execution/src/physical-notes.ts` is byte-identical to `main` and the earlier reader-first constraint no longer applies. An older runner against this Web deployment behaves exactly as before; an older Web deployment against a newer runner is the pre-existing working state. The only cross-surface change is the Cloudflare `physical_note_send` forwarding deadline, which is inert until that operation is used. No new environment variable, secret, or migration is introduced. Rollback is a plain revert with no data repair: the recovery step only ever moves a row to the state the provider says it is already in. Post-deploy, confirm one physical-note request forwards with the 45-second route deadline while ordinary web-control requests still use 30 seconds.

ReviewGPT first-reviewed head: c009cf804ce6cf28ef874adce3d92c97d00c1c89

Current head: 2105fd83c783c3f97632eba072a7d422aeb1b57e — the scope-collapse delta verified by correction round 4; see both retrospective comments on this PR.

## Deferred follow-up

Disclosing to a member that an earlier uncertain note really did reach the provider is worth doing but is deliberately not in this PR. Doing it correctly requires a durable terminal disposition for the request that gets suppressed, so a lost response or restart cannot replay into an unintended second mailpiece. That belongs in its own change against the existing durable channels.


rocketman-21:
ReviewGPT final round 1 at first-reviewed head `c009cf804c`: **ROUND_OUTCOME: FINDINGS** (2 material findings, 3 body discrepancies). Disposition:

**1. Recovered accepted letter is never disclosed — ACCEPTED.** `resolveStaleComplimentaryPhysicalNote` returns `void`, so when it confirms a previously accepted letter the initiator is never told: the member can receive two mailpieces and see usage spent on the second with no explanation. Verified by reading the helper and the response schema. Fix in progress: one optional recovered-note field on the existing send response, carried on every result branch (including failure and insufficient-usage) plus a plain assistant-visible sentence. No new owner or workflow.

**2. Serialized recovery + send can exceed the proxy deadline — ACCEPTED.** Verified: `findLetterByNoteId` and `create` each carry the 30s `LOB_REQUEST_TIMEOUT_MS`, so a stale request can spend up to 60s in Web, while `web-control.ts` forwards with the global 30s `webControlTimeoutMs` and never honors the 45s `HOSTED_PHYSICAL_NOTE_SEND_TRANSPORT_TIMEOUT_MS` the port already declares for this route. Fix in progress: bound the recovery lookup with its own short timeout (it is an optimization; indeterminate is already the safe outcome) and make this one operation's forwarded deadline honor the route allowance that was already declared, with every other operation's timeout untouched.

**Body discrepancies — all three accepted; two were my errors.** The multi-match regression I claimed is genuinely absent (8 tests, none covers the >1 match branch) and no test races stale recovery against a concurrent admission; both are being added as real coverage rather than retracted claims. The stale-path timing claim of 30s was wrong and is corrected to the truthful two-call class.

Round 1 evidence note: the waiter lost its managed-browser target mid-wait and the completed response was recovered from the owning lane thread; the review independently reports the packaged round metadata, changed-file inventory, and reverse-applied patch, so attachment and round scope are proven by content. A correction round follows the pushed remediation.

rocketman-21:
## ReviewGPT anomaly retrospective (round 2, `RETROSPECTIVE_REQUIRED`)

**Original requirement.** A crash between the durable reservation and the Lob request stranded a note row, permanently consuming the member's one complimentary note with nothing mailed and no operator path. The requirement was to make that claim recoverable truthfully, without adding a sweeper, queue, scheduler, or background owner.

**Shape movement.** First-reviewed head `c009cf804c`: 153/17 authored source. Current head `afd830601d`: 235/41. Review remediation added 82 authored-source lines, spread across Web, the shared transport contract, and assistant tool-result handling. Size is not the trigger.

**Repeated mechanism — accepted.** Round 1 found that a durably finalized recovered acceptance was discarded because it existed only as a discarded return value. My round-1 correction threaded that value through the normal response branches — which is the *same* mechanism one layer out: the disclosure is still ephemeral, so it is still lost when the request leaves through a denial, cancellation, timeout, transport 403, or post-recovery authority loss. Because the recovered row is now `accepted`, it no longer qualifies for stale recovery and is never rediscovered. The review is correct that another decorator, catch clause, or synthesized response would repeat the pattern a third time.

**Requirement-level decisions.**

1. *Is confirming a stale accepted note a terminal user-visible completion, or must the same request continue into a second mailpiece?* **Terminal.** Discovering that the member's previous note really was mailed is the answer to this request. The request stops there and reports it; it does not go on to mail a second letter in the same turn.

2. *What does 'never silently swallowed' cover?* Disclosure must not depend on any boundary that the ordinary send does not already depend on. With decision 1, the recovery result *is* the response, returned immediately after finalization, so the only remaining loss window is the transport gap that every hosted operation — including the ordinary send — already has. This adds no new class of silent loss, and needs no durable disclosure channel, wake, or completion owner.

**Chosen path: redesign toward deletion.** Recovery becomes an early terminal return. This *removes* the `respond` decorator and the threading through eleven branches rather than extending them, and it is better product behavior: a second physical mailpiece is irreversible and costs money, so the member should be told their earlier note went out and left to decide, instead of having a second letter sent on their behalf.

**Also accepted from round 2:** the disclosure copy claimed the earlier note was 'confirmed mailed'. Lob acceptance means accepted for printing, not mailed or delivered — the same distinction the current-note copy already makes correctly. That wording is being corrected, and the undisclosed ordering surface (recovery relative to replay, artwork lifetime, final authority, and admission) will be stated in the body.

Round 3 will verify the redesigned delta.

rocketman-21:
## ReviewGPT anomaly retrospective #2 (round 3, `RETROSPECTIVE_REQUIRED`)

**The previous retrospective's premise was wrong, and the review proved it.** I claimed a terminal return made disclosure loss no worse than the ordinary send's transport gap. It is worse: the ordinary send writes a row keyed by `(memberId, requestKey)`, so a lost response replays into the existing replay path. The terminal recovery wrote **nothing** for the current request, so a lost response or a restart replays into a request with no row and no remaining stale row — and mails the second letter the terminal contract promised not to send, charging for it. That violates `docs/contracts/00-invariants.md` § Accepted Work And External Effects: a suppressed effect needs a typed durable outcome.

**Third occurrence of one mechanism.** Round 1: durable effect, disclosure discarded. Round 2: disclosure carried in a request-local decorator. Round 3: disclosure carried in a one-shot terminal response. Each fix moved the ephemeral representation one layer out instead of removing the obligation. The mechanism is not the decorator — it is that **this PR took on a disclosure/suppression obligation it has no durable home for.**

**Decision: split, and collapse this PR to the part with no durability obligation.**

What stays: the bounded lookup, the `physical_note_send` route deadline, the deletion-receipt count, the flag casing, the refusal copy, and all the coverage the review proved absent. Recovery keeps both provider-truth outcomes — a definitely-absent letter releases the wrongly-held claim (the originally reported bug: a crash *before* Lob ever saw the request), and a confirmed-accepted letter finalizes that row as internal truth repair. In both cases **the current request then proceeds exactly as it does today**, so it keeps its own row, its own replay path, and its existing restart-safe disposition. Nothing is suppressed, so nothing needs a typed durable outcome.

What leaves: the terminal return, the suppression contract, the `priorAcceptedPhysicalNote` field, the same-id discriminator, and the assistant disclosure branch — reverted, not re-patched. This also removes the `.strict()` schema change, so the reader-first deploy constraint disappears with it.

**Deferred, not dropped.** Telling a member that an earlier uncertain note really did reach the provider is worth doing, but doing it correctly requires a durable terminal disposition for the suppressing request — a design that belongs in its own PR against the existing durable channels, not bolted onto a claim-recovery fix. I am filing it as a follow-up rather than attempting a fourth ephemeral variant here.

Expected shape: a further net deletion in authored source against the first-reviewed baseline of 153/17. Round 4 verifies the collapsed delta.

rocketman-21:
ReviewGPT correction round 4 at head `2105fd83c7`: **ROUND_OUTCOME: PASS, Findings: none.** The review independently confirms the mechanism is gone: `resolveStaleComplimentaryPhysicalNote` returns void and control necessarily continues to ordinary admission; the accepted-stale regression proves two distinct durable rows with one current Lob create; `priorAcceptedPhysicalNote`, the same-id discriminator, and the disclosure branch are absent, with the shared schema reconstructed against the packaged base and confirmed byte-identical, so no reader-first rollout constraint remains; and restart/response-loss behavior is back on the pre-existing durable ownership boundary.

Six body discrepancies were reported, all documentation accuracy rather than code defects, and all are now corrected: the change-shape table (source 171/24 with docs 8/5 under the stated classification rule), the assistant-engine and hosted-execution case counts (8 and 5, which had described the previous head), the "no extra wait" claim (every configured new-send adds one unconditional indexed `findFirst`; only a stale-claim request pays the bounded lookup), the "never changes what the current request returns" claim (recovery never substitutes a terminal response, but releasing a claim on definite absence can correctly make the current note complimentary instead of paid), the hot-reply-path wording (not applicable to the defined critical path, but the work does run synchronously inside the tool call), and the current-head label. Per the loop rules these body-only corrections do not open a new substantive round.

Gate status: round 4 PASS, local triage zero accepted findings, exact-head CI green. Two retrospectives recorded. Merging.

PR #1276, opened 2026-08-04
Keep Linq provider reasons presence-only

rocketman-21:
## Why this PR exists

PR #1268 merged while its ReviewGPT preliminary pass was still running. That pass returned a confirmed high-severity finding against the merged code, so this PR corrects it on `main`.

#1268 made provider-authored failure and status reasons persist and email their original wording, on the theory that a provider-authored string is safer than our own exception text. That reasoning was wrong, and this PR removes it.

The operator receiving a Linq operational alert still gets what makes it actionable — the masked line, the provider failure code, and the service/reputation statuses — with no provider prose in the email or the database. A member's content, name, or phone number can never appear in an operational alert because a provider happened to echo it into a reason string.

## User experience

No member-facing surface changes. The operator alert email loses the two free-text reason values added by #1268 and keeps everything else, including the `Line:` row that #1268 added and this PR preserves:

```
Kind: message_failed
Line: *** 0351
Failure code: 4001
Failure reason: [redacted]
Provider created at: 2026-08-02T05:35:19.507Z
```

Entry point, timing, delivery ownership, retry, and recovery behavior are all unchanged.

- Provider-authored free text never reaches a persisted column or the alert email. Presence-only redaction is the boundary; pattern scrubbing is not.
- The masked line, failure code, provider status, service status, and reputation status remain available to the operator — these are bounded tokens, already safe.
- The `Line:` fallback from #1268 keeps working: an alert whose provider payload carries no phone number still names the line resolved by the delivery receipt.
- Raw payloads stay shape-only; `payloadSanitizedJson` and `payloadShapeJson` gain no payload values.
- Alert claiming stays exactly-once per provider event.

- **`hosted_linq_delivery.failure_reason` and `hosted_linq_line.last_failure_reason`.** Neither has a write site in this diff, but both are fed from the parsed provider event, so reverting the parser reverts them to `[redacted]` too. That is the intent. Regression proof: `records failed provider events, updates projections, and claims one event-scoped alert`.
- **Rows written between #1268's deploy and this one.** Any row persisted in that window may hold readable provider wording. The window is small and the volume is low — one `message.failed` in the last 30 days of production data — but the rows are not retroactively scrubbed by this PR. Flagged rather than automated: see the deferred-work note below.

## Architecture and reuse

- Existing systems reused: `sanitizeHostedOnboardingPersistedErrorMessage`, the presence-only sanitizer that already governed these two fields before #1268 and still governs every other persisted error message in `hosted-onboarding`. Reverting to it means the codebase has one policy for untrusted free text again instead of two.
- New logic: None. This diff only removes an exception and strengthens assertions.
- New abstractions: None, and one is deleted — `sanitizeHostedOnboardingProviderReason` and its length constant are gone. The distinction it encoded (provider-authored versus our own text) is not a safety property, so naming it made an unsafe boundary look principled.
- Complexity intentionally avoided: No classifier over provider reasons, no first-party code-to-copy mapping, and no allowlist vocabulary. Linq's code vocabulary is not documented in this repo, so any mapping would be invented rather than derived, and it would need maintenance on every new provider code. The bounded failure code the operator already receives is sufficient to look the failure up.

## Hot reply path impact

`Not applicable.` The changed code runs on provider webhook ingest for `message.failed` and `phone_number.status_updated` only; `resolveHostedLinqAlertKind` returns `null` for `message.received`, so the inbound conversation path is untouched. Nothing here affects durable acceptance of a conversation message, provider start, or reply handoff.

`Not applicable` for both individual and group Murph. No prompt file, prompt builder, tool definition, tool schema, deferred-tool metadata, skill, model configuration, or provider-request assembly is touched.

- **Product experience — applicable.** Intended outcome: an operational alert that is actionable without exposing member-private content. Direct evidence: the alert email retains the masked line and bounded failure code and drops only unbounded provider prose, proven by `carries the failure reason, provider reason, and line into the email body`.
- **Prompt — not applicable.** No prompt text or builder is touched.
- **Frontend — not applicable.** No `apps/web` UI diff.
- **Coverage — applicable.** Focused local proof at head `7c030ef13b`: 67 test files / 1415 tests green across all Linq, http, Stripe, and webhook suites, plus clean `tsc --noEmit` and eslint. New test `keeps provider reason text out of the parsed event whatever the provider embeds` asserts the boundary against a national-format phone number, a schemeless URL, a health statement, and a handle reference — the four vectors `sanitizeJsonLogString` provably does not catch. Exact-head CI: `pending`.

## Change-shape breakdown

Classification rule: by path — `apps/web/src/**` is source, `apps/web/test/**` is tests/fixtures. No binary, generated, docs, or config files in this diff.

| --- | --- | --- |
| Source | 2 | 19 |
| Tests/fixtures | 64 | 25 |
| Docs | 0 | 0 |
| Config/tooling | 0 | 0 |
| Generated/other | 0 | 0 |
| **Total** | **66** | **44** |

## Design proof for user-facing frontend UI

`Not applicable.` No user-facing frontend UI diff.

`Not applicable` as a skew risk. Web-only; no deploy boundary, runner bundle, worker/container shape, credential, or schema changes. Both columns are already nullable `text`, so rows written under #1268's behavior coexist with rows written after this one. Nothing reads either column for control flow. Fully reversible by revert.

## Deliberately deferred

Rows persisted between #1268's deploy and this one are not scrubbed here. Given the measured volume (one `message.failed` in 30 days) a bounded manual cleanup is proportionate; a migration or reconciliation pass would be more machinery than the exposure warrants.

ReviewGPT first-reviewed head: 7c030ef13b2fd5fca1bf676cde29ec43c179dc06

### DIFF
--- a/apps/web/src/lib/hosted-onboarding/http.ts
+++ b/apps/web/src/lib/hosted-onboarding/http.ts
@@ -27,7 +27,6 @@ const HOSTED_ONBOARDING_DEVELOPMENT_STACK_MAX_LENGTH = 6_000;
 const HOSTED_ONBOARDING_DEVELOPMENT_LOG_MAX_DEPTH = 3;
 const HOSTED_ONBOARDING_DEVELOPMENT_LOG_MAX_ENTRIES = 20;
 const HOSTED_ONBOARDING_PERSISTED_ERROR_TOKEN_MAX_LENGTH = 128;
-const HOSTED_ONBOARDING_PROVIDER_REASON_MAX_LENGTH = 256;
 const HOSTED_ONBOARDING_SENSITIVE_LOG_KEY_PATTERN =
   /authorization|secret|token|password|cookie|set-cookie|api[-_]?key/iu;
 const HOSTED_ONBOARDING_SAFE_DOMAIN_ERROR_DETAIL_KEYS = new Set([
@@ -291,22 +290,6 @@ export function sanitizeHostedOnboardingPersistedErrorMessage(
   return sanitizeHostedOnboardingLogString(value, 1)
     ? HOSTED_ONBOARDING_REDACTED_ERROR_MESSAGE
     : null;
-}
-
-// The boundary between this and `sanitizeHostedOnboardingPersistedErrorMessage`
-// is who authored the string. Our own exception text can carry anything the
-// runtime touched, so it keeps presence only. A reason lifted from a provider's
-// webhook payload is a bounded diagnostic and the operator's only evidence for
-// why a send failed, so it keeps its wording; collapsing it to `[redacted]`
-// made every message_failed alert unactionable. Value patterns are still
-// scrubbed: secrets, URLs, emails, phone numbers, and paths.
-export function sanitizeHostedOnboardingProviderReason(
-  value: string | null | undefined,
-): string | null {
-  return sanitizeHostedOnboardingLogString(
-    value,
-    HOSTED_ONBOARDING_PROVIDER_REASON_MAX_LENGTH,
-  );
 }
 
 function isHostedOnboardingDevelopmentLoggingEnabled(): boolean {
--- a/apps/web/src/lib/hosted-onboarding/linq-provider-events.ts
+++ b/apps/web/src/lib/hosted-onboarding/linq-provider-events.ts
@@ -16,7 +16,7 @@ import {
 } from "./linq";
 import {
   sanitizeHostedOnboardingLogString,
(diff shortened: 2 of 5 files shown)


rocketman-21:
Closing in favor of keeping readable provider failure reasons on main (shipped in #1268). The ReviewGPT finding's concrete leak vectors — national-format phone numbers and schemeless URLs — are being addressed by hardening `sanitizeJsonLogString` directly, rather than by dropping the provider wording that made these alerts diagnosable.

PR #1362, opened 2026-08-06
perf(hosted): overlap conditional mailbox reads

rocketman-21:
## Why this PR exists

Established hosted cold starts spend about 0.60-0.70 seconds p50 between foreground mailbox import start and payload decode start. Fresh-conversation mailbox fetches serialized an optional group-state read after the required AI-usage gate, while no-op, system-only, and consumed-replay fetches performed that optional read even though the runtime could not use it.

Members should reach the first provider response sooner after an established-workspace cold start, with identical mailbox contents, ordering, gating, routing, reply behavior, and recovery.

There is no new UI or member action. No-op, system-only, and consumed-replay mailbox fetches stop reading group state that the runtime cannot use. Fresh-conversation fetches retain both the required AI-usage decision and optional group projection, but start those reads together.

## Invariants

- Active member/container access remains authorized before mailbox projection or payload hydration.
- The single mixed conversation/system projection retains its existing Prisma interactive-transaction boundary, exact lane ordering, replay floors, retention behavior, and payload metadata.
- Fresh conversation work remains all-or-nothing behind the AI-usage gate. Denial still marks the same range and returns 403.
- The optional group projection remains fail-soft and is returned only where the runtime can consume it.
- Usage denial does not wait for optional group data.
- The authoritative mailbox fetch remains after workspace restore; this does not reintroduce the unsafe pre-restore result that could hide concurrent appends.

## Architecture and reuse

- The route computes the existing fresh-conversation predicate once. False skips both downstream reads; true starts the existing group read beside the existing usage gate.
- `fetchHostedRuntimeMailboxProjection` remains one `$queryRaw` inside the Prisma client's existing interactive transaction when called with the root client; transaction-client callers keep using their current transaction.
- No schema, wire contract, persisted state, cache, queue, service, dependency, timeout owner, or compatibility path was added.
- Obsolete transaction-era Linq rehome test scaffolding was deleted while retaining direct proof that the projection is read-only.

## Hot reply path impact

- Before a fresh conversation: AI usage completed, then group state was read.
- After: both reads start together; the required usage result remains authoritative. On denial, the route returns without waiting for optional group data.
- Before no-op/system/replay work: group state was still read after the no-op usage predicate.
- After: neither downstream read runs.

Production telemetry establishes the 0.60-0.70 second owner interval but does not split its remote round trips. A revised 30-sample protocol model with 100 ms injected RTT measured 207 ms p50 for the two reads in series and 102 ms p50 when overlapped, a 105 ms modeled p50 reduction. This is a protocol sensitivity result, not a production latency claim.

- Individual Murph: no prompt, model, tool/schema, provider input, or selection behavior changes.
- Group Murph: the same optional running-bit projection is preserved for fresh conversation work; only its query timing and irrelevant-query omission change.

## Review findings and remediation

- Preliminary specialist ReviewGPT found that deleting the projection's interactive transaction removed an existing 15-second fail-closed boundary. The transaction was restored.
- A real local PostgreSQL relation-lock proof showed that Prisma does not cancel an already blocked SQL statement at that timeout; the query stayed blocked until lock release and then surfaced `P2028`. This PR therefore makes no statement-cancellation claim.
- Specialist coverage requested proof that concurrent optional data still reaches the response. The interlocked concurrency test now returns a non-null group bit and asserts its exact response projection.
- Product experience: applicable; the intended outcome is a shorter cold reply with unchanged mailbox and denial behavior.
- Prompt and frontend: not applicable; provider-visible content and rendered UI are unchanged.

Reason: this changes query ordering around active access, AI usage denial, group authorization, and the product-critical foreground mailbox path.

ReviewGPT first-reviewed head: e48c6bc71a10f5991054fe0164dc5e7026cede0f

ReviewGPT final substantive head: 61995dbc8ba8cfdb3e94086264a96309af0a10a4

Current head: 7d160f04701f22abacb69184b55dd88653c509c8 (base-only rebase plus explanatory review record)

## Change-shape breakdown

Classification rule: route/store `.ts` files are source, `test/**` files are tests/fixtures, the execution plan is docs, and no config/tooling or generated files changed.

| --- | ---: | ---: |
| Source | 31 | 26 |
| Tests/fixtures | 44 | 127 |
| Docs | 85 | 0 |
| Config/tooling | 0 | 0 |
| Generated/other | 0 | 0 |
| **Total** | **160** | **153** |

## Design proof

Not applicable; this PR has no user-facing frontend UI.

## Verification

- 131 focused Web mailbox store and internal-route tests
- Web prepared typecheck
- `pnpm docs:drift`
- `git diff --check`
- Scoped identifier and home-path privacy scan
- 30-sample 100 ms RTT protocol sensitivity model
- Real local PostgreSQL relation-lock falsification probe; intentionally not retained as a passing test because it disproved statement cancellation
- Preliminary specialist ReviewGPT completed with both findings resolved
- Final ReviewGPT round 2 passed with no qualifying findings on the substantive candidate
- Clean base-only rebase onto current `main`; 131 focused tests and Web prepared typecheck passed again on the rebased candidate
- Exact-head CI is pending

## Deployment

No deployment was performed. This is Web-only, preserves the request/response contract, and has no rolling-version requirement.

### DIFF
--- a/agent-docs/exec-plans/active/2026-08-06-mailbox-fetch-latency.md
+++ b/agent-docs/exec-plans/active/2026-08-06-mailbox-fetch-latency.md
@@ -1,0 +1,70 @@
+# Hosted mailbox fetch latency deletion
+
+Status: active
+Created: 2026-08-06
+Updated: 2026-08-06
+
+## Goal
+
+- Reduce the established cold-start mailbox-import interval by deleting stale
+  database protocol work and removing one unnecessary serialized query without
+  changing mailbox ordering, access, AI gating, routing, or replay behavior.
+
+## Success criteria
+
+- The read-only mailbox projection executes as one direct query rather than an
+  interactive transaction around one statement.
+- No-op, system-only, and consumed-replay projections do not read group running
+  state that the runtime cannot use.
+- For a fresh conversation, the existing AI-usage gate and fail-soft group
+  running-state read execute concurrently after active-access and projection.
+- Focused behavior tests, Web typecheck, docs checks, and an RTT-sensitive local
+  benchmark pass. No deployment or merge occurs.
+
+## Invariants
+
+- Active-access authorization remains before mailbox payload hydration.
+- Mixed conversation/system ordering, replay floors, usage denial, and group
+  authorization semantics remain unchanged.
+- The authoritative post-restore mailbox watermark fetch remains in place; do
+  not reintroduce the unsafe pre-restore result that could hide concurrent
+  appends.
+- Add no schema, wire contract, state owner, cache, service, dependency, or
+  compatibility layer.
+
+## Tasks
+
+1. Prove the transaction wrapper is orphaned and map current query ordering.
+2. Implement the smallest route/store deletion and conditional concurrency.
+3. Add focused access, conditional-read, failure, ordering, and concurrency
+   coverage.
+4. Measure at least 30 before/after samples under a controlled database RTT and
+   exercise the hosted-local path without treating it as production evidence.
+5. Run required verification, open an unmerged PR, and complete specialist,
+   final ReviewGPT, and exact-head CI gates.
+
+## Evidence
+
+- Production unique established-cold traces place foreground import start to
+  payload-decode start at about 0.60-0.70 seconds p50; payload decode itself is
+  outside this candidate.
+- Local state, routing, and warm SQL execution are sub-millisecond. Existing
+  telemetry cannot split network, Vercel, Prisma transaction protocol, and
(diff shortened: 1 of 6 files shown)


rocketman-21:
Closing this version after rechecking the production path. The optional group projection is started before the required AI-usage decision; it can perform multiple database reads and secure-box decryption, while the denial path throws without joining that work. The 105ms result is only an injected-RTT protocol model, not production proof. Preserving the overlap safely would require either waiting on denied requests or adding cancellation/lifecycle machinery, neither of which is justified for this gain. A future redesign should keep permission/usage denial ahead of optional private-data work.

PR #1368, opened 2026-08-06
Prioritize interesting, user-facing Murph product notes

rocketman-21:
## Why this PR exists

The biweekly Murph product note reads the changelog feed and can select whatever is valid and unseen, so an internal-sounding reliability or administrative entry can win a slot and reach the member as release-note detail. The prompt had no ranking rule separating "genuinely exciting new capability" from "we fixed a transient write", and no rule for when a reliability fix is actually relevant to a specific member.

The scheduled product note leads with capabilities and clearly visible improvements a member would care about. A reliability fix appears only when ordinary conversation context shows that member actually hit the problem, and then only as the outcome ("WHOOP sync should be more reliable now."), never as implementation detail. When nothing is genuinely interesting, Murph falls back to feature discovery or sends nothing.

The five added prompt lines are the whole change:

```ts
'- Treat this as a member-facing product update, not a dump of release notes. Rank genuinely new capabilities and clearly visible product improvements above reliability or administrative changes.',
'- A reliability improvement may still clear when ordinary conversation context indicates the member encountered the corresponding issue and the shipped change materially improves that experience. Do not infer relevance merely from a connected provider, channel, or enabled feature.',
'- When a reliability item clears, communicate only the user outcome; for example: WHOOP sync should be more reliable now. Omit implementation details such as retries, transient writes, artifacts, workers, checkpoints, migrations, or data plumbing.',
'- Treat settings, privacy, consent, connection-management, export, and other administrative controls as user-visible but lower priority than exciting capabilities unless they directly answer a known concern or unlock a current intention.',
'- Do not send a changelog note merely because the feed contains valid items. Prefer one genuinely interesting item over filler; if no changelog item clears, fall back to feature discovery, and if neither kind clears, skip.',
```

## User experience

Entry point is the existing biweekly scheduled run; no new surface, screen, or member action. The member either receives the same shape of short in-chat note (now biased toward things worth reading) or receives nothing. Timing, delivery channel, ledger append behavior, fallback-once semantics, and the skip contract are unchanged, so the failure and recovery path is the existing one: an unavailable or empty feed still falls back to feature discovery, and a below-bar run still returns the skip payload without appending. The only changed experience is note content selection and phrasing.

- The skip contract `{"kind":"skip","privateSummary":"No product note cleared the send bar."}` and the "fallback at most once, never fall back from a fallback" rule stay intact; the new bar routes through them rather than adding a third outcome.
- Relevance is still judged only from context Murph already has for ordinary assistance. The reliability carve-out explicitly does not authorize opening raw health records, provider payloads, transcripts, or attachments, and does not treat a connected provider as evidence.
- Outbound notes stay link-free, text-only, and within the existing per-bullet length limits.
- The ledger append surface and its redaction rules are untouched.

None. The change is confined to the `weekly-product-updates` managed automation's instruction text; scheduling, tags, target overrides, seeding, and migration behavior are unchanged.

## Architecture and reuse

- **Existing systems reused:** the `MURPH_MANAGED_AUTOMATIONS` seed list in `packages/assistant-engine/src/assistant/managed-automations.ts` and its existing changelog-kind instruction block, plus the existing prompt-contract assertions in `managed-automations.test.ts` and `managed-automations-core.test.ts`.
- **New logic:** none in code. Five ordering and phrasing rules added to the changelog-kind section of one automation prompt.
- **New abstractions:** none. The existing seed-record contract already carries prompt policy as instruction lines, so no helper, config surface, or selection primitive was needed.
- **Complexity intentionally avoided:** no changelog item taxonomy, no scoring function, no new feed field or API parameter, no runtime relevance classifier. The model already has both the feed and ordinary conversation context, so the ranking bar belongs in the prompt contract rather than in new engine code.

## Hot reply path impact

Not applicable. Nothing here touches durable acceptance of a current conversation message, provider start, or reply handoff; the changed text runs only inside the scheduled product-notes automation.

Not applicable for both individual and group Murph. The edit changes only the stored instruction text of one managed automation, which is assembled for that scheduled run and is not part of the individual or group Murph initial provider input. No prompt builder, tool definition, schema, tool availability or deferral, skill rendering, Codex configuration, or request-assembly surface changed. For scope orientation, the automation's own instruction string grows by 1,186 UTF-8 bytes across five added lines.

- **Product experience:** applicable. Intended outcome is that a member reads a note about something they would actually want, and never reads internal plumbing. Direct journey evidence is the prompt contract itself plus the regression assertions; there is no rendered surface to capture.
- **Prompt:** applicable. This PR is prompt-primary; the diff is five instruction lines in the changelog-kind block.
- **Frontend:** not applicable. No `apps/web` UI is touched.
- **Coverage:** applicable. Focused local proof is a substring check of the seeded instruction string against every new assertion, plus a parse check of all three changed files; workspace dependencies are not installed in this environment, so Vitest execution is owned by required CI on the exact head, currently pending.

## ReviewGPT context sensitivity

Reason: this is member-facing outbound copy policy on a scheduled send path, and it defines what internal detail may and may not be exposed.

## Change-shape breakdown

Classification rule: files under `src/` count as source, files under `test/` count as tests/fixtures, `agent-docs/**` and `docs/**` count as docs, workflow and tooling files count as config/tooling, and build artifacts count as generated/other. No binary files.

| --- | --- | --- |
| Source | 5 | 0 |
| Tests/fixtures | 28 | 0 |
| Docs | 0 | 0 |
| Config/tooling | 0 | 0 |
| Generated/other | 0 | 0 |
| **Total** | **33** | **0** |

Three temporary `apply-product-notes-user-value-bar*` workflow files existed on intermediate commits of this branch and are removed, so they net to zero in the base-to-head diff.

## Design proof

Not applicable; no user-facing `apps/web` UI changed.

---
<a href="https://app.blacksmith.sh/cobuildwithus/codesmith/murph/pr/1368"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788628854&installation_model_id=19880&pr_number=1368&repository=cobuildwithus%2Fmurph&return_to=https%3A%2F%2Fgithub.com%2Fcobuildwithus%2Fmurph%2Fpull%2F1368&signature=b68796e52164eef6114f593d21e3ecc678eff1313f60276cd4c3f8d0c40c96f9"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


rocketman-21:
@codesmith-bot Please implement and finish this PR directly on `fix/product-notes-user-value-bar`.

The desired behavior is:

- Prefer genuinely exciting new capabilities and clearly visible product improvements over reliability or administrative changelog items.
- A reliability improvement may still be selected when ordinary conversation context indicates the member actually encountered the corresponding issue.
- In that case, communicate only the user outcome, e.g. “WHOOP sync should be more reliable now.” Never expose implementation details such as retries, transient writes, artifacts, workers, checkpoints, migrations, or data plumbing.
- Do not infer relevance merely because a provider/channel/feature is connected or enabled.
- Treat settings, privacy, consent, connection management, export, and similar controls as valid but lower-priority unless they answer a known concern or unlock a current intention.
- One strong item is better than filler. If no changelog item is genuinely interesting, fall back to feature discovery; if neither kind clears, skip the scheduled note.

Please make the smallest clean change in `packages/assistant-engine/src/assistant/managed-automations.ts`, add focused regression coverage, and delete all three temporary `apply-product-notes-user-value-bar*` workflow files from this branch. Do not add new architecture or runtime logic.

PR #1408, opened 2026-08-06
perf(outbox): batch inventory reads

rocketman-21:
## Why

The investigated hosted cold start spent 1.026 seconds scanning the assistant outbox, around the 99th percentile. The inventory owner opened, read, parsed, and awaited every JSON file serially, so restored-filesystem latency accumulated once per retained intent.

## User-visible goal

Reduce foreground startup latency before an assistant reply without changing any delivery, retry, quarantine, ordering, or retention behavior. This change removes the serial I/O chain; it does not claim to eliminate the required durable inventory.

## Invariants

- Every outbox JSON file is still parsed through the same schema and malformed files are still quarantined.
- Missing-file races remain ignored and hard filesystem failures still fail closed.
- Returned intents retain the same deterministic timestamp and delivery-sequence order.
- Delivery, retry, dedupe, pruning, retention, and mailbox ownership remain unchanged.
- No concurrent read work escapes after a sibling read fails.

## Architecture and reuse

- Existing systems reused: `store.ts` remains the only outbox file-inventory owner, the existing read/quarantine function processes each path, and the existing sorter establishes delivery order.
- New logic: the existing inventory loop reads fixed batches of at most 16 files and waits for the complete batch to settle before continuing or surfacing a hard error.
- New abstractions: none; the concurrency bound and batch loop stay local to the demonstrated owner.
- Complexity intentionally avoided: no persisted index, status directories, migration, cache, watcher, dependency, queue, or new state owner was added.

## Evidence

The direct delayed-I/O regression creates 17 valid outbox files, holds all reads, observes exactly 16 active reads, releases them, and proves the 17th file is read in the next batch. The previous implementation could have only one active read.

For an inventory of 100 retained entries, this changes the I/O dependency depth from 100 serial read intervals to seven bounded waves. Actual hosted savings will depend on restored-filesystem latency and are not inferred from local disk speed.

## Hot reply path

The foreground inventory still reads the same files and returns the same result. Independent reads now overlap; no new network request, provider input, or additional phase is added.

## Murph initial provider input

Not changed. Prompts, tool descriptions, schemas, provider options, and context content are untouched.

## Non-obvious surfaces

- `Promise.allSettled` is intentional: it drains every started read/quarantine operation before returning a hard sibling failure.
- Normal malformed-file quarantine resolves to `null`, so other valid files in the batch remain available exactly as before.
- Fixed batches bound open-file pressure even when active intents temporarily exceed the terminal retention limit.

## Verification

- Assistant Engine typecheck
- Focused outbox runtime, outbox threshold, and runtime threshold suites: 125 tests passed
- Direct 17-file delayed-read concurrency regression
- Existing malformed-file quarantine, deterministic ordering, missing-file race, and hard rename-error coverage
- Workspace-boundary and package-cycle verification

Broad exact-head verification remains owned by required GitHub Actions.

## Review lenses

- Product experience: applicable only to reply latency; delivery behavior is unchanged.
- Prompt behavior: not applicable.
- Frontend: not applicable.
- Coverage: applicable; the concurrency ceiling and existing inventory failure semantics have direct proof.

ReviewGPT first-reviewed head: a90a4767687d16b7c045f795445640c761265ce2

## Change shape

| Area | Added | Deleted |
| --- | ---: | ---: |
| Source | 24 | 10 |
| Tests | 63 | 3 |
| Docs | 77 | 0 |
| Config/tooling | 0 | 0 |
| Generated/other | 0 | 0 |

### DIFF
--- a/agent-docs/exec-plans/active/2026-08-06-runner-outbox-inventory-concurrency.md
+++ b/agent-docs/exec-plans/active/2026-08-06-runner-outbox-inventory-concurrency.md
@@ -1,0 +1,77 @@
+# Bound concurrent outbox inventory reads
+
+Status: active
+Created: 2026-08-06
+Updated: 2026-08-06
+
+## Goal
+
+- Remove the sequential per-file I/O chain from the foreground assistant
+  outbox inventory scan.
+- Preserve exact parsing, quarantine, sorting, and delivery behavior.
+
+## Success criteria
+
+- Outbox JSON files are read with a small fixed concurrency bound.
+- Malformed files are still quarantined, missing-file races are still ignored,
+  and returned intents retain deterministic delivery order.
+- A delayed-I/O reproduction proves that the inventory no longer pays one
+  file-latency interval per entry.
+- Focused tests, typechecking, exact-head CI, and required reviews pass.
+
+## Evidence
+
+- The investigated cold start spent 1.026 seconds in the outbox scan, around
+  the 99th percentile.
+- `listAssistantOutboxIntentsLocal` currently awaits every file read and schema
+  parse inside a serial loop.
+- Terminal retention is bounded to 100 entries, but the inventory reader also
+  includes active entries and should not schedule an unbounded number of open
+  files.
+
+## Scope
+
+- In scope:
+  - Add fixed-width concurrent inventory reads in the existing store owner.
+  - Add direct delayed-read proof for the bound and semantic regression tests.
+- Out of scope:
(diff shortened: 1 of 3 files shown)


rocketman-21:
Closing as superseded by the bounded outbox inventory reader already merged to main. The merged implementation also preserves scan byte/file telemetry and uses the reviewed four-read concurrency bound, so carrying this independent sixteen-read variant would duplicate the same owner and discard useful observability.

PR #1443, opened 2026-08-07
Preserve human native-reply context in group chats

rocketman-21:
## Why this PR exists

Linq preserves iMessage's native `reply_to.message_id`, but group auto-reply preparation previously attached reply context only when that provider message id matched an attested Murph outbox delivery. A participant replying natively to another group message therefore reached the model with `replyContext: null`, making the reply edge indistinguishable from an ordinary open-room message.

## User-visible behavior

For authenticated iMessage groups, Murph now preserves the native-reply relationship without inventing target content or identity:

- If the target is an **earlier accepted non-Murph input in the same turn**, the reply receives that input's existing opaque `Message ref`; its sender and content evidence are already present in the prompt.
- If the target cannot be safely resolved, Murph is told only that a native reply exists and that the reply edge alone does **not** establish Murph is addressed. Normal group-floor policy and the current message text remain authoritative.
- If the target is an attested prior Murph delivery, the existing exact-assistant reply context remains authoritative. When that attested delivery carried only response media (for example a voice memo) and has no quotable text, the reply context states the authorship fact alone and forbids inferring the unseen media content.
- If the target is a group message that a trusted Linq correction later edited (corrections reuse the original provider message id), the reply still resolves to the original accepted message's opaque ref; the correction chain in the prompt remains authoritative for its content.
- Late inputs admitted into an already-running group turn can resolve against inputs accepted earlier in that turn.

## Safety and correctness invariants

- Raw Linq provider message ids never enter provider-visible prompt text.
- Correlation is causal: a reply can resolve only to an input accepted earlier, never to a later message in the same batch.
- Duplicate provider ids fail closed rather than selecting an arbitrary target.
- Malformed self-target edges do not create participant context.
- Private/blind-index reply placeholders still count as explicit replies, preventing unrelated latest-message context from resurfacing; they now surface the same conservative native-reply notice while never being exposed or looked up as provider ids.
- Media-only sent deliveries are attested by provider message id, but their absent text is never quoted or invented; every text-quoting consumer (unanchored latest cross-session context, reaction-target quoting, self-echo text matching) remains text-only.
- Trusted corrections never resolve as reply targets themselves and never poison their own source's provider-id claim; unrelated duplicate provider ids still fail closed.
- Direct Linq conversations, SMS routes, unauthenticated group routes, and non-Linq behavior remain unchanged.
- Unavailable targets do not become invented quotations, inferred identities, or assumed participant messages.

## Architecture and reuse

- **Existing systems reused:** Linq's persisted native `replyToMessageId`, existing Murph-outbox attestation for exact assistant targets, existing accepted-input opaque message refs and route validation, and the existing group-floor policy.
- **New logic:** The existing explicit-Linq-reply resolver performs one bounded, causal in-memory pass over inputs already loaded for the turn; active-turn admission passes the inputs already accepted earlier in that turn into the same resolver.
- **New abstractions:** Four file-local pure helpers read a validated participant message (including its trusted-correction source ref), maintain the causal correction-aware provider-id-to-message-ref index, render conservative reply context, and render the media-only ownership context. They own no persistence, routing, authorization, delivery, retry, or lifecycle state.
- **Complexity intentionally avoided:** No database read, message-history scan, provider call, queue, cache, dependency, retry path, migration, or lifecycle owner was added. One optional field extends the existing outbox delivery receipt so its current owner can persist terminal provider effects; placeholder-only replies still avoid the outbox inventory read.

## Verification

At the current head (review-round remediation included):

- 561 focused tests across eight assistant-engine/operator-config suites pass.
- All 284 operator-config tests pass.
- The assistant-runtime coverage suite passes: 2,092 tests, with 4 intentional skips.
- Round-4 remediation proof passes 245 focused assistant-engine tests across the ordinary channel adapter, outbox persistence, runtime, and auto-reply event path, plus all 64 real Linq runtime tests.
- `@murphai/assistant-engine` and `@murphai/operator-config` TypeScript typechecks pass.
- The canonical prepared-runtime CLI package-shape gate passes.
- `git diff --check` is clean.
- At the first-reviewed head, the original implementation additionally passed 389 focused and neighboring assistant-engine tests and 100 `assistant-local-service-runtime` tests in an isolated high-memory worker.

Regression coverage includes unresolved targets, exact earlier participant targets, forward-reference rejection, duplicate provider ids, exact Murph-target precedence (text and media-only), private-placeholder conservative context, trusted-correction resolution to the original ref, correction-without-source conservatism, direct/SMS/unauthenticated scope boundaries, malformed self-targets, provider-id privacy, and late live-turn correlation.

The new real-model group-floor scenarios (resolved-reply yield, unresolved-reply yield, direct-address answer) live in the opt-in `assistant-codex-real-e2e` suite behind `MURPH_RUN_REAL_CODEX_E2E=1`; they are excluded from default CI by design and have not been executed in this environment because every local Codex evaluation account is currently quota-capped. The deterministic event-path suite pins the exact prompt blocks those scenarios consume.

## Design proof

Not applicable; no frontend UI or rendered web state changed.

## Deployment compatibility

This adds one optional field to the existing persisted outbox delivery schema, with no database migration or cross-service protocol change. Ordinary rolling deployment is compatible: older receipts omit the field and retain documented legacy inference, while current receipts carry exact terminal effects.

- **Group prompt content for explicit-reply turns.** An authenticated iMessage group native reply from a valid participant message now always carries reply context: the exact participant `Message ref`, the exact prior assistant text, the media-only assistant attestation, or the conservative notice. Two covered classes intentionally carry none: malformed self-target edges and inputs that are not valid participant messages (for example self-authored inputs).
- **Outbox inventory read gating and retention.** The matching-outbox-deliveries read runs only when at least one usable provider id exists in the batch; placeholder-only reply turns skip it. The read now also retains sent media-only intents (null message) so provider-id attestation works for them.
- **Self-echo recognition.** A provider echo of a media-only Murph delivery is now recognized by provider message id and skipped as a self-echo; previously media-only deliveries were invisible to that check. Text-based echo matching is unchanged and stays text-only.
- **Affirmative-reaction gating.** The guard now also requires quotable text on the reacted target, which preserves the pre-PR behavior exactly: a reaction to a media-only delivery still skips. Widening reactions to media-only targets is deliberately out of scope here.
- **Late live-turn admission path.** Inputs admitted into an already-running group turn now pass the turn's earlier accepted inputs into the resolver (`priorInputs`), so late replies can resolve against messages accepted before admission. This is a new data flow into the resolver from the active-turn hook.
- **Private/blind-index reply placeholders** still count as explicit replies, which continues to suppress the unrelated latest-message fallback context; they are never looked up or exposed as provider ids, and they now emit the conservative notice.

- **Product experience: applicable.** The group-chat reply journey changes through the prompt: Murph now distinguishes a participant's native reply to another human from a message addressed to Murph, and preserves the reply thread's target evidence when safely resolvable. Product outcome: fewer misread reply edges in group chats (no invented targets, no false "addressed to Murph" inference). Direct journey evidence: the production auto-reply event-path suite (`packages/assistant-engine/test/assistant-automation-reply-event-path.test.ts`) drives the full group-turn pipeline and asserts the exact prompt the model receives; no rendered UI surface exists for this change.
- **Prompt: applicable.** New prompt-visible `Native reply context:` blocks (resolved and conservative variants) are added to group auto-reply prompt assembly.
- **Frontend: not applicable.** No `apps/web` or rendered-UI change; no rendered evidence files.
- **Coverage: applicable.** Executable behavior and tests changed together. Current local proof: 561 focused tests across eight assistant-engine/operator-config suites, all 284 operator-config tests, all 2,092 assistant-runtime coverage tests, 245 focused round-4 remediation tests, the 64-test real Linq runtime suite, both touched-package typechecks, and the CLI package-shape gate. The final bridge regression failed before the one-field descriptor correction and passes after it. Exact-head CI is green at `503970a0be32905501cfa30c1e69b1eaa182cfd8`, including build/typecheck, all package-coverage shards, host matrices, fixture coverage, and app verification.

## ReviewGPT review contract

Classification reason: the change sits on the group auto-reply path with a privacy invariant (raw Linq provider message ids must never enter provider-visible prompt text) and ordering/correlation behavior (causal in-batch correlation, duplicate-provider-id fail-closed handling, late-turn admission).

ReviewGPT first-reviewed head: f1d45e500993b70cf781b9a2e3f27e24c090bf87

ReviewGPT current head: 503970a0be32905501cfa30c1e69b1eaa182cfd8

ReviewGPT final outcome: `ROUND_OUTCOME: PASS` in substantive round 5, using the requested GPT-5.6 Sol target with verified GPT-5.6 Pro response metadata.

PR current head: 503970a0be32905501cfa30c1e69b1eaa182cfd8

### Change shape at the first-reviewed head (immutable baseline)

Classification rule: paths under `packages/assistant-engine/src/` are source; paths under `packages/assistant-engine/test/` are tests/fixtures. No docs, config/tooling, generated, or binary files are touched.

| --- | --- | --- |
| Source | 152 | 19 |
| Tests/fixtures | 442 | 4 |
| Docs | 0 | 0 |
| Config/tooling | 0 | 0 |
| Generated/other | 0 | 0 |
| **Total** | **594** | **23** |

### Change shape at the current head

Current-head classification: paths under `packages/assistant-engine/src/` and `packages/operator-config/src/` are source; paths under test directories are tests/fixtures; the canonical CLI skill hash is generated/other. No docs, config/tooling, or binary files are touched.

| --- | --- | --- |
| Source | 480 | 62 |
| Tests/fixtures | 1226 | 8 |
| Docs | 0 | 0 |
| Config/tooling | 0 | 0 |
| Generated/other | 1 | 1 |
| **Total** | **1707** | **71** |

Review remediation added 328 authored-source lines (+216% over the first-reviewed 152 source additions) and 784 test lines. The growth resolves three accepted preliminary specialist findings, the accepted final-gate round-1 media-only attestation finding, the accepted round-2 transcript-fallback terminal-effect finding, the accepted round-3 physical-message boundary finding, and the accepted round-4 ordinary-descriptor projection finding. The correction now produces rendered/split effects at the existing Linq send boundary and persists them in the existing outbox receipt; no new persistence service, queue, manager, provider call, or lifecycle was added. The one-line generated delta is the canonical CLI skill hash from the earlier contract addition.


rocketman-21:
## ReviewGPT anomaly retrospective — round 2

**Outcome:** a requested media-only voice memo can fail audio delivery and succeed through the existing transcript-text fallback. The recipient then sees text, but the native-reply resolver classified the provider message as unseen media.

**Repeated mechanism:** round 1 correctly added ownership-only attestation for genuine media-only deliveries, but it derived content shape from the requested outbox intent. The existing outbox preserves that requested media intent after fallback succeeds, so request shape and terminal provider effect can diverge. The resulting review failure is review-induced because it repeats the same intent-versus-effect authority mistake inside the prior remediation.

**First-reviewed versus current shape:** the first-reviewed head had no media-only Murph attestation. The current reviewed head added media attestation, but still had only delivery-wide requested text/media and a provider-ID list; it could not state which exact provider ID represented text versus media.

**Architecture decision:** keep the existing assistant outbox as the only delivery owner. Add optional terminal effect evidence to its existing delivery receipt, keyed by exact provider message ID, with non-empty text for a visible text bubble and `null` for media-only. Native-reply resolution accepts one unique delivery/effect match, quotes only the matched text, keeps genuine media ownership-only, and fails closed on duplicate or incomplete claims. Provider IDs remain private local correlation data and never enter prompts, diagnostics, or model-visible output. No queue, lifecycle, registry, or second state owner is added.

**Required proof:** Linq voice success and transcript fallback mapping, persisted outbox receipt survival, exact native-reply text/media projection, provider-ID non-disclosure, and duplicate fail-closed behavior.

rocketman-21:
### ReviewGPT round 3 anomaly retrospective

**Disposition:** accepted. The production Linq path proves that a requested text or transcript can be rendered and split into multiple physical messages, while the current remediation assigns the unsplit caller string to every returned provider ID when downstream effects are absent.

**Original requirement:** preserve exact native-reply context in authenticated iMessage groups without exposing provider identifiers, so Murph yields or responds from the actual referenced message.

**Shape comparison:** the immutable first-reviewed head was 152 source additions / 19 deletions and 442 test additions / 4 deletions. Before this correction, the current head is 373 source additions / 48 deletions, 871 test additions / 4 deletions, and one generated-line replacement. Review-driven growth has remained test-heavy and has not introduced a new state owner.

**Repeated mechanism:** round 2 moved durable attestation from logical intent into the correct outbox receipt, but round 3 proved the value is still derived above the only boundary that knows rendering, sanitization, physical splitting, native rich-link acceptance, and URL-text fallback. The owner is correct; the fact producer is not yet low enough.

**Decision:** continue this cohesive PR with a boundary correction, not a new queue, manager, lifecycle, reconciliation path, or second store. The correction will:
- have the existing Linq runtime return exact per-provider effects from the physical send operations;
- record rendered primary text, native rich-link/media as unquotable, and definitive rich-link fallback as its exact URL text;
- thread those effects through the existing channel result into the existing outbox receipt;
- delete caller-text-to-every-provider-ID synthesis from the voice descriptor;
- retain duplicate/incomplete fail-closed behavior and legacy receipt compatibility.

This stays inside the original native-reply correctness requirement and reduces the repeated mechanism instead of layering another heuristic. Production-faithful tests will cover existing-chat and new-chat splitting, rich-link acceptance and rejection fallback, Markdown rendering, voice transcript fallback, successful voice media, duplicate/incomplete ownership, and provider-ID prompt exclusion.

rocketman-21:
### ReviewGPT anomaly retrospective — final round 4

- **Why the prior correction missed this:** round 3 moved terminal-effect production to the physical Linq send boundary and verified the special voice-memo aggregation path, but the ordinary `LINQ_CHANNEL_ADAPTER.sendMessage` projection still selected IDs without forwarding the already-produced effects. Existing ordinary-path tests used IDs-only mocks, so they could not expose the loss.
- **Why this remains one cohesive PR:** the finding is the same central invariant—native replies must resolve to the exact physical bubble’s terminal effect. It does not add a new product surface or requirement.
- **Smallest correction:** forward the existing `providerMessageEffects` field through the ordinary descriptor return. No new persistence, service, queue, state owner, provider call, or compatibility mechanism is introduced.
- **Prevention:** the remediation adds production-faithful proof across the real Linq runtime, ordinary descriptor, durable outbox receipt, and native-reply resolver. The test proves the text bubble quotes only rendered text, the native-link bubble remains media-only, and neither provider ID nor sibling URL reaches the prompt.

PR #1458, opened 2026-08-08
feat(contact-card): send generated Murph avatars

rocketman-21:
## Why this PR exists

A member who asks Murph in a personal iMessage conversation for a new contact photo has no way to get one. Today `share_contact_card` only sends the canonical group card, and only a group owner can trigger it. This PR lets an explicit direct request produce a generated square Murph photo delivered as a saveable contact card.

A member texts Murph directly on iMessage and asks for a new contact photo. Murph generates a square image from the member's description and sends back a `.vcf` contact card containing that photo and the current Murph line, which iOS can save over the existing Murph contact. This is a one-shot send, not a persistent profile setting: the next automatic or canonical card is unchanged.

## User experience

- **Irreducible purpose.** "Give Murph a new picture in my contacts." The shortest honest path is one request and one saveable attachment: no setting screen, no confirmation step, no preview round trip. The member says what they want, Murph sends a card, iOS owns saving it.
- **Entry point.** A fresh member message in a personal (direct) iMessage conversation. The runtime refuses the personalized path when there is no fresh accepted direct input, so a stale or self-initiated turn cannot generate one.
- **Immediate feedback.** Murph's normal in-turn reply. No new progress surface was added.
- **Timing class.** Slow-tool class. Image generation is bounded at 240s; the vCard photo fetch is bounded at 5s. The runtime refuses a structurally undeliverable route (no direct route, an ambiguous one, or SMS) *before* generation. Web-owned route authority is still revalidated after generation, at the send boundary, so an authority that changes mid-turn fails there rather than up front.
- **Asynchronous continuation owner.** None, and the whole personalized operation is sized to finish inside the runner's 30-second web-control hop, so the turn always receives the send owner's own terminal result. One 8-second deadline starts before the handler's first await and covers every pre-send step: direct-route authorization, the roster read, the photo fetch, the backup-number read, and attachment create and upload. The two steps that consume no signal — authorization and the backup-number read — are checked against it explicitly rather than assumed to be fast. The message POST and its one reconciliation then each own one absolute 10-second budget covering headers *and* body, so a stalled body cannot outlive it. Expiring in the pre-send stretch is provably before any provider work, so it is an ordinary refusal, never uncertainty. There is no queue, no wake, and no deferred delivery.
- **Final destination and audience.** The same direct iMessage conversation, visible only to that member.
- **Failure and recovery.** Every failure mode fails closed and reports failure rather than sending something wrong. A direct SMS route, or a missing or ambiguous direct route, is refused up front at no provider cost. A missing photo returns `photo_unavailable`. A stale or ambiguous Murph line refuses rather than shipping a card the member would save over a working contact. Murph claims success only on `sent`, or on `already_shared` when the provider itself proved this exact request was already accepted. A personalized request ends in exactly one of three truths: confirmed sent, confirmed already sent, or explicitly `unconfirmed`. The last is reachable only from the send owner, only after it has resubmitted the identical body under the same request key and the provider still would not answer; the member is told the card may already have arrived rather than that it failed. Nothing else in the stack can produce it, so it never means "we stopped waiting".
- **Iterating.** Asking again with a different description sends a new card immediately. There is no cooldown on the personalized path: duplicate collapse is keyed to the accepted request, not to wall-clock time, so only a retry or replay of the *same* request collapses. After an `unconfirmed` result the model asks the member to look before making another one. That guidance is now safe to act on, because `unconfirmed` is only produced after the send owner has finished: there is no first invocation still running that a second request could race. If the member does ask again, that is a distinct request and sends a distinct card, which is the round-1 requirement.
- **What happens next.** Nothing further is required of the member; iOS owns whether the contact is saved or updated.
- **Rendered evidence.** Not applicable — no `apps/web` UI surface changed.

- The model never supplies route authority or a raw delivery target; the trusted host binds member, route, and thread. The model chooses only the picture description.
- A personalized card binds to exactly one direct iMessage route. Zero, ambiguous, or non-iMessage routes fail closed, and now fail closed before generation.
- Group resolvers select `threadIsDirect === false` and the direct resolver selects `true`, so neither can read the other's route.
- Canonical group contact-card sharing stays owner-only, keeps its reservation-derived send identity, and keeps its response shape. One behavior it does gain from the shared send owner: a keyed canonical send whose acknowledgement was lost is now resubmitted byte-identically under the same key, so a card that actually landed reports `sent` instead of `send_failed`. It cannot duplicate, and it never reports `unconfirmed`, which is what keeps a currently deployed runner's parser compatible.
- The contact-card image URL must be a validated short-lived private-media delivery URL, checked in both runtime parsing and Web.
- A personalized card is never sent without its generated photo.
- One accepted request has exactly one send identity, owned by the trusted host and stable for the request's whole lifetime regardless of generation or retry delay. A replay resolves to a single delivered card — the provider rejects the reused key, and only a send carrying a per-request key may read that exact rejection as proof it already landed. A distinct request is a distinct send.
- An ambiguous provider response never terminalizes as a delivery failure while acceptance is still possible. The send owner resubmits the byte-identical body under the same key exactly once; only a definitive pre-acceptance rejection stays an ordinary failure, and an unresolved acknowledgement is reported as `unconfirmed` rather than as `send_failed`.
- Reconciliation cannot create a second card: it is byte-identical under the same idempotency key, so the provider replays the accepted message instead of accepting a new one, and an unkeyed send is never resubmitted.
- A personalized card carries an unambiguous current Murph line, or it is not sent.
- Contact-card and group-avatar generated captures must not collide, even when a provider retry reuses one RPC call id.
- No new persisted state, schema migration, queue, or delivery subsystem. The personalized path writes no reservation row at all.

- **Operation-scope Linq delivery-context producer (`readHostedAssistantInputLinqDeliveryContext`).** Previously admitted only events with `threadIsDirect === false` and hardcoded `threadIsDirect: false` on the returned context, so no direct route ever reached the group tool. It now admits either boolean and carries the event's real value. On its own this was necessary but not sufficient; the round-3 correction below completes it. Every existing group consumer (`resolveHostedGroupToolLinqRouteContext`, `resolveHostedGroupToolLinqSenderHandles`) already filters `threadIsDirect === false`, so group routing and sender-handle authority are unchanged. Proof: the new production-path test in `packages/assistant-runtime/test/hosted-runtime-workspace-assistant-phase.test.ts` persists a direct and a group input, runs the real `operationScope.runAutoReplyGroup`, and asserts the direct input binds the exact direct thread while the group input binds none; the existing 279-test file passes unchanged.
- **Direct vs group route ownership.** The personalized path never touches `hostedThreadRoute`; it carries `directLinqChatId` and is authorized against `hostedMemberRouting` through the existing runtime egress assertion. Group actions are unchanged. This is the PR's central trust-boundary decision — see Round 3 below.
- **`share_contact_card` access classification stays `owner_active`.** An interim head reclassified it to `participant_aware` on the reasoning that the personalized path is a direct-conversation member action. That was reverted: the classification map has no production consumer, so the change bought nothing, while the enforced owner-active check on the canonical branch is real. Web still runs `readHostedRuntimeGroupOwnerActiveAccess` whenever no `contactCardImageUrl` is present, and skips it only for the personalized variant. Proof: `apps/web/test/hosted-group-tool.test.ts`.
- **Group-tool port gains `directAttachmentRouteStatus`.** An optional trusted-host probe answered locally in the turn-context wrapper with no Web round trip, so route refusal costs nothing. Ports that are not route-aware omit it and fall back to the existing post-generation binding. `normalizeAssistantGroupTool` binds it alongside `request`; round 3 caught that the first version of this change dropped the probe at that boundary, silently restoring post-generation refusal. Proof: `packages/assistant-runtime/test/hosted-runtime-group-tool-linq-context.test.ts` for the wrapper and a normalization-boundary case in `packages/assistant-engine/test/assistant-codex-group-tool.test.ts`.
- **Shared contact-card share reservation (`reserve`/`release`).** Unchanged in shape and key derivation: canonical shares reserve exactly as before. The personalized path deliberately does not use it, because one personalized attempt contains an image generation bounded far above the 90-second reservation window. Proof: `apps/web/test/hosted-onboarding-linq-contact-card-share.test.ts` asserts the canonical reservation-keyed identity and that a personalized send writes no row.
- **Personalized provider idempotency key.** Derived from the trusted-host accepted-request id (`personalized-contact-card:<chatId>:<acceptedInputId>`) rather than a reservation instant, so the key is stable for the whole lifetime of one request. Canonical shares keep the reservation-instant key byte-for-byte.
- **`sendHostedLinqAttachmentMessage` now finishes its own acknowledgement (round 5).** A keyed final-message POST that fails ambiguously — timeout, transport loss, or a retryable status — is resubmitted once with the byte-identical body inside the same call. This also changes the *canonical* vCard share, which is the module's only other caller: a canonical send whose acknowledgement was lost now returns `sent` instead of `send_failed`. It cannot duplicate, because the resubmission carries the same idempotency key the first attempt used, and an unkeyed send is never resubmitted. Only the *reporting* differs by caller: an unresolved acknowledgement becomes `unconfirmed` for a per-request key and keeps the existing `send_failed` for a canonical, time-derived one. Proof: `apps/web/test/hosted-onboarding-linq-attachment-idempotency.test.ts` drives the real send owner over real HTTP against a loopback provider that models the local Linq support's exactly-once contract.
- **`share_contact_card` gains a fourth response status, `unconfirmed`.** Added to the Web response, the runtime response type, and the runtime parser. Only the personalized branch can produce it, and only a runner that sends `contactCardImageUrl` can reach that branch, so a currently deployed runner can never receive a status its parser would reject. See Deployment concerns. Proof: `packages/hosted-execution/test/parsers.test.ts` and the model-visible payload case in `packages/assistant-engine/test/assistant-codex-group-tool.test.ts`.
- **Linq roster self-handle selection.** The personalized path now requires exactly one *active* self handle, matching the existing native line-card path; the canonical vCard path keeps its prior first-`isMe` selection. Proof: stale-handle and multiple-active-handle cases in the same suite.
- **`fetchMurphHostedLinqContactCardVcfPhoto` abort composition.** The caller signal is now composed with the local 5s timeout via `AbortSignal.any` instead of replacing it, so a long-lived caller signal can no longer disable the fetch timeout. This also affects the canonical path when a caller passes a signal. Proof: `apps/web/test/hosted-onboarding-linq-contact-card.test.ts` aborts the local timeout and asserts the fetch resolves null while the caller signal stays active.
- **Route resolvers are separate owners.** `resolveHostedGroupToolLinqRouteContext` is unchanged from `main` — an interim version that parameterized it over `threadIsDirect` was reverted in round 3, because a direct route carries no thread-route authority and the two shapes genuinely differ. `resolveHostedDirectToolLinqRouteContext` is its own resolver over `{chatId, service}`.
- **Generated-image capture idempotency key.** A third scope, `contact-card-avatar`, joins `generate-image` and `group-avatar`, and the key's second component is now a caller-supplied request id rather than always the tool call id. The contact-card branch passes its accepted-request id so a replay reuses the capture instead of paying for a second stochastic generation; `generate-image` and `group-avatar` pass the tool call id exactly as before.
- **`share_contact_card` now has two exact, mutually exclusive request contracts.** Canonical carries optional group-thread authority and no personalized fields; personalized carries the private image URL, the accepted-input send key, and the exact direct chat id, and no group-thread authority. The runtime parser rejects mixed or partial shapes rather than letting one fall through to the other's authorization, `shareMurphHostedLinqContactCardVcfToChat` types `imageUrl` and `shareKey` as an all-or-nothing pair, and a personalized request that resolves no direct route is refused locally instead of being forwarded without a chat id. This removes the possibility of a half-formed personalized request rather than adding a guard against one. Proof: `packages/hosted-execution/test/generated-contact-card-parser.test.ts`, `apps/web/test/hosted-group-tool.test.ts`, `apps/web/test/hosted-onboarding-linq-contact-card-share.test.ts`.
- **Provider conflict recognition is an exact response shape, not a phrase match.** `isHostedLinqIdempotencyKeyReuseConflict` now requires a bounded JSON body whose single `error` key equals the provider's exact conflict message, so a proxy that wraps or prefixes that phrase can no longer be read as proof a card already landed. Proof: the proxy-wrapped 409 case in `apps/web/test/hosted-onboarding-linq-attachment-idempotency.test.ts`.
- **`buildOpenAiImageGenerationRequest` / `executeGenerateImageTool`.** Gain an optional `outputCompression` used only by the contact-card branch (JPEG, `output_compression: 40`). Omitted for every existing caller.

## Architecture and reuse

- Existing systems reused: the existing `group_tool.share_contact_card` action, the existing generate-image tool and OpenAI image pipeline, the existing group-avatar prepare/capture/private-publication path (`prepareGroupAvatarRuntimeRequest`), the existing private-media URL validator, the existing bounded-text parser helper, the existing first-party vCard builder, the existing durable contact-card share reservation table and blind-index mechanism, the existing Linq attachment delivery path, and the existing operation-scope Linq context producer. No new table, queue, service, owner, or delivery subsystem was introduced.
- New logic: a direct-conversation branch in the runtime group tool that requires fresh direct user input, refuses an ineligible route before any provider work, generates a compact JPEG under its own capture scope keyed to the accepted request, and passes the private image URL plus that host-owned request key to Web; a Web branch that validates the URL, skips the owner-active check only for the personalized variant, requires an unambiguous active line and a loaded photo before send, and keys the provider send on the accepted request; and one widened condition in the Linq delivery-context producer so a direct thread's route survives to the tool boundary.
- New abstractions: one small named production type, `HostedDirectToolLinqRouteContext` (`{chatId, service}`), the direct counterpart to the existing group route-context type; a direct route has no thread authority, so it cannot reuse the group shape. The round-2 retrospective deleted the other type this PR had introduced (`HostedLinqContactCardShareScope`) along with its lookup and release plumbing. What remains reuses existing contracts: `directAttachmentRouteStatus` is an optional method on the existing group-tool port rather than a new protocol action or port, `contactCardShareKey` is a bounded field on the existing request parsed by the existing bounded-text helper, and duplicate collapse reuses the provider idempotency key the send already took. No new tool, table, queue, or service boundary.
- Complexity intentionally avoided: no Murph profile-photo setting or persisted avatar state, no schema migration, no new tool or preflight protocol action, no Web round trip for route refusal, no background job or wake, no preview/confirm round trip, no second dedupe owner beside the provider idempotency key, no per-request reservation rows and therefore no row-retirement or cleanup machinery, and no model-facing controls that do not change what the member receives — `avatarAlt` and `avatarQuality` were both removed, leaving one fixed image configuration and a single `avatarPrompt` input.

## Hot reply path impact

Not applicable to ordinary replies. The PR adds no database call, network call, or awaited operation to durable acceptance, provider start, or durable reply handoff for any turn that does not explicitly call `group.share_contact_card` with `avatarPrompt`. Every added await sits inside assistant execution on that opt-in tool branch.

Added awaits on that branch, at maximum admitted cardinality (one card per tool call):

| Added await | Count | Serial/parallel | Timeout | Retry/fallback |
| --- | --- | --- | --- | --- |
| Direct-route eligibility probe | 1 | serial, in-process | none (no I/O) | refuses before any provider work |
| OpenAI image generation | 1 | serial | 240s | no in-path retry; failure returns a failed tool result |
| Durable capture + private publication (existing group-avatar path) | reused unchanged | serial | inherited | inherited |
| Web `share_contact_card` RPC | 1 | serial | inherited | none |
| vCard photo fetch (in Web) | 1 | parallel with the backup-phone lookup | 5s, composed with the caller signal | none; a missing photo fails closed with `photo_unavailable` |
| Reservation insert / conditional release (in Web) | canonical only: 1 + at most 1; personalized: 0 | serial | inherited | release is best-effort |

Before/after call counts on the reply path itself: unchanged (0 added).

The `group` dynamic tool's description and input schema changed, so this is measured rather than asserted.

- Method: rendered `MURPH_GROUP_TOOL` from the PR base (`a8aa91eca0`) and the current head with identical construction, and measured the exact provider-visible strings. Included fields: the tool description (visible in the deferred-tool listing) and the full serialized definition `{name, description, inputSchema}` (visible only after the model loads this deferred tool). Excluded: transport-only metadata, applied identically to both sides. Tokenizer: `o200k_base` (`gpt-tokenizer@3.4.0`), the vocabulary used by the target GPT-5-family model. Bytes are UTF-8 over the same included fields.
- Individual Murph and group Murph: **identical deltas**. The tool definition is shared and this PR introduced no runtime-conditional variation; availability for both runtimes is gated by the same unchanged `groupAvailable` predicate.

| Included field | Base tokens | Head tokens | Δ tokens | Δ % | Base bytes | Head bytes | Δ bytes |
| --- | --- | --- | --- | --- | --- | --- | --- |
| Tool description (always visible) | 146 | 140 | −6 | −4.11% | 786 | 729 | −57 |
| Full definition (after deferred load) | 2,725 | 2,823 | +98 | +3.60% | 11,722 | 12,146 | +424 |

Attribution: the description shrank because the rewrite traded two existing sentences for the new capability sentence, and is unchanged by the round-5 correction. The full-definition growth is entirely tool-schema growth from one new optional `share_contact_card` property, `avatarPrompt`, and its description. Removing `avatarAlt` and `avatarQuality` during review cut that growth from +141 to +63 tokens; the round-5 acknowledgement contract added the `unconfirmed` recovery sentence to that same property description, taking it to +98. That sentence sits on the property rather than the tool description on purpose, so the always-visible listing cost stays unchanged and only a turn that has already loaded this deferred tool pays for it. No authored prompt file, skill, Codex configuration, or model tool mode changed; no other tool definition changed.

Changed surfaces: `packages/assistant-engine/src/assistant-codex/dynamic-tool-catalog.ts` (`MURPH_GROUP_TOOL` description and `inputSchema`).

- Product experience: **applicable.** Intended outcome: an explicit direct request yields one saveable contact card with a generated photo, or an honest account of what happened. Direct journey evidence: the focused suites below exercise the real operation-scope route binding, authorization, required-photo composition, SMS rejection, ambiguous-route refusal, fresh-direct rejection, per-request dedupe, and the full acknowledgement contract over real HTTP through to the model-visible tool result. Gap: no live end-to-end iMessage send has been performed; that is the post-deploy smoke below.
- Prompt: **applicable.** The `group` tool description and schema changed; measured above.
- Frontend: **not applicable.** No `apps/web/app/**`, `apps/web/src/components/**`, or asset path changed; the diff is `apps/web/src/lib/**` and packages only.
- Coverage: **applicable.** Focused local proof is listed under Verification. Exact-head CI: pending on this push.

Reason: the diff changes an authority classification, a trust boundary (model-supplied image URL, route binding), an external delivery boundary, and idempotency/throttling keys.

ReviewGPT first-reviewed head: c86b0ead495b6a8fb91dcc1406e78c9c55ac0bd8

## Review remediation

Rounds 1–5 returned `FINDINGS`, `RETROSPECTIVE_REQUIRED`, `FINDINGS`, `FINDINGS`, and `RETROSPECTIVE_REQUIRED`; the preliminary specialist pass returned `FINDINGS`. Every accepted finding from all six stages is resolved in the current head. Round 4 independently re-verified the round-1 through round-3 corrections as effective. The only item still open is the S-3 end-to-end evidence gap, which needs a deploy.

| Finding | Disposition | Landed correction |
| --- | --- | --- |
| R1-1 High: production operation scope discards every direct Linq route, so the feature is unreachable | Accepted, confirmed against the code | `readHostedAssistantInputLinqDeliveryContext` admits either `threadIsDirect` boolean and carries the real value; new production-path test through `operationScope.runAutoReplyGroup` |
| R1-2 / S-2: unsupported routes refused only after generation | Accepted | `directAttachmentRouteStatus` probe on the group-tool port, checked before `prepareGroupAvatarRuntimeRequest`; tests assert zero provider, capture, and publication calls for SMS and missing/ambiguous routes |
| R1-3 / S-1: personalized throttle treats a fresh second request as a replay | Accepted | Reservation scope carries a host-owned `requestKey` derived from the accepted input id; tests assert same-request collapse and distinct-request send at +45s |
| R1-4 / S-4: remove model-facing controls that do not affect the card | Accepted | `avatarAlt` and `avatarQuality` removed from the catalog schema and Zod parser; one fixed internal configuration; provider-input growth cut from +141 to +63 tokens |
| S-5: local 5s photo timeout not behaviorally tested | Accepted | Applied the returned coverage patch (tests only), which aborts the injected local timeout and asserts a null result with the caller signal still active |
| S-6: fresh-input authority clause has no direct proof | Accepted | Applied the returned coverage patch (tests only): direct scope with empty `acceptedInputIds` makes zero provider and host calls |
| S-3: no production-faithful end-to-end iMessage proof | Accepted as an evidence gap, not resolvable pre-deploy | Unchanged: a live send requires the deployed runner and Linq line. Recorded as the post-deploy smoke below rather than claimed as proven |
| R1 body discrepancies (route binding, "asking again", "no new type") | Accepted | Corrected; the first two are now true of the code, and the type introduced at that head has since been deleted |
| R2-A: a same-request replay can resend once its generation ages the reservation past 90s | Accepted | Personalized sends key the provider idempotency on the accepted-request id and take no reservation; test proves an identical key at +5 minutes |
| R2-B: one durable reservation row per distinct request key, with no retirement path | Accepted | Per-request rows removed entirely rather than given cleanup machinery; test asserts zero rows for personalized sends |
| R2-C: the vCard can carry a stale or ambiguous Murph line | Accepted | Personalized sends require exactly one active self handle, matching the native line-card path; inactive-only and multi-active rosters refuse |
| R5-1: an ambiguous final Linq response is terminalized as `send_failed`, so a delivered card can be reported as a failure | Accepted; retrospective recorded first (below) | The send owner reconciles the byte-identical body under the same key once, inside the same call; an unresolved acknowledgement becomes `unconfirmed` for a per-request key, and the model is told to ask before making another card |
| R5 body discrepancies (test subtotal, `resolveHostedToolLinqRouteContext`, pre-generation refusal, `already_shared` as success) | Accepted | All four corrected in this body before round 5 closed; the arithmetic, the two-resolver split, the "refused before generation *and* revalidated at the send boundary" wording, and the explicit `already_shared` success clause now match the code |

### Round 2 retrospective (required before round 3)

Round 2 returned `ROUND_OUTCOME: RETROSPECTIVE_REQUIRED` because the round-1 remediation
reused the same deduplication mechanism that round 1 had already faulted. It was right to
stop us.

**Direction, first-reviewed head vs now.** At `c86b0ead49` the personalized path generated
an image and then consulted a chat-scoped 90-second reservation, so distinct requests
collapsed. The round-1 remediation kept that reservation and made its *identity* per
request. That fixed the reported symptom and left the mechanism wrong: one personalized
attempt contains an image generation bounded at 240s, so the 90s reservation can expire
mid-attempt, mint a new `attemptedAt`, derive a new provider idempotency key, and send a
second card for the same request. Direction changed from "make the throttle's identity
correct" to "the throttle is the wrong owner for this effect".

**Concepts and owners, added and removed.** Removed: the `HostedLinqContactCardShareScope`
type, its scope parameter on reserve/release, its lookup-key branch, the personalized
release plumbing, and per-request reservation rows (the durable-growth surface round 2
flagged). Retained and necessary: the direct-route producer correction, the pre-generation
route refusal, the `avatarAlt`/`avatarQuality` deletion, and the tests-only coverage
corrections. Added: nothing new — duplicate collapse moved onto the provider idempotency
key the send already accepted, and the capture key's second component became a
caller-supplied request id rather than always the tool call id.

**Decision: shrink.** Not continuation, and not a redesign. The personalized send now has
exactly one owner for its whole lifetime — the accepted-request id — and the local
reservation is simply not used for it. Net effect on authored source is −6 lines across
the two files the retrospective named, while deleting a type, a durable-row growth path,
and a second dedupe owner. Answering round 2's five questions in order: (1) the
trusted-host accepted-request id owns the irreversible send; (2) no, personalized sending
needs no local reservation owner, because the provider primitive already accepts a stable
key and nothing else required a second owner; (3) a replay reuses the already-captured
image, via the same accepted-request id, with no queue, scheduler, or reconciliation loop;
(4) request-scoped rows are not necessary and were removed rather than given cleanup
machinery; (5) the current line is proven by requiring exactly one *active* self handle,
matching the existing native line-card path, and failing closed otherwise.

Coverage patch artifact disposition: one `reviewgpt-coverage.patch` was downloaded from the owned specialist thread, read in full, verified to touch only two test files, passed `git apply --check`, and was applied deliberately. No production, config, schema, workflow, or docs hunks were present.

| R3-2: `normalizeAssistantGroupTool` dropped `directAttachmentRouteStatus`, so the pre-generation refusal never reached production | Accepted, confirmed against the code | The probe is bound alongside `request`; new test crosses the normalization boundary instead of hand-building the tool |
| R6-1: the runner's 30s web-control hop can discard the terminal truth while Web finishes the send, so a delivered card is reported as a generic tool failure | Accepted, confirmed against the timeout constants and call order | The group-tool port, which owns that hop and already varies behavior by action, returns `unconfirmed` when a personalized card request gets no answer at all; an answer from Web, error status included, stays a failure |
| R6 body discrepancies (canonical "behaviorally unchanged", "whole flow completes inside the tool call") | Accepted | Both corrected above: the canonical path's recovered-acknowledgement change and the hop's deadline relationship are now stated plainly |

### Round 4: a replayed send now reports truthfully

Round 4 verified R1-1, R1-2, R1-4, S-4, S-5, S-6, R2-B, R2-C, R3-1 and R3-2 as effective, and found that R2-A's correction was incomplete. The stable idempotency key was necessary but not sufficient: `sendHostedLinqAttachmentMessage` creates a new provider attachment on every attempt, so a replay submits a *different body* under the same key and the provider answers with a key-reuse conflict. The card was already delivered, but the turn reported failure and could never close the loop truthfully.

Resolution, with no new ledger, reservation, queue, or reconciliation owner: the Linq client classifies that one exact same-key/different-payload response, and only a send carrying a per-request key converts it to `already_shared`. A canonical share's key is time-derived and proves nothing about intent, so it stays a failure — as do generic 409s, transport failures, timeouts, and every other provider error. Proof: `apps/web/test/hosted-onboarding-linq-attachment-idempotency.test.ts` drives the real `sendHostedLinqAttachmentMessage` against the local Linq stub's exact conflict body and asserts the unrelated-409 and server-error cases stay unclassified; the share suite asserts the conversion happens only with a request key.

### Round 3 (HIGH): direct sends now use the direct route owner

Round 3 proved the feature was *still* undeliverable, for a deeper reason than round 1 found. Direct home conversations are owned by `hostedMemberRouting`; group threads by `hostedThreadRoute`; and `assertHostedLinqChatNotOwnedByThreadRouteTx` forbids one chat from existing in both. This PR was routing a direct-conversation effect through the *group* thread-route authority chain, which such a chat can never satisfy:

- an ordinary active-member direct inbound builds its mailbox wake with no `routeAuthority`, so `externalThreadRouteAuthorityPresent` persists as `false` and the direct delivery context was still dropped; and
- supplying a synthetic authority would not have helped, because Web's `authorizeHostedRuntimeGroupLinqThread` validates the thread-route store that the chat is prohibited from joining.

Resolution — each owner stays in charge of its own routes, with no new store or durable record:

- The delivery-context producer admits a direct Linq event on the trusted reply target alone and returns `routeAuthority: null` for it. Group resolvers require an authority *and* `threadIsDirect === false`, so they remain structurally unable to select a direct route.
- `resolveHostedDirectToolLinqRouteContext` is now its own resolver over `{chatId, service}` rather than a parameterization of the group resolver, since the two shapes genuinely differ. The group resolver is restored to its original form.
- The request carries `directLinqChatId` — the trusted host's exact chat — instead of a fabricated `linqThread` authority.
- Web authorizes that chat with the existing `assertHostedLinqRecentInboundEngagementForRuntime` (`authorityCheckOnly: true`) and requires `threadIsDirect === true`. Group actions keep the thread-route assertion unchanged.

Proof: the production-path test now persists an *ordinary* direct wake with `externalThreadRouteAuthorityPresent: false` — the shape the webhook actually writes — and asserts the group tool receives `directLinqChatId` for the direct input and nothing for the group input. Web-side tests assert the direct owner is consulted, the thread-route assertion is *not*, and that both a rejecting owner and a non-direct result refuse before any send.

### Branch history note

An automated workflow on this branch (`agent-contact-card-final-boundary.yml`, then
`agent-contact-card-boundaries-f254-v2.yml`) ran on every push with `contents: write` and
force-pushed a `git reset --hard` to a pinned SHA plus one generated commit. It rewrote the
branch at `77eb4a82d5`, which landed a generated "enforce personalized request boundaries" patch,
dropped the round-6 correction, and replaced this PR body with one that omitted the required
sections and reset the immutable `ReviewGPT first-reviewed head` line. The workflow removed itself
in that same rewrite, so nothing on the branch triggers it now.

The round-6 correction is restored at `6c94df537f` and this body is restored from the round-6
review package, with the first-reviewed head back to its original immutable value
(`c86b0ead49`). The generated patch's product changes were kept and are disclosed under
Non-obvious affected surfaces; the generator scripts are recoverable in history at `93080775ff`.

### Round 5 retrospective (required before round 6)

Round 5 returned `ROUND_OUTCOME: RETROSPECTIVE_REQUIRED` because the round-4 correction and the
new finding are the *same* mechanism: the boundary between an irreversible Linq send and a
truthful terminal result. Round 4 fixed the second half of that boundary (a replay's
same-key/different-body conflict). Round 5 pointed at the first half, which round 4 had not
touched: the original submission's own ambiguous response — a timeout, a lost connection, or a
post-accept 503 — was still terminalized as `send_failed`, so a card that had already arrived
could be reported as a failure. Two consecutive rounds narrowing the same boundary one response
code at a time is exactly what the retrospective rule exists to stop, and the reviewer was right
to require the whole acknowledgement contract instead of a third classification.

**Restated requirement.** A member asks in a direct iMessage conversation for a new contact
photo and gets one saveable card, or an honest account of what happened. "Honest" is the part
this round is about: the interaction must not close by telling the member something failed when
it may have succeeded, and must not turn that uncertainty into a second generated card.

**First reviewed shape vs now.** At `c86b0ead49` the send was one POST whose only non-failure
outcomes were `sent` and a reservation-derived `already_shared`; every provider error, ambiguous
or not, became `send_failed`. Now there is one stated contract with three terminal truths —
confirmed sent, confirmed already sent, explicitly unconfirmed — and `send_failed` is reserved
for a definitive pre-acceptance rejection. Authored source is 695 added / 74 deleted against the
base, 769 lines of churn, below the 2,000-line trigger; the growth is dominated by the round-3
authority correction rather than by this boundary.

**The three requirement-level decisions the reviewer asked for.**

1. *The terminal truth contract.* Three states, named above. An accepted request may not become
   an ordinary `send_failed` while provider acceptance is still possible.
2. *Reconcile, rather than expose an unresolved state and hope.* The existing send owner
   resubmits the exact original final-message body under the same accepted-request key, once,
   before returning. This is not a new mechanism: the provider already owns exactly-once for that
   key, and the repository's own local Linq support proves that a byte-identical resubmission
   recovers the original message identity without accepting a duplicate. So the overwhelming
   majority of ambiguous responses resolve to a definite answer inside the invocation. The
   member-facing unconfirmed state is the residual — what is left when the provider will not
   answer twice — not the primary strategy. No durable owner, queue, reservation, or
   reconciliation lifecycle was added; the whole contract lives inside the one call that was
   already there.
3. *A later retry against an unconfirmed request.* In-band, not durable. An `unconfirmed` result
   tells the model to say the card may already have arrived and to ask the member to look before
   making another one. A durable "pending request" record would be a new state owner for a
   condition that is rare, self-resolving (the member can see the conversation), and reversible;
   the architecture pressure check rejects it in favor of the interaction the member is already
   having.

**Decision: continue, with the boundary closed rather than narrowed again.** Not a redesign and
not a split. The test for whether this is genuinely the last correction of this mechanism is that
the contract is now stated over *response classes* — resolved, proven-already-delivered,
unresolved — rather than over individual status codes, and the six cases the reviewer named are
each proven at the send owner over real HTTP. A future provider response that is none of these
falls into "unresolved" by construction and is reported honestly without another code change.
Cost: +126 authored source lines, one new response status, and one sentence of tool-description
text. Nothing was deleted by this round, which is why the decision is "continue" rather than the
round-2 "shrink".

### Round 6: the terminal truth now survives the transport

Round 6 verified every prior accepted finding as still effective and found that R5-1 was
correct at the send owner but not at the wired boundary. The runner forwards a group-tool call
to Web with a 30-second deadline, while Web's personalized sequence has a much larger worst-case
sum of bounded provider waits and does not receive the runner's abort. So the runner can give up
while Web goes on to accept the card, and the turn saw only `group tool request failed` — the
same "reported failed while delivered" shape round 5 set out to remove, one layer up.

Resolution, at the layer that owns the hop rather than by adding one: `createHostedRuntimeGroupToolPort`
already varies behavior by action and already has the transport's own pre-response failure
diagnostics. A personalized contact-card request that gets no answer at all now returns the
`unconfirmed` result the model already knows how to explain. A response from Web, including an
error status, is a real answer and stays a failure — the same definitive-versus-ambiguous rule
the send owner uses.

Two parts of the reviewer's proposed cure were deliberately not taken, and this is the load-bearing
judgment of the round. It suggested propagating the runner's abort into Web and giving the send an
inner end-to-end deadline. Both convert "the card probably arrived" into "the card was abandoned
mid-send" for an irreversible effect the member explicitly asked for. What that first attempt left
unclosed was the 30s-versus-worst-case gap itself.

**Superseded by round 7.** That transport branch, its helper, and its three port tests are all
deleted. Round 7 showed the hop can expire while Web is still doing pre-send work, so the branch
made `unconfirmed` mean two different things and could invite a second card. The gap it declined
to close is now closed the other way: the operation is bounded to finish inside the hop, so the
transport never has to invent a result. The section below is the current design.

### Round 7 retrospective (loop cap reached)

Round 7 returned `ROUND_OUTCOME: RETROSPECTIVE_REQUIRED` and it was right. The round-6
remediation let the *transport* say `unconfirmed`, and that hop can expire while Web is still
doing pre-send work. So one word covered two materially different states: the send owner tried
twice and could not learn the answer, and Web has not attempted a send at all. In the second
case the model tells the member the card may already have arrived and invites another if it is
not there — while the first invocation is still free to send. Two cards, from the mechanism
introduced to prevent exactly that.

**Restated requirement.** One explicit direct request produces one saveable card and one honest
terminal result. "One" is the part round 7 is about.

**Direction, first reviewed head vs now.** At `c86b0ead49` the send was one POST whose only
non-failure outcomes were `sent` and a reservation-derived `already_shared`. Now there is one
terminal contract with three truths, produced by exactly one owner, inside a window the calling
turn is guaranteed to still be waiting on.

**Decision: shrink.** Not another status and not a pending contract with a new owner. The
round-6 transport branch is reverted, and the personalized operation is instead sized to finish
inside the hop that calls it: an 8-second deadline on the phases that provably precede the
message POST, with the POST and its reconciliation keeping their full budgets. That is one
constant and one phase-scoped signal, against a deleted branch, its helper, and its three tests.
`unconfirmed` regains the single meaning the round-5 retrospective gave it, which is why the two
body claims round 7 flagged are now true rather than reworded.

**What this deliberately does not do.** It does not propagate the runner's abort into Web, and it
does not bound the send itself. Both would convert "the card probably arrived" into "the card was
abandoned mid-send" for an irreversible effect the member asked for.

**Cap status.** This was substantive round 7, the loop's hard cap, and the accepted finding is
fixed rather than left in place. Per `agent-docs/operations/pr-reviewgpt-loop.md` the loop is now
paused: an eighth round needs an explicit continuation decision, and the options on the table are
continue, shrink further, split, or hold. My recommendation is one more round, because the three
rounds since the last `PASS` have each been a single narrowing of one boundary that is now closed
by deletion rather than by addition, and because the current head also carries generated content
that only round 7 has seen.

## Change-shape breakdown

Classification rule: `**/src/**` and other non-test production files are source; `**/test/**` is tests/fixtures; `agent-docs/**` is docs; build/CI/lockfiles are config/tooling; committed build output is generated/other. Counts are `git diff --numstat origin/main...HEAD`. No binary files.

Immutable first-reviewed-head baseline (`c86b0ead49`):

| --- | --- | --- |
| Source | 317 | 45 |
| Tests/fixtures | 683 | 3 |
| Docs | 36 | 0 |
| Config/tooling | 0 | 0 |
| Generated/other | 0 | 0 |
| **Total** | **1036** | **48** |

Current head:

| --- | --- | --- |
| Source | 825 | 101 |
| Tests/fixtures | 1924 | 3 |
| Docs | 39 | 0 |
| Config/tooling | 7 | 1 |
| Generated/other | 0 | 0 |
| **Total** | **2795** | **105** |

Authored-source growth caused by review remediation: +508 added and +56 deleted source lines against the first-reviewed baseline; 926 lines of authored-source churn in total, below the 2,000-line retrospective trigger. Round 7 moved authored source *down*: its shrink deleted more than it added. Config/tooling is the runner bundle's total ceiling, ratcheted to the size this feature's lazy output actually measures. The largest single share is the round-3 authority correction, which is the change that makes the feature deliverable at all: the route-producer widening, the eligibility probe and its port declarations, the stable send identity, and the active-line check, net of the schema and reservation-scope lines deleted by the two complexity collapses. The round-2 retrospective alone moved authored source by −6 lines while removing a named type, a durable-row growth path, and a second dedupe owner. The round-5 acknowledgement contract added +126 and deleted 8, all inside the existing send owner and its two callers; round 6 added ~35 inside the group-tool port and round 7 deleted all of it again, replacing it with one constant and one phase-scoped signal in the owner that was already there. The remaining movement is the exact-request-shape work described under Non-obvious affected surfaces, which is net-neutral to negative in concepts: it deletes the possibility of a partial personalized request rather than adding a guard for it. Test growth is +1,129 lines, much larger than the source delta because each accepted finding required its own fail-closed proof: one production-path test that traverses persisted input rather than injecting a synthetic context, and one that drives the real send owner over real HTTP through all six acknowledgement cases.

## Design proof

Not applicable. No user-facing `apps/web` UI path changed (`apps/web/app/**`, `apps/web/src/components/**`, `apps/web/app/globals.css`, and `apps/web/public/**` assets are all untouched), so the design-page catalog and screenshot requirements do not trigger.

## Verification

All of the following were run on the exact current head, `e9485836aa`.

- `pnpm typecheck:packages` — pass
- `pnpm --dir apps/cloudflare typecheck` — pass
- `pnpm provider-requests:guard` — pass
- `pnpm --dir apps/web exec eslint` on every changed `apps/web` file — pass, 0 warnings
- `apps/web`: `hosted-group-tool.test.ts`, `hosted-onboarding-linq-contact-card-share.test.ts`, `hosted-onboarding-linq-contact-card.test.ts`, `hosted-onboarding-linq-attachment-idempotency.test.ts` — 210 passed
- `apps/cloudflare`: `group-tool-port.test.ts`, `group-tool-port-replay.test.ts`, `runner-bundle-entrypoint-bundle.test.ts` — 54 passed
- `packages/assistant-runtime`: `hosted-runtime-group-tool-linq-context.test.ts`, `hosted-runtime-workspace-assistant-phase.test.ts` — 313 passed
- `packages/hosted-execution`: `parsers.test.ts`, `generated-contact-card-parser.test.ts` — 72 passed
- `packages/assistant-engine`: `assistant-codex-group-tool.test.ts`, `openai-image-generation.test.ts` — 89 passed
- Total: 738 focused tests passing (210 + 54 + 313 + 72 + 89).
- **Exact-head CI is fully green at `e9485836aa`: 16 required checks pass, 1 expected skip, 0 failures.**

The acknowledgement contract is proven at the send owner over real HTTP, not through a mock of our own code. `apps/web/test/hosted-onboarding-linq-attachment-idempotency.test.ts` starts a loopback provider that models the exactly-once contract of the repository's local Linq support (`apps/cloudflare/test/helpers/hosted-local-linq-support.ts`) — one accepted message per idempotency key, byte-identical resubmission replays the original message identity, a changed body under that key conflicts, and the post-accept controls record acceptance before failing — and drives the real `sendHostedLinqAttachmentMessage` through every case:

| Case | Terminal result | Cards accepted |
| --- | --- | --- |
| Post-accept 503, then reconciliation succeeds | `sent`, original message id | 1 |
| Connection destroyed after acceptance | `sent`, original message id | 1 |
| Send times out after acceptance | `sent`, original message id | 1 |
| Post-accept 503 on both attempts | `unconfirmed` | 1 |
| Later request reuses the key with a new attachment | key-reuse conflict → `already_shared` | 1 |
| Definitive 400 pre-acceptance | ordinary failure, no resubmission | 0 |
| Proxy-wrapped or unrelated 409 | ordinary failure, no resubmission | 0 |
| Pre-send deadline expires | prepare-phase failure, never `unconfirmed` | 0 |
| Accepted 200 whose body stalls | `sent`, message identity lost | 1 |
| 409 whose body stalls | reconciled, then `sent` | 1 |
| Unkeyed send, acknowledgement lost | ordinary failure, never resubmitted | 1 |

The pre-send row is the round-7 correction: the deadline only bounds phases that provably precede the message POST, so expiring under it can never be mistaken for provider uncertainty. The two stalled-body rows are the round-8 correction: `fetchLinqApi` stops its own timer once headers arrive, so each attempt now tracks an absolute deadline and reads its body against what is left of it. A body that does not finish reads as absent, never as a different answer — on an accepted 200 that costs only the message identity, and on a 409 it enters reconciliation instead of being silently downgraded to "not a conflict". There is no transport-owned `unconfirmed` any more. The remaining layers are proven at their own boundaries — `hosted-group-tool.test.ts` also proves the deadline covers slow direct authorization and refuses before any provider call — `hosted-onboarding-linq-contact-card-share.test.ts` for reporting `unconfirmed` only per request and never freeing a reservation on it, `hosted-group-tool.test.ts` for the Web response, `packages/hosted-execution/test/parsers.test.ts` for the runtime parser, and `assistant-codex-group-tool.test.ts` for the model receiving a *successful* tool result carrying `status: "unconfirmed"` rather than a failed call it cannot explain. Not proven locally: a single test traversing runner → proxy → the real Web route → Linq in one process; that lane is the hosted-local stack, and the live path is covered by the post-deploy smoke below.

## Deployment concerns

Deploy **Web first, then the Cloudflare Worker/runner bundle.**

- Skew behavior, correct order: new Web accepts both canonical requests (no `contactCardImageUrl`) and personalized ones, so it is compatible with the old runner. Once the new runner ships, it can start sending `contactCardImageUrl` and `contactCardShareKey`.
- Skew behavior, wrong order: a new runner against old Web would send fields that the parser rejects as unknown keys, so the tool call fails closed. No data is corrupted and no wrong card is sent, but the member's request fails until Web catches up.
- Worker/container skew: during gradual rollout a new Worker version can serve traffic while warm `RunnerContainer` processes still run the previous runner bundle. Both bundles are compatible with new Web, so this window is safe in the correct order.
- Reservation keys: the canonical reservation key derivation is unchanged from the current release, so in-flight canonical reservations keep matching. Personalized sends no longer write or read reservations; during the skew window an older runner that omits `contactCardShareKey` still falls back to the reservation-keyed path, which is the previous behavior.
- New `unconfirmed` response status, correct order: this is the one place the deploy order could have been wrong, and it is not. New Web can return `share_contact_card` `status: "unconfirmed"`, which a currently deployed runner's parser would reject as an unknown status. It is unreachable for that runner: `unconfirmed` is produced only when the share carries a per-request key, and only a request carrying `contactCardImageUrl` supplies one, which only the new runner sends. During the Web-first window the old runner keeps receiving exactly the three statuses it already parses.
- Canonical send behavior during the window: a canonical vCard share whose acknowledgement is lost now resolves to `sent` instead of `send_failed`, because the send owner resubmits the identical body under the key that send already used. The response shape is unchanged, so an old runner reads it exactly as before; the reconciliation cannot duplicate a card, and an unkeyed send is never resubmitted.
- No schema migration, no backfill, no new environment variable.
- Post-deploy check: make one explicit generated contact-card request in a direct iMessage thread and confirm the `.vcf` arrives with the generated photo and can be saved over the existing Murph contact; confirm a canonical group contact-card share still works and is still owner-only. This is the S-3 evidence gap above.

---

This PR remains draft and unmerged.


rocketman-21:
**Removed a self-pushing workflow from this branch.**

`.github/workflows/agent-contact-card-final-boundary.yml` (added at a9d920121d) triggered on every push to this branch with `contents: write`. Its finalize step did `git reset dd5ea281e7`, committed a generated patch on top, and force-pushed the result back to the branch. `dd5ea281e7` predates both the round-5 acknowledgement correction and the current `main` merge, so a successful run would have silently dropped them from the PR. Its last run failed; the generator scripts under `.agent-patch/` are string-replacement patchers against source that has since changed.

Both are branch scaffolding rather than product and would have merged into `main` as-is, so they are removed at 83e070096b. The content is recoverable from history at a9d920121d if any of it is still wanted.

One substantive idea in that patch was reviewed and deliberately not adopted: making `workspace-assistant-phase.ts` return `direct_attachment_route_unavailable` locally instead of forwarding a personalized `share_contact_card` that resolved no direct route. That path is already unreachable through this port — `directAttachmentRouteStatus()` refuses first, using the same resolver over the same turn-fixed contexts — and if it were reached, Web refuses a personalized request carrying neither `directLinqChatId` nor an authorized `linqThread`. It fails closed either way, so it is a defensive tidy-up rather than a defect fix.

rocketman-21:
## Final audit complete

Final audited head: `d726c9cec9320d9519f69202996064ee3a3ea642`

I completed three review passes and pushed the fixes rather than leaving findings as notes.

Fixed in the final pass:

- exact canonical versus personalized request shapes;
- refusal of missing, ambiguous, and SMS direct routes before generation;
- prevention of direct requests falling through to group authorization;
- conservative owner-only classification for canonical sharing;
- required image/send-identity pairing at each boundary;
- stable per-request duplicate handling for personalized cards while preserving canonical reservation behavior;
- honest `sent`, `already_shared`, and `unconfirmed` delivery outcomes;
- unambiguous active Murph-line selection;
- operation-scope regression coverage;
- stale implementation comments and execution-plan wording.

Verification passed twice on the functional candidate: focused package and Web suites, `pnpm typecheck:packages`, `pnpm --dir apps/web typecheck:prepared`, `git diff --check`, exact changed-path checks, and clean-working-tree checks. The final comment-only alignment also passed its exact-path and clean-tree workflow. Vercel is green on the final head.

I deliberately did not add a persistent avatar setting, schema, queue, preview flow, second duplicate store, or a new generic response utility; those would add more architecture than this one-shot contact-card experience needs.

There are no unresolved review threads or temporary audit files in the PR diff. The PR remains draft, open, and unmerged.

PR #1488, opened 2026-08-09
fix(contact-card): bound the personalized send to the turn waiting on it

rocketman-21:
## Why this PR exists

PR #1458 shipped the generated Murph contact card. Its final review round found two ways the send could outlive the window its caller was waiting on, and the fix landed on the branch after the merge, so it is not in `main`. This PR carries that correction forward, with one further change to how an unread answer is reported.

Both defects have the same shape: the turn stops waiting, the Web invocation keeps going, and the member is told something the send owner never established. That is the exact failure #1458 spent its last three rounds closing, so leaving it out of `main` would ship the feature with the hole still open.

Unchanged from #1458: a member asks Murph in a direct iMessage conversation for a new contact photo and gets one saveable `.vcf` back. What changes is honesty at the edges. A card that arrived is never reported as failed, a card that may have arrived is never reported as delivered, and the member is never invited to ask again while the first request can still send.

## User experience

- **Irreducible purpose.** Unchanged. One request, one saveable attachment, one honest result.
- **Entry point.** Unchanged: a fresh member message in a personal iMessage conversation.
- **Immediate feedback.** Unchanged. No new surface.
- **Timing class.** Slow-tool class, unchanged. What is new is that the personalized operation is bounded to finish inside the runner's 30-second web-control hop, and the bound is anchored where that hop actually reaches Web: the route stamps the arrival time as its first statement, before the callback body read, signature verification, and nonce consumption. From that anchor the handler gets a 20-second budget, leaving the remaining third of the hop for transit, cold start, and the response return. Within it, a 6-second pre-send deadline covers direct-route authorization, roster read, photo fetch, backup-number read, and attachment create and upload including their response bodies; the message POST and its one reconciliation then each take up to 7 seconds covering headers *and* body. The pre-send deadline is derived as the operation deadline minus both attempt budgets, so reaching it *is* the admission check for the irreversible POST — and it is enforced by comparing the clock immediately before that POST, not by trusting an abort callback. `AbortSignal.timeout` schedules work; a busy event loop can run it after its own deadline, by which time the request has already gone out. For the same reason an attempt with no remaining budget throws instead of dispatching, since a zero-millisecond signal still lets `fetch` start first.
- **Asynchronous continuation owner.** None, with the boundary stated precisely: the *irreversible message POST* is refused by a synchronous clock comparison, so no card can be admitted after the deadline. Earlier provider work in the pre-send stretch — the roster read, photo fetch, attachment create and upload — is bounded by the same deadline but through an abort signal, so a delayed event loop can let one of those start or finish slightly late; none of them delivers anything to the member. Every response is owned: contents are read through the bounded reader, which joins its own cancellation before settling, and status-only responses are cancelled and awaited. A read can likewise finish marginally past the nominal deadline if the cancel callback is delayed, but it cannot outlive it indefinitely. What the deadline does *not* govern is database latency: callback nonce consumption, direct-route authorization, and the backup-number read are all Prisma work that takes no abort signal, so a hung database can delay the route's settlement past the budget regardless. The nonce read runs before the handler is even entered, which is precisely why the anchor is taken at the top of the route — its cost is charged even though it cannot be cut short. Those are checked against the deadline before the irreversible POST, so a slow database can never produce a late card; it can only produce a late refusal. That limit is a pre-existing property of this route, not something this PR introduces, and closing it would mean abandoning queries that keep running anyway — the same false bound this PR just removed.
- **Final destination and audience.** Unchanged: the same direct conversation, visible only to that member.
- **Failure and recovery.** A personalized request still ends in exactly one of three truths — `sent`, `already_shared`, or `unconfirmed` — and `unconfirmed` is still producible only by the send owner, only after it has resubmitted the identical body under the same request key. Expiring anywhere in the pre-send stretch is provably before the irreversible message POST — that stretch does include provider roster, photo, and attachment work, none of which reaches the member — so it is an ordinary refusal (`contact_card_presend_deadline_exceeded`), never uncertainty.
- **Iterating.** Unchanged. After an `unconfirmed` result the model asks the member to look before making another one, and that is safe to act on because no first invocation is still running.
- **What happens next.** Nothing further is required of the member; iOS owns whether the contact is saved.
- **Rendered evidence.** Not applicable — no `apps/web` UI surface changed.

- One member request produces at most one card. Every resubmission this PR adds is byte-identical under the same idempotency key, so the provider replays the accepted message instead of accepting a second one.
- An unkeyed send is never resubmitted. It has no key to make that safe.
- The turn always receives the send owner's own terminal result; no other layer invents one.
- An answer the send owner could not finish reading is never reported as a different answer.
- Expiry before provider entry is an ordinary refusal; only post-submission ambiguity can become `unconfirmed`.
- Canonical group contact-card sharing keeps its reservation-derived send identity, its response shape, and its owner-only check.
- No new persisted state, schema migration, queue, owner, or delivery subsystem.

- **The web-control route now owns the operation's start time.** `apps/web/app/api/internal/hosted-execution/groups/tool/route.ts` stamps `requestStartedAtMs` as its first statement and threads it through `handleHostedRuntimeGroupTool` into the personalized share branch, which derives both deadlines from it. Production always supplies it; the handler-local `Date.now()` fallback exists only for direct unit callers. Authorization and the backup-number read consume no abort signal, so both are checked against the deadline explicitly instead of being assumed fast. Proof: `apps/web/test/hosted-group-tool.test.ts` drives the real route with signed-callback verification delayed past the pre-send deadline and asserts the request is refused with zero provider admission and that the anchor predates the delay; `apps/web/test/hosted-group-tool-route.test.ts` asserts the route passes the anchor at all.
- **Every response on this path has exactly one terminal owner.** `fetchLinqApi` clears its timer once headers arrive, so any response nobody reads has nothing left to end it. Bodies whose contents are needed — an accepted send, a 409 — go through the bounded reader, which joins its own cancellation before settling. Bodies where only the status matters — a failed attachment create, the upload response, any non-409 send or reconciliation response — are cancelled and awaited explicitly. Proof: four loopback cases stall exactly those branches and assert the provider's sockets end with no further client action; each fails against the previous implementation.
- **`fetchLinqApi` stops its own timer when response headers arrive.** Reading the body was therefore unbounded, so a stalled body could keep the send owner blocked well past the budget its caller was promised. This applied to both ends of the operation: each attachment-send attempt now tracks an absolute deadline and reads its body against what is left of it, and the attachment-*create* body is read against the pre-send deadline that was already meant to cover it. Each attempt also takes the lower of its own budget and whatever the caller's deadline still allows, so the send and its one reconciliation cannot together outlive one operation. This is a real behavior change for the attachment-send path only; other Linq callers are untouched.
- **A body that does not finish reading is now its own outcome, not a downgraded one.** Previously an unreadable 409 body silently became "not a conflict" (an ordinary failure) and an unreadable 200 body became a terminal `sent`. Both now enter the same-key reconciliation that already yields `sent`, `already_shared`, or `unconfirmed`. This is the change beyond what landed on the #1458 branch: reporting delivery from a status line alone claimed something the send owner never established.
- **Unkeyed sends keep the previous behavior deliberately.** With no key, a resubmission would accept a second message, so an accepted 200 with an unreadable body stays `sent` and loses only the message identity, which no caller of this send uses.
- **Prepare-phase failure tagging widened.** Everything inside the attachment prepare closure provably precedes the message POST, so aborts and transport errors there are now tagged as prepare failures too, not only errors that were already hosted-onboarding errors. That is what lets a pre-send deadline expiry be read as "nothing reached the chat".
- **The 7-second attempt budget has one owner.** `HOSTED_LINQ_ATTACHMENT_SEND_ATTEMPT_TIMEOUT_MS` lives in the Linq client, which is the layer that spends it; the share owner derives its pre-send deadline from that value rather than restating the number, so the two cannot drift apart.
- **New unavailable reason `contact_card_presend_deadline_exceeded`.** Produced only on the personalized branch. The runtime parser already accepts any `unavailableReason` string, so this needs no contract change and no runner deploy to be understood.

## Architecture and reuse

- Existing systems reused: the existing send owner and its reconciliation path, the existing prepare-phase concept and its failure predicate, the existing `unconfirmed` outcome and its model-facing guidance, and the existing per-request idempotency key. No new table, queue, owner, status, or protocol field.
- New logic: one route-owned arrival timestamp threaded to the share branch, one pure resolver deriving both deadlines from it, two explicit deadline checks at the steps that cannot consume a signal, one bounded response-body reader used at three response sites, per-attempt budgets clamped to the caller's remaining time, and a three-way conflict-read result so "unreadable" stops masquerading as "not a conflict".
- New abstractions: none. The named additions are two exported constants (`HOSTED_LINQ_PERSONALIZED_CONTACT_CARD_OPERATION_BUDGET_MS` and `HOSTED_LINQ_ATTACHMENT_SEND_ATTEMPT_TIMEOUT_MS`, each owned by the layer that spends it), one pure resolver that derives the two deadlines from one anchor, one private bounded-body reader, and a three-value return type on the existing conflict reader. One optional field was added to the existing `handleHostedRuntimeGroupTool` input. No new type, owner, port, protocol field, or module boundary was introduced.
- Complexity intentionally avoided: no pending state, no reconciliation lifecycle, no propagation of the runner's abort into Web, and no inner deadline on the message POST itself. The last two were considered and rejected: both convert "the card probably arrived" into "the card was abandoned mid-send" for an irreversible effect the member explicitly asked for.

## Hot reply path impact

Not applicable to ordinary replies. Every change is inside the `group.share_contact_card` tool branch. No database call, network call, or awaited operation was added to durable acceptance, provider start, or durable reply handoff. Before/after call counts on the reply path: unchanged (0 added).

No change. No tool description, input schema, prompt file, skill, or model tool mode was touched; `packages/assistant-engine/src/assistant-codex/dynamic-tool-catalog.ts` is not in the diff.

- Product experience: **applicable.** Intended outcome: the member is never told a delivered card failed, never told an unestablished card was delivered, and never invited to duplicate a request that can still send. Direct journey evidence: the suites below drive the real send owner over real HTTP through every acknowledgement case and the real Web handler through a slow authorization. Gap: no live end-to-end iMessage send; that is the post-deploy smoke below, inherited from #1458.
- Prompt: **not applicable.** No prompt or tool-definition text changed.
- Frontend: **not applicable.** The only `apps/web/app/**` change is the internal web-control route handler `app/api/internal/hosted-execution/groups/tool/route.ts` (+6/−0), which renders nothing; no component, stylesheet, or asset path changed.
- Coverage: **applicable.** Focused local proof is listed under Verification.

ReviewGPT first-reviewed head: d5bddaa5f0bd5025c0f1b70217436ca85af608ed

Reason: the diff changes deadline and abort behavior around an irreversible external delivery boundary and changes how provider responses are classified.

### Round 2 retrospective (required before round 3)

Round 2 returned `ROUND_OUTCOME: RETROSPECTIVE_REQUIRED` because the round-1 remediation repeated
the mechanism round 1 had faulted. Round 1's High was that work could continue past the caller's
deadline. The round-2 fix moved deadline ownership to the route and bounded the response bodies —
and its bounded reader raced `response.text()` against a timer, then cancelled through
`response.body`. `response.text()` has already locked the stream, so that cancel threw, the helper
swallowed the error, and the losing read and its socket stayed alive. The deadline returned a
value without ending anything. That is the same "the caller stops waiting, the work does not stop"
shape, one layer down, and the reviewer was right to stop us rather than take another patch.

**Restated requirement.** One direct request produces at most one card and one truthful result
that reaches the turn while it is still waiting. "Reaches" is the part these rounds are about, and
an operation is not complete while something it started is still running.

**First reviewed head vs now.** At `d5bddaa5f0` the deadline was created inside the group-tool
handler, after the route had already read, verified, and nonce-checked the callback, and the
response bodies were unbounded. Now the route stamps the arrival time as its first statement and
both deadlines derive from that one anchor; every response read on this path is bounded *and*
terminated; and each send attempt takes the lower of its own budget and the caller's remaining
time.

**The requirement-level decisions round 2 asked for.**

1. *Completion means the consumers are finished, not that the caller gave up.* The bounded reader
   owns its stream reader and cancels through it, which ends the read and the connection. The
   proof is at the provider: the loopback double counts its own sockets and the test asserts they
   return to zero through production code, before teardown touches anything. It fails against the
   previous implementation, which is what makes it evidence rather than decoration.
2. *Body contents are required, so the body consumer owns consumption and termination.* They are
   required: the message identity comes from the accepted body, and the provider's exact conflict
   shape is the only thing that distinguishes a proven duplicate from an ordinary 409. So the
   consumer was fixed rather than the classification deleted. No new timer, state owner, or
   reconciliation lifecycle was added — the change is that one reader holds its own lock.
3. *The 20-second contract governs provider admission, not database latency.* Direct-route
   authorization and the backup-number read take no abort signal. They are checked against the
   deadline before the irreversible POST, so a slow database can only produce a late refusal,
   never a late card. Bounding them would mean returning while the query keeps running — exactly
   the false bound this round removed — so the honest move is to state the limit rather than
   simulate a guarantee. The body now says this plainly instead of claiming the operation cannot
   outlive its budget.

**Decision: continue, with the boundary closed rather than re-timed.** Not a redesign and not a
split. The round-2 delta is +118/−35 production lines against the first-reviewed head; this round
adds ~20 more and deletes the race. Authored-source churn is well under the retrospective
threshold, and the direction is now converging on deletion of false guarantees rather than
addition of new ones.

## Change-shape breakdown

Classification rule: `**/src/**` and other non-test production files are source; `**/test/**` is tests/fixtures; `agent-docs/**` is docs; build/CI/lockfiles are config/tooling; committed build output is generated/other. Counts are `git diff --numstat origin/main...HEAD`. No binary files.

| --- | --- | --- |
| Source | 353 | 66 |
| Tests/fixtures | 470 | 1 |
| Docs | 0 | 0 |
| Config/tooling | 0 | 0 |
| Generated/other | 0 | 0 |
| **Total** | **823** | **67** |

## Design proof

Not applicable. The one `apps/web/app/**` file in this diff is the internal web-control route handler `app/api/internal/hosted-execution/groups/tool/route.ts` (+6/−0), which renders nothing; no page, component, stylesheet, or asset changed (`apps/web/src/components/**`, `apps/web/app/globals.css`, and `apps/web/public/**` are untouched), so the design-page catalog and screenshot requirements do not trigger.

## Verification

- `pnpm typecheck:packages` — pass
- `pnpm --dir apps/web exec eslint` on every changed file — pass, 0 warnings
- `apps/web`: `hosted-group-tool.test.ts`, `hosted-group-tool-route.test.ts`, `hosted-onboarding-linq-contact-card-share.test.ts`, `hosted-onboarding-linq-contact-card.test.ts`, `hosted-onboarding-linq-attachment-idempotency.test.ts`, `hosted-onboarding-linq-http.test.ts` — 284 of 285 passed
- `packages/hosted-execution`: `parsers.test.ts`, `generated-contact-card-parser.test.ts` — 72 passed
- `packages/assistant-engine`: `assistant-codex-group-tool.test.ts` — 86 passed

The one non-pass is `hosted-onboarding-linq-http.test.ts` → "keeps the chat-read deadline active through a stalled response body", a wall-clock assertion (`< 1500ms`) that only fails in the combined local run under load. It passes alone, and it reproduces identically with this PR's source changes stashed, so it is pre-existing and unrelated. Exact-head CI is the authority on it.

`apps/web/test/hosted-onboarding-linq-attachment-idempotency.test.ts` starts a loopback provider that models the exactly-once contract of the repository's local Linq support (`apps/cloudflare/test/helpers/hosted-local-linq-support.ts`) — one accepted message per idempotency key, byte-identical resubmission replays the original message identity, a changed body under that key conflicts, and the post-accept controls record acceptance before failing — and drives the real `sendHostedLinqAttachmentMessage` through every case:

| Case | Terminal result | Cards accepted |
| --- | --- | --- |
| Post-accept 503, then reconciliation succeeds | `sent`, original message id | 1 |
| Connection destroyed after acceptance | `sent`, original message id | 1 |
| Send times out after acceptance | `sent`, original message id | 1 |
| Post-accept 503 on both attempts | `unconfirmed` | 1 |
| Attachment-create headers arrive, body stalls | prepare-phase failure at the pre-send deadline, zero message POSTs, connection ended | 0 |
| Attachment-create failure whose body stalls | ordinary failure, connection ended | 0 |
| Upload response whose body stalls | proceeds, connection ended before the POST | 1 |
| Definitive rejection whose body stalls | ordinary failure, connection ended | 0 |
| Retryable send responses whose bodies stall | reconciled, both connections ended | 1 |
| Pre-send deadline already passed | prepare-phase failure, zero message POSTs | 0 |
| Send deadline already elapsed | failure, no request dispatched | 0 |
| Both keyed attempts stall, connections checked | `unconfirmed`, provider sockets closed by production code | 1 |
| Both keyed attempts stall, one caller deadline | `unconfirmed` inside the caller's window | 1 |
| Accepted 200 whose body stalls, keyed | reconciled → `sent`, original id | 1 |
| Accepted 200 whose body stalls on both attempts | `unconfirmed` | 1 |
| Accepted 200 whose body stalls, unkeyed | `sent`, identity lost, never resubmitted | 1 |
| 409 whose body stalls | reconciled → `sent` | 1 |
| Later request reuses the key with a new attachment | key-reuse conflict → `already_shared` | 1 |
| Definitive 400 pre-acceptance | ordinary failure, no resubmission | 0 |
| Proxy-wrapped or unrelated 409 | ordinary failure, no resubmission | 0 |
| Pre-send deadline expires | prepare-phase failure, never `unconfirmed` | 0 |

Every row accepts at most one card, which is the invariant the reconciliation exists to protect. The stalled-body rows additionally assert that the provider's own socket count returns to zero through production code, checked before teardown closes anything; that case fails against the previous bounded-reader implementation, which returned on time while leaving the read and connection alive. The boundary itself is proven separately: `apps/web/test/hosted-group-tool.test.ts` drives the real route with signed-callback verification delayed past the pre-send deadline and asserts the refusal happens with zero provider admission, that the anchor predates the delay, and that a fast boundary still admits the send.

Not proven locally: a single test traversing runner → proxy → the real Web route → Linq in one process, with real transit and a real cold start. That lane is the hosted-local stack. What is proven here is the part this PR owns — that the Web-side operation is anchored at the route boundary, that no card is admitted after the deadline, and that every response it opens is ended — and the live path is covered by the post-deploy smoke below.

## Deployment concerns

Deploy **Web first, then the Cloudflare Worker/runner bundle**, the same order #1458 used.

- Skew behavior, correct order: every change here is inside Web. The new `contact_card_presend_deadline_exceeded` reason travels in the existing `unavailableReason` string field, which the currently deployed runner's parser already accepts, so new Web is compatible with the old runner and no runner deploy is required for this PR to be correct.
- Skew behavior, wrong order: no new runner-to-Web request field was added, so a new runner against old Web is also compatible. This PR does not create a skew hazard in either direction; the ordering above is inherited from #1458's own contract rather than required by this diff.
- Worker/container skew: during gradual rollout a new Worker version can serve traffic while warm `RunnerContainer` processes still run the previous runner bundle. Both are compatible with this Web change.
- Canonical shares: unchanged in key derivation, reservation behavior, and response shape.
- No schema migration, no backfill, no new environment variable, no runner bundle size change.
- Post-deploy check: make one explicit generated contact-card request in a direct iMessage thread and confirm the `.vcf` arrives with the generated photo and can be saved over the existing Murph contact; confirm a canonical group contact-card share still works and is still owner-only. This is the same end-to-end evidence gap #1458 recorded, still owed.


rocketman-21:
### ReviewGPT anomaly retrospective — round 2

**Original requirement.** Personalized contact-card delivery must not outlive the runner's control decision in a way that permits an unattended or duplicate send. A terminal provider result requires the owned provider response-body consumer to have completed or been terminated; exactly one idempotency identity remains the source of truth.

**Shape movement.** The first-reviewed head contained 173 additions / 40 deletions of authored production source and 108 additions / 1 deletion of tests. The round-2 head contained 273 additions / 57 deletions of production source and 301 additions / 1 deletion of tests. Review-driven growth introduced the route-owned timestamp, derived prepare/send deadlines, bounded attachment-create and send-body reads, and their focused timing proofs.

**Repeated mechanism.** The bounded reader raced `response.text()` against a timer. `response.text()` had already locked the stream, so the timer's `response.body.cancel()` failed and its error was swallowed. The helper returned on time while the losing reader and provider socket remained alive. This repeated the accepted deadline-ownership failure: the promise was bounded, but all work it owned was not terminal.

**Decision: continue this PR with a smaller correction.** Keep the route-owned timestamp, single idempotency key, and existing reconciliation path. Replace only the bounded-body implementation so it owns the stream reader directly and cancellation is terminated and joined; add provider-side connection accounting that fails against the broken reader and proves production code closes stalled connections without teardown doing it. Add no owner, state machine, queue, lease, lifecycle, compatibility path, or extra reconciliation mechanism.

**Contract decision for non-cancelable reads.** The 20-second inner deadline governs admission to the irreversible provider POST and the lifetime of provider response consumption. It does not claim to abort authorization or backup-number database operations that expose no cancellation primitive. Those reads are checked after completion, so an expired pre-send deadline refuses before any provider POST; the runner's existing 30-second control timeout remains the outer bound. The PR body will state this narrower contract rather than claiming every Web-side await is cancellable.

rocketman-21:
**ReviewGPT final gate: `ROUND_OUTCOME: PASS` at `537a92a81c`** (round 4, `gpt-5-6-pro`). Zero accepted findings; the six items it raised were wording discrepancies, all corrected in the body above.

Round history: R1 High — the deadline started inside the handler, after the route had already read, verified, and nonce-checked the callback, and the attachment-create body was unbounded. R2 `RETROSPECTIVE_REQUIRED` — the first bounded reader raced `response.text()` and cancelled through `response.body`, which throws on a locked stream, so it stopped waiting without ending anything; retrospective recorded in the body. R3 two accepted findings — status-only responses had no terminal owner at all, and `AbortSignal.timeout` was being used as a deadline predicate when it is only a schedule. R4 PASS.

**CI status.** Every check passes except **No horizontal overflow on marketing pages**, which is a `main`-side breakage this PR does not touch: `apps/web/e2e/pr-1498-design-proof.spec.ts` (added by #1498) asserts `DESIGN_PROOF_OUTPUT_DIR`, and Playwright's `testDir` runs it in that job too. The job is red on `main` itself from `cca3c15b99` onward, and PR CI tests the merge with `main`, so it is red on every open PR. #1511 fixes it. This PR goes green once that lands.

PR #1533, opened 2026-08-10
Preserve generated-image continuity for group avatars

rocketman-21:
## Why this PR exists

A trusted background image completion was forking from the foreground provider thread that launched it, so a later explicit group-avatar request could not see the generated capture and incorrectly treated an existing capability as missing.

After Murph generates and visibly delivers an image, a user can explicitly ask Murph to reuse that exact generated capture as the group photo. Native provider resume keeps the fast path coherent; if resume is rejected after a route or contract change, runtime-owned transcript provenance restores the exact ref without granting delivery or mutation authority. Before private publication or mutation, the group-avatar boundary classifies generated origin from the retained marker, current compound-turn restriction, or existing generated-capture lookup, then independently requires current-session outbox proof that the exact media was physically accepted.

## User experience

- Entry: a user asks Murph to generate an image. Existing progress/acknowledgement behavior begins immediately, while the existing hosted image-generation owner handles the roughly minute-class background wait.
- Continuation: the trusted completion resumes the initiating provider thread and attaches only its exact hash-bound vault image. The completion itself cannot mutate the group, alter settings, file feedback, or borrow generic user authority.
- Destination: after the image is visible, a later explicit user request can reuse its exact `raw/captures/**` reference through `murph.group set_chat_avatar`; the resulting group photo is visible to the existing group audience.
- Recovery: failed or malformed completions keep the existing plain-language failure path and fail closed. Pending, mismatched, or unreadable generated-image delivery evidence cannot publish or mutate the avatar; Murph says the image must be visible first. An accepted Linq image primary remains eligible while its logical rich-link delivery is retryable, sending, or terminally failed because the exact physical media effect already proves visibility.

This is the smallest safe flow: one generation request, the existing asynchronous delivery, and one explicit group-mutation request; there is no new screen, choice, queue, or state owner.

- Provider continuity is not effect authority: a trusted system completion may retain the foreground tool contract without gaining user-action authority.
- Completion effects are turn-local and limited to no reply, the exact validated media attachment, or the existing exact-origin physical-note continuation.
- A later accepted foreground input independently authorizes group mutation and may reuse only an exact owner-written `raw/inbox/**` or `raw/captures/**` reference.
- Provider continuity and transcript provenance do not prove visibility: a generated completion ref becomes eligible for later avatar reuse only when the existing outbox proves its exact media was visible.
- Vault bytes, media metadata, hash, size, content type, filename, source, and saved ref must agree before attachment.
- Existing mailbox ordering, capture ownership, group preflight, publication, delivery idempotency, and failure behavior remain intact.

## Changelog

- Changelog: updated
- Items: 2026-08-10 · generated-image-group-photo

- Dynamic-tool fingerprinting/provider thread reuse: trusted completion retains the foreground style-settings contract so native continuation does not fork. Native resume remains an optimization; exact generated-image response provenance is also restored from the existing committed transcript owner when route support or a contract fingerprint rejects resume. That history is truthfully labeled as neither delivery nor effect authority. Regression proof: hosted completion continuity, finalizer, response-media, and turn-planning tests.
- Explicit media replies: each current multi-message Linq delivery records a true-only media-owner fact on the exact physical provider effect that carried the intent media. Image/text primary effects and successful dedicated voice effects retain media; link and text siblings do not. An accepted image primary whose terminal rich-link sibling fails preserves that same effect through the existing retryable, sending, and terminal-failed checkpoint lifecycle; only an explicit Linq native reply to a non-confirmation-pending checkpoint with exactly one accepted marked effect may use it. Native replies derive exact generated media directly from their already provider-id-bound outbox record instead of rejoining transcript state. Generic outbox readers remain sent-only; pending, awaiting-approval, abandoned, unmarked, duplicate-owner, unknown-ID, and confirmation-pending records remain excluded. Single-message legacy deliveries remain unambiguous, while multi-message legacy records without ownership evidence fail closed for media. Two-capture, successful and partial image/link, terminal-before-scan, text/voice, and voice-fallback coverage prove that a native reply cannot inherit an adjacent physical effect.
- Generated-avatar delivery eligibility: after the existing group preflight and source resolution but before provider or publisher egress, a retained exact generated-completion marker must join the same session and exact resolved ref/hash/content-type/size on a sent outbox intent or the narrowly attested accepted Linq primary above. The delivery may come from a later turn that visibly sends the same immutable media. If bounded transcript retention removed the marker, the current completion restriction or lazily materialized generated-capture lookup recovers origin classification only; a singleton same-session outbox intent must still carry the exact resolved descriptor and accepted physical-delivery evidence. The direct publisher and generated-reference provider consume the same byte snapshots that passed verification. Healthy ordinary canonical capture refs absent from all three provenance sources preserve the existing path. Pending, awaiting-approval, abandoned, missing, unmarked, byte-replaced, duplicate-owner, unknown-ID, confirmation-pending, and unreadable generated delivery evidence fail closed. Regression proof covers later-turn exact delivery, wrong-session/hash rejection, same-ref byte replacement, compound completion admission, markerless known-generated rejection, trimmed-marker delivery recovery, and the existing production boundary before provider/publication/mutation.
- Hosted tool authority: the delivery context carries an engine-owned completion restriction and hides accepted-input, feedback, personalization, and unrelated mutation authority until a later accepted foreground input becomes current. Adjacent trusted completions are isolated by the existing grouping owner, so each background turn retains exactly one restriction while the final completion may still fold its later same-route foreground input. Regression proof: the dedicated image-completion-authority, grouping, and automation reply event-path tests.
- Native completion capabilities: while that trusted completion scope is current, the existing turn runner forces approval `never`, a read-only sandbox, native-disabled thread config, and process config overrides that cannot be re-enabled by requested settings; injected environments, hosted fetch adapters, and live provider steering are absent. Native resume and the hosted exact-media/physical-note effects remain wired. A later accepted foreground input stays with the existing mailbox owner for the next unrestricted turn, where the ordinary native route and its existing steering controller are restored. Regression proof: the provider-boundary final-coverage and active-turn controller suites.
- Media reference semantics: the group schema now truthfully distinguishes user-sent inbox media from generated captures, while the resolver continues to treat provenance separately from action consent. Regression proof: group-tool and image-completion tests.
- Deferred capability discovery: eager product-feedback guidance now requires deferred-tool discovery before declaring a Murph capability missing. Regression proof: product-feedback and scripted-runtime tests.
- Physical-note continuation: exact-origin continuation remains available only through its existing owner and exact-media validation. Regression proof: image-completion-authority coverage.

## Architecture and reuse

- **Existing systems reused:** the hosted mailbox/session owner, native provider thread identity, canonical `raw/captures/**` store, delivery context, dynamic-tool executor, existing group-avatar action, and physical-note continuation.
- **New logic:** parse exactly one trusted completion even inside its intentional completion-first compound batch, retain its foreground provider contract, derive a turn-local effect restriction, compare any attachment against the exact saved vault-image metadata, commit one bounded runtime provenance marker for every trusted ready completion, attest which physical Linq effect carried intent media, and require existing outbox visibility before a generated ref classified by the marker, restriction, or capture lookup can drive a later avatar effect.
- **New abstractions:** one narrow `AssistantHostedImageCompletionEffectRestriction` contract bridges auto-reply admission to the existing hosted tool context; the generated-image marker reuses the existing transcript owner and status-entry schema. One optional true-only fact extends the existing provider-effect record, and one hosted verification callback derives from transcript, generated-capture lookup, and outbox state instead of creating another persistence owner.
- **Complexity intentionally avoided:** no queue, session manager, capability fork, media database, replay service, compatibility shim, or second authority owner; the correction stays inside current session, transcript, mailbox, outbox, media, and tool boundaries.

## Hot reply path impact

Applicable only when a provider turn calls `set_chat_avatar` with a direct `raw/captures/**` image ref or includes such refs as generated-avatar inputs. After the existing group preflight and the route's existing source resolution, but before provider or publisher egress, the effect boundary invokes the existing generated-delivery verifier with each exact resolved ref/hash/type/size descriptor. Every invocation starts current-session transcript, outbox-inventory, and generated-capture-lookup reads in parallel; the existing hosted workspace materializer first hydrates the compact lookup when it is lazy. A direct ref adds one verifier call after its one required source read. Generated-avatar requests add one call per resolved capture input, bounded by the existing 16-reference schema; `raw/inbox/**` inputs add none. There is no extra source read, provider request, retry loop, or persistence write; any rejected descriptor fails the complete operation before the image provider, private publication, or group mutation. Healthy ordinary capture refs remain eligible only when marker, current restriction, and lookup classify them as ordinary. Native explicit replies still derive exact delivered media from the already provider-id-bound outbox record. Ordinary messages, text/link siblings, background completion, and requests that do not call the avatar action add no operation.

Deterministic final-boundary capture compared merge base `dbfa6ae1292102063b9829d153d390e5e13a14ce` with the initial implementation using identical synthetic individual/group Linq inputs, pinned `gpt-5.6-terra`, low reasoning, production code mode with route-specific dynamic tools, the real pinned Codex App Server, and a local scripted Responses recorder. Token counts use `gpt-tokenizer` 3.4.0 `o200k_harmony`. The specialist remediation restored two conditional required-field phrases. The delivery-eligibility correction then tightened the already changed generated-image descriptions without removing request/preference/privacy, character-sheet, asynchronous-hosted, exact-scheduled, local-save, reference-order, or capture-reuse rules. Exact serialization of the two complete changed tool objects removes 173 tokens / 767 bytes from both previously captured requests. The current remote base is a descendant of the measured merge base and has no prompt-bearing change in this comparison.

| Runtime | Base | Head | Token delta | Byte delta |
| --- | ---: | ---: | ---: | ---: |
| Individual | 38,177 tokens / 173,190 bytes | 38,063 / 172,716 | -114 (-0.2986%) | -474 (-0.2737%) |
| Group | 30,067 tokens / 139,329 bytes | 29,953 / 138,855 | -114 (-0.3792%) | -474 (-0.3402%) |

Included fields: `include`, `input`, `instructions`, `parallel_tool_calls`, `text`, `tool_choice`, and `tools`, with paths and unstable UUIDs normalized identically. Excluded transport-only fields: `client_metadata`, `metadata`, `model`, `prompt_cache_key`, `prompt_cache_retention`, `reasoning`, `service_tier`, `store`, and `stream`. The assembled authored instructions are byte- and token-identical in both routes; the complete delta is `dynamic-tool-catalog.ts` tool/schema/generated guidance. Request assembly, wrappers, fixtures, tokenizer, and configuration were held constant.

- **Product experience — applicable:** intended outcome is the complete generated-image-to-explicit-group-avatar journey. The accepted specialist finding added an opt-in real App Server three-turn journey: truthful generation acknowledgement, exact completion attachment without mutation, then explicit avatar reuse with feedback available but unused. Deterministic suites own fallback and authority proof; the member-visible changelog journey is rendered at desktop and mobile widths through the existing production archive component and its inert synthetic design study.
- **Prompt — applicable:** model-facing group media and product-feedback discovery contracts changed. The accepted specialist finding restored both conditional required-field descriptions; deterministic schema assertions and provider-input measurement cover the final wording.
- **Frontend — applicable:** the member-visible changelog item is rendered through the existing archive component. The existing `/design?tab=sections#changelog-archive` study exercises the generated-group-photo feature card with synthetic inert props; the complete focused changelog/design set passes 54 tests and the full Web typecheck passes. Desktop and mobile captures cover the inert study; static-render tests prove the current production August 10 edition. No production component, layout, interaction, request, or live-data dependency was added.
- **Coverage — applicable:** the accepted final findings added exact marker persistence, two native-resume rejection modes, two-capture reply disambiguation, a mismatched-media negative seam, exact physical-note scope proof, and generated-delivery eligibility at the real group-tool boundary. Production-shaped regressions exercise two same-alt images, image-plus-link success, accepted-primary retry, terminal-before-reply-scan, one text-plus-voice response, compound completion admission, trimmed-marker recovery, and the generated-avatar visibility boundary through existing owners. The image primary retains exact provenance across its accepted physical lifecycle, link and text siblings do not inherit media, the dedicated voice effect remains attested, and no generated avatar is published before visibility. The complete current affected set passes 623 tests with 47 credential-gated cases skipped. Package build/typecheck, Cloudflare Node coverage, and exact runner assembly are green.

## Final review remediation

- Round 1 findings were resolved by restoring exact conditional prompt fields, adding the complete journey coverage, and persisting bounded runtime transcript provenance for the native-resume fallback.
- Round 2 found that ordinary Linq image delivery persists non-null alt text in `providerMessageEffects`, while the reply resolver incorrectly used a null message as the gate for exact generated-image provenance. The retrospective reset the invariant to provider message id → exact sent outbox delivery → exact media → runtime marker; visible fallback text cannot erase media identity and the marker grants no action authority.
- The remediation deletes that null-message proxy, resolves a matching exact marker before the text fallback, and retains the bounded visible text. The production-shaped same-alt two-image regression failed before the fix and passes afterward, exposing only the replied-to image's ref and hash.
- Round 3 found that Linq can split one image intent into a primary image/text bubble and a link-only sibling, while intent-level media was copied onto both exact provider-effect projections. Its first correction preserved ordered ids and stripped media from non-primary effects; the image/link regression proved both physical replies.
- Round 4 found that position is not a generic media-owner fact: successful Linq text-plus-voice delivery sends the text first and the actual voice memo second, so the positional correction erased attested media from a reply to the voice bubble. The required retrospective reset ownership to an explicit fact emitted by the physical sender.
- The final correction deletes positional inference and adds an optional true-only `carriesIntentMedia` fact to the existing provider effect. Current image primaries and successful dedicated voice sends set it; text/link siblings do not. Multi-message records without it retain no media, while one-message legacy records remain unambiguous. Delivery equality includes the fact so mirror reconciliation cannot discard marked ownership evidence. The real text-plus-voice dispatch/reload regression proves the text reply quotes only text and the voice reply receives media/no-text context without transcript, filename, or provider-id leakage. The image/link and split-fallback regressions remain green, and downstream group mutation retains its existing accepted-input, preflight, vault, publication, and provider owners.
- Round 5 found that a version-15 rollback would reject the new key inside the strict provider-effect schema, quarantine the complete sent outbox intent during an ordinary vault-backed read, and omit that quarantine from a later portable snapshot. The retrospective reset the persisted-state invariant: a writer may activate only after every supported rollback target can parse and retain its records, or after the existing platform enforces a reader-capable hard floor.
- The remediation advances the existing `RunnerStateStore` schema to version 16. `HostedUserRunner` constructs that store before any invocation, workspace snapshot, or container service, so a version-15 Worker rejects the Durable Object before it can wake an old runner or read/quarantine a workspace. Production immediate rollout remains fail-closed in preflight, exact bundle-fingerprint admission blocks stale warm runners from becoming writers, and the durable deploy contract makes version 16 a forward-fix rollback floor without adding a compatibility service, queue, or second state owner.
- Round 6 found that the accepted primary image in a partial Linq image/link delivery remained visible and had an exact provider id, but its physical media-owner effect was dropped from the partial error and the resulting retryable checkpoint was excluded from exact reply lookup. A native reply before `nextAttemptAt` could therefore lose the exact image join even though retrying the absent link sibling later converged.
- The correction carries the existing primary effect through the Linq partial error and reconstructed delivery. Explicit reply resolution alone admits a retryable Linq checkpoint when `deliveryConfirmationPending` is false and an exact provider effect carries the media-owner fact. The real endpoint regression persists/reloads that checkpoint, selects only the replied-to capture when another generated image exists under fresh planning, then proves same-idempotency retry convergence without a second primary acceptance or replaying the processed inbound.
- The first attempted round-7 capture did not produce a valid result for its intended head. A later same-thread export explicitly checked older head `a8807b6ef2` rather than the `a0f8d3044d` candidate and lacked concrete model confirmation, so it remains diagnostic only and does not advance the substantive round or satisfy the final gate.
- That stale response nevertheless identified an ORIGINAL_PR ordering defect which parent inspection reproduced before fixing: provider completion persists native-resume and transcript refs before the exact outbox media is visible, allowing a later accepted avatar request to use an unseen generated capture. The first correction required exact existing outbox delivery evidence at the group effect boundary, before private publication or mutation. On that candidate, pending, failed, unmarked retryable, confirmation-pending, mismatched, and unreadable evidence failed closed; sent and the already attested visible Linq primary qualified. Native replies deleted their redundant transcript join and used the stronger delivered outbox proof they already owned. Its regression proved the pending block, ordinary-capture preservation, and successful reuse after the same intent became sent; the following round exposed and corrected the overly broad terminal exclusion.
- A subsequent fresh round-7 full snapshot validly checked `a0f8d3044d`, while the delivery-eligibility correction advanced the PR to `960c898fb7` before that response completed. It therefore did not certify the newer head, but it found a repeated status-coupling gap which parent inspection reproduced on the current shared predicate: the exact accepted primary effect survives a retry becoming `sending` and then terminal `failed`, while both native reply context and direct generated-ref reuse discarded it solely because the logical intent was no longer `retryable`.
- The round-7 correction keeps the existing outbox as the only delivery owner. The opt-in evidence predicate admits `retryable`, `sending`, or `failed` only when confirmation is not pending, the delivery is Linq, and exactly one accepted provider effect carries intent media. Generic readers stay sent-only, and pending, awaiting-approval, abandoned, missing, unmarked, duplicate-owner, unknown-ID, and sibling effects fail closed. The real endpoint regression retains the pre-retry convergence path and adds the terminal-before-scan order: due outbox work terminally fails first, the same pass scans the accepted reply second, only the replied capture's exact ref/hash appears with another generated image present, the primary is accepted once under the same idempotency key, and the inbound is not replayed.
- The seven-round hard-cap retrospective and explicit `continue` decision are recorded in the [continuation comment](https://github.com/cobuildwithus/murph/pull/1533#issuecomment-5241117061). Reverting or splitting would leave the exact generated-image journey or a proven authority/delivery invariant unresolved, while the current shape reuses existing owners and adds no queue, state machine, reconciliation service, database owner, or compatibility subsystem. A first round-8 invocation was invalid and did not advance the counter because a concurrent docs-only plan commit raced packaging; the ZIP correctly rejected its old-diff/new-plan mismatch before substantive review.
- Valid round 8 found that a trusted ready completion could finish without attaching media, retain provider continuity, and persist no generated provenance marker. Because markerless ordinary captures intentionally remain eligible, that generated ref could later bypass outbox visibility proof. The [repeated-mechanism retrospective](https://github.com/cobuildwithus/murph/pull/1533#issuecomment-5242330701) kept the existing transcript and outbox owners: every trusted ready completion now persists the existing provenance-only marker, using the actual media ordinal when attached and the completion response ordinal otherwise. The marker grants no authority; empty or mismatched outbox evidence still rejects avatar reuse. A focused pre-fix regression failed with no marker, then passed while preserving no-reply provider continuity and proving empty-outbox rejection.
- Two ordinary merges reconciled current `main`; the first resolved four textual conflicts and the second was conflict-free. Exact post-merge assembly deliberately remeasured the combined Health Commons and generated-image graph, preserving the existing entry/static tolerances and 32 KiB runner allowance.
- Round 9 found the same provenance invariant at two remaining boundaries: the restriction builder discarded an intentionally compound completion-first batch, and bounded transcript retention could later remove the only generated-origin marker while the capture lookup remained live. The [repeated-mechanism retrospective](https://github.com/cobuildwithus/murph/pull/1533#issuecomment-5243496768) retained the existing owners. Exactly one trusted completion now carries its restriction through later same-route input; the verifier classifies origin from that restriction or the lazily materialized existing lookup and still requires same-session exact-ref outbox delivery. A retained marker keeps the stricter turn/ref/hash/type/size join. Refs absent from marker, restriction, and lookup preserve the ordinary path. Round 10 is the required fresh full-snapshot audit for this behavior fix.
- Valid round 10 found two remaining cardinality/turn-coupling defects. Multiple adjacent trusted completions could enter one group and make the exact-one restriction builder fail open, while a retained no-attachment marker rejected a later same-session delivery of the same exact immutable media solely because its outbox intent belonged to a later turn. The [round-10 repeated-mechanism retrospective](https://github.com/cobuildwithus/murph/pull/1533#issuecomment-5246231513) keeps grouping, transcript, and outbox as the existing owners and rejects a multi-completion restriction type, queue, manager, lifecycle, compatibility path, or alternate delivery truth.
- The correction isolates completion A from completion B in the existing adjacency predicate but still folds completion B with its later same-route foreground input. The delivery join deletes only original-turn equality while retaining same-session, exact ref/hash/type/size, and accepted physical-delivery proof. The grouping regression failed before the source guard; the merged affected set passes 270 tests with 55 live-provider cases gated, including later-turn delivery plus wrong-session/hash negatives.
- Settled-tree correction proof passes 184 tests across grouping, group-tool, and runtime coverage, plus the Assistant Engine typecheck. The earlier broad diff run's single grouping failure was a timing artifact from loading the pre-correction source while the guard landed; the exact settled-tree file and affected suite both pass.
- Current `main` merged with one overlapping live-provider test conflict. The resolution preserves this PR's complete generated-image journey and main's fresh shared-data/detached-consultation cases.
- A latest-package ReviewGPT 0.5.124 round-11 attempt on `409d0da7baba` verified requested `gpt-5.6-sol` and response `gpt-5-6-pro`, but emitted unsupported `ROUND_OUTCOME: FAIL`. It therefore remains diagnostic only and does not advance the substantive round or satisfy the merge gate. Its one finding was reproduced: the hosted restriction denied Murph dynamic effects while the provider attempt still inherited foreground native Codex capabilities.
- The accepted correction reuses the existing turn-runner restriction boundary. Trusted completion receives a read-only sandbox, approval `never`, native-disabled thread and process config after any requested overrides, no injected environments, and no hosted fetch adapters. It preserves native resume and exact hosted effects; the later accepted foreground route remains unchanged. No queue, owner, state machine, compatibility path, or alternate authority source was added. Current `main` was then merged ordinarily, preserving this PR's generated-photo changelog alongside Starter, web-search, appointment-reminder, and friction-log changes. A valid substantive round 11 must review the exact pushed merged head.
- Valid substantive round 11 reviewed exact head `54e026cf2777` with ReviewGPT 0.5.124, verified requested `gpt-5.6-sol` and response `gpt-5-6-pro`, and returned one accepted finding: a later foreground input could still live-steer into an already launched native-restricted completion attempt even though native sandbox, thread, process, environment, and fetch capabilities cannot recover mid-turn.
- The [round-12 repeated-mechanism retrospective](https://github.com/cobuildwithus/murph/pull/1533#issuecomment-5247477661) keeps the existing mailbox and turn-runner owners. The correction passes `null` for provider steering only on the already restricted completion attempt; the later foreground input remains queued for the next unrestricted turn, which receives the unchanged existing steering controller. The focused regression failed before the one-boundary source correction and passes after it. No new queue, owner, state machine, lifecycle, compatibility path, or dynamic capability mutation was added.
- Current `main` was merged normally twice after the correction. The first resolution preserves this PR's generated-photo changelog alongside current Starter, patterns, reminders, cards, and managed web-search items. The later Mobvoi/Health Connect merge was conflict-free and is base-only relative to the current PR patch. Round 12 is the required sensitive fresh full-snapshot audit of exact pushed head `0da0e52b6bcf`.
- Valid substantive round 12 reviewed exact head `0da0e52b6bcf` with ReviewGPT 0.5.124, verified requested `gpt-5.6-sol` and response `gpt-5-6-pro`, and returned one accepted review-induced finding: the direct generated `image_ref` path passed the visible-first verifier, but the sibling generated-avatar route could consume unseen generated `referenceImageRefs` before generation, private publication, and group mutation.
- The [round-13 repeated-mechanism retrospective](https://github.com/cobuildwithus/murph/pull/1533#issuecomment-5247961936) keeps the existing group-tool boundary and verifier as the sole owner. The correction derives candidates from both avatar modes, deduplicates `raw/captures/**` refs, and fails the complete operation if any existing verifier result is false. The red/green regression proves unseen, mixed, delivered, and ordinary-capture cases without adding a verifier, queue, state owner, authority source, lifecycle, database, or compatibility path. Current `main` was then merged normally; five textual conflicts preserve the latest runner ratchet, generated-photo plus selected-voice changelog copy, and a canonically regenerated CLI skill hash. Round 13 is the required sensitive fresh full-snapshot audit of exact pushed head `3fdf0b6cedda`.
- Valid substantive round 13 reviewed exact head `3fdf0b6cedda` with ReviewGPT 0.5.124, verified requested `gpt-5.6-sol` and response `gpt-5-6-pro`, and returned one accepted original-PR finding: vault-wide count pruning could delete recent generated-image delivery evidence after 100 newer terminal intents while the generated capture remained live for 14 days.
- The [round-14 repeated-mechanism retrospective](https://github.com/cobuildwithus/murph/pull/1533#issuecomment-5248422217) keeps the existing outbox pruning owner. Singleton `gpt-image-2` vault-image intents under `raw/captures/**` are excluded only from count eviction and retain the existing 14-day age cutoff; ordinary terminal inventory remains capped at 100. Transcript, provider continuity, and capture lookup remain provenance only. No state, lookup, queue, lifecycle, database, delivery owner, or alternate authority source was added. A final conflict-free base-only main merge and its required changelog schema refinement followed. Round 14 is the required sensitive fresh full-snapshot audit of exact pushed head `99ff781624e8`.
- A separate concurrent ReviewGPT 0.5.124 round 14 checked `99ff781624e8`, verified requested `gpt-5.6-sol` and response `gpt-5-6-pro`, and returned PASS. The owning independent round 14 checked that same exact head with the same concrete model verification and returned one reproducible review-induced finding: historical delivery was approved by mutable ref before the direct publisher or generated-reference provider path independently reread the current bytes. The owning response SHA-256 is `e9123cc61d07359d886ceab00e4f9295e7612f539c3fb31f70e82f0ec69e5e5c`, so its qualifying finding keeps the gate open.
- The [round-15 repeated-mechanism retrospective](https://github.com/cobuildwithus/murph/pull/1533#issuecomment-5248893637) keeps the existing resolver, outbox verifier, provider, and publisher owners. The correction deletes the ref-only pre-load loop. Direct reuse verifies the descriptor returned by its one source read and publishes that same byte snapshot; generated references verify immediately after their existing resolution and before provider egress. The production-shaped regressions compute the real fixture hash, succeed with unchanged bytes, replace the same ref with a valid different image, and then fail both routes before publisher, provider, or mutation. Ordinary captures retain their existing path. ReviewGPT 0.5.124 round 15 audited exact pushed head `878a0ffe38c3`, verified requested `gpt-5.6-sol` and response `gpt-5-6-pro`, and returned `ROUND_OUTCOME: PASS` plus `REVIEW_COMPLETE` with response SHA-256 `c77f1f45c49d42d1ec445b0c9908da6fb4c408f6ea528baa9ed809999d27c253`.
- After that zero-finding round, current `main` advanced twice and was merged normally. The additive conflict resolutions preserve generated group photos alongside health-history, challenge-card, and personality additions, retain the runner bundle guards with exact measured ratchets, and regenerate the CLI skill hash from the merged source. Per the review contract, base-only reconciliation after a zero-finding round does not rerun ReviewGPT.
- Exact-head CI then exposed two stale main-owned prompt-header assertions after the appointment-reminder guidance renamed that section. The test-only correction matches the current literal and the focused file passes all 73 tests. An isolated regression-test correction does not rerun ReviewGPT.
- At the authorized merge boundary, current `main` advanced with the changelog-fragment architecture and produced three overlapping changelog conflicts. The behavior-preserving resolution takes main's fragment registry and generalized archive tests, then ports this PR's already-reviewed generated-group-photo entry unchanged into one independent fragment at the same leading position. It authors no new runtime logic or contract, so the base-update-only ReviewGPT exception applies. Final pushed head is `a927f315d4ff` and required exact-head CI owns the merge gate.

Reason: this changes hosted-runtime trust boundaries, background/foreground ordering semantics, provider continuity, and group-mutation authority.

ReviewGPT first-reviewed head: ba55cd37241d6fc777f1fc389e940441428d858a

## Change-shape breakdown

Classification rule: authored `apps/**/src/**` and `packages/**/src/**` are source; `test/**` is tests/fixtures; Markdown architecture/plan files are docs; config/tooling and generated output are separate. No binary files.

| --- | ---: | ---: |
| Source | 1,029 | 160 |
| Tests/fixtures | 3,938 | 81 |
| Docs | 918 | 8 |
| Config/tooling | 19 | 9 |
| Generated/other | 23 | 15 |
| **Total** | **5,927** | **273** |

## Design proof

- Design page: `/design?tab=sections#changelog-archive`
- Coverage: the real `ChangelogEditionSection` rendered against inert synthetic props for the generated-group-photo feature state at desktop and mobile widths. Static-render tests verify the current production August 10 title, summary, leading item, stable order, and selected-voice copy.
- Desktop screenshot: Desktop changelog design proof
- Mobile screenshot: Mobile changelog design proof
- Verification: the current focused changelog/design set passes 54 tests and the full Web typecheck passes. The latest source correction changes no production Web component, layout, interaction, request, or live-data dependency. Only the current desktop/mobile synthetic design-study screenshots are attached; the stale production captures are intentionally excluded.

## Verification

- Assistant Engine final focused tests: 318 passed, 35 live-provider cases skipped behind their explicit gate.
- Outbox delivery-equality proof: 29 passed; generated CLI skill-hash proof: 2 passed; exact CLI package-shape verification passes.
- Hosted provider recovery: 23 passed; Assistant Runtime typecheck passes.
- Cloudflare rollback-floor and deploy-preflight proof: 98 passed; Cloudflare package typecheck passes.
- Operator Config Linq runtime: 66 passed; Assistant Engine and Operator Config package typechecks pass.
- Accepted-primary retry proof and its affected delivery/reply suite: 372 passed across seven files.
- Accepted-delivery lifecycle affected suite: 499 passed across seven files; focused pre-fix regressions failed on terminal-before-scan native reply context and direct generated-ref reuse, then passed after the shared predicate correction.
- Broader affected run: 363 passed, 35 skipped, and one unrelated timing assertion failed under contention; its exact isolated rerun passed in 3.8 seconds.
- `pnpm --filter @murphai/assistant-engine typecheck`
- `pnpm --filter @murphai/assistant-engine build`
- Cloudflare Node lane: 140 files passed, 2,376 tests passed, and two expected cases skipped before the latest base merge. On the exact latest merged tree, runner bundle policy passes 42 tests and full hosted-local assembly passes at 8,689,343-byte Vault CLI total, 1,596,214-byte runner entry, 7,726,737-byte static closure, and 9,618,405-byte runner total with no forbidden boot input.
- Final post-merge focused proof: 439 Assistant Engine tests passed with 47 credential-gated cases skipped; 66 Operator Config Linq tests passed; Assistant Engine, Operator Config, and Cloudflare typechecks passed; CLI package-shape verification passed.
- Round-10 affected proof: 623 Assistant Engine tests passed with 47 credential-gated cases skipped; the focused compound/retention suite passed 176 tests; Core capture-owner tests passed 13 tests; Assistant Engine and Core typechecks plus package builds passed.
- Exact-head CI remediation proof: the canonical App Server authority regression passes 3 tests; changelog and product-feedback archive proof passes 44 tests; the full Web typecheck passes after its required Health Commons and Prisma generators. After CI exposed one stale edition-title assertion, the focused changelog/page/feedback set passes 52 tests and verifies the new stable item anchor.
- `pnpm docs:drift`
- `git diff --check`
- Exact-head added-line privacy/secret scan
- Exact merge-base/head provider-input measurement for individual and group routes
- Round-11 merged affected proof: 270 Assistant Engine tests passed with 55 explicit live-provider cases skipped; Assistant Engine typecheck passed.
- Design-catalog/changelog proof: 18 targeted catalog/page tests and the complete 52-test focused changelog/design set passed; the full Web typecheck passed. Desktop/mobile design-study and production-page captures are included as exact rendered evidence for round 11.
- Exact merged runner proof: 67 bundle-policy tests passed; fresh assembly measured 8,690,580-byte Vault CLI total, 1,596,168-byte runner entry, 7,727,794-byte static boot closure, and 9,619,462-byte runner total without a ratchet change.
- Native completion correction proof: the focused provider-boundary file passes 23 tests; five affected runner files pass 397 tests; Assistant Engine typecheck passes. Seven broader affected files pass 281 tests with 55 explicitly credential-gated live-provider cases skipped. The merged changelog/design set passes 53 tests and the full Web typecheck passes after its required generators.
- Fresh correction assembly passes at an 8,691,853-byte Vault CLI total, 1,596,168-byte runner entry, 7,729,067-byte static boot closure, and 9,620,735-byte runner total without a ratchet change. Docs drift, whitespace, and candidate identifier/secret scans pass.
- Round-11 steering correction proof: the focused regression failed before the source change because the restricted attempt received the live steering controller, then passed after suppression. The merged six-file Assistant Engine set passes 419 tests; the existing controller suite proves unregistered foreground input remains with the mailbox owner. The focused changelog/design set passes 53 tests. Assistant Engine and Web typechecks, 58 runner bundle-policy tests, docs drift, whitespace, and candidate identifier/secret scans pass. Fresh full assembly measures an 8,691,916-byte Vault CLI total, 1,599,516-byte runner entry, 7,732,478-byte static boot closure, and 9,624,146-byte runner total without a ratchet change.
- Round-12 visibility correction proof: the new production-shaped regression failed before the source change because an unseen generated reference reached the image provider instead of returning the visible-first result. It now proves unseen and mixed generated references block provider, publisher, and mutation; exact accepted delivery enables reuse; and an ordinary capture remains eligible. The full group-tool file passes 100 tests, five adjacent generation/authority files pass 149 tests, Assistant Engine typecheck passes, and 58 runner bundle-policy tests pass. Fresh full assembly measures an 8,693,269-byte Vault CLI total, 1,599,840-byte runner entry, 7,734,484-byte static boot closure, and 9,626,336-byte runner total without a ratchet change. Docs drift, whitespace, and candidate identifier/secret scans pass.
- Final main reconciliation proof: five adjacent Assistant Engine files pass 149 tests; the combined changelog/design set passes 54 tests with generated group photos and selected-voice copy retained; Assistant Engine and Web typechecks pass; the two generated CLI skill-hash tests and exact CLI package-shape verification pass; 58 runner bundle-policy tests pass. The first exact combined assembly exposed main's total budget 6,189 bytes below the new graph; the ratchet now records the exact combined measurement plus the unchanged allowance. Fresh full assembly passes at an 8,716,121-byte Vault CLI total, 1,599,840-byte runner entry, 7,753,694-byte static boot closure, and 9,675,965-byte runner total. Docs drift, whitespace, and candidate identifier/secret scans pass.
- Round-13 retention correction proof: the real terminal-outbox pruning regression proves accepted generated delivery survives more than 100 newer ordinary terminal intents, ordinary terminal inventory remains capped at 100, and generated evidence is removed after the existing 14-day age cutoff. The complete outbox, response-media, and group-tool set passes 199 tests; Assistant Engine typecheck and 58 runner bundle-policy tests pass. Fresh full assembly measures an 8,716,582-byte Vault CLI total, 1,599,840-byte runner entry, 7,754,155-byte static boot closure, and 9,676,426-byte runner total without a ratchet change. Docs drift, whitespace, and candidate identifier/secret scans pass.
- Round-14 exact-byte correction proof: the two focused direct/generated-reference regressions pass; the complete group-tool, outbox, service, response-media, and both image-generation files pass 303 tests; Assistant Engine typecheck passes. Runner CLI/entry bundle policy passes 50 tests and deploy-preflight coverage passes 85 tests. After the final conflict-free main merge, fresh full assembly measures an 8,717,119-byte Vault CLI total, 1,619,381-byte runner entry, 7,778,630-byte static boot closure, and 9,711,325-byte runner total against the retained 9,711,424-byte ceiling. Whitespace and candidate identifier scans pass.
- Final post-PASS base reconciliation proof: the six affected Assistant Engine files pass 304 tests; Assistant Engine typecheck passes; the combined changelog page/registry set passes 41 tests; runner CLI/entry bundle policy passes 50 tests; deploy-preflight coverage passes 85 tests. Fresh full assembly measures an 8,815,310-byte Vault CLI total, 1,619,381-byte runner entry, 7,845,802-byte static boot closure, and 9,808,583-byte runner total against the ratcheted 9,841,351-byte ceiling. Whitespace and candidate identifier scans pass.
- Exact-head CI repair proof: the failing assistant coverage log identified only two stale prompt-header assertions; the focused `model-behavior` file reproduced 2 failures before the test-only correction and passes all 73 tests afterward.
- Merge-boundary changelog-fragment proof: fragment generation passes, the fragment/page/registry suites pass all 47 tests, and the full Web typecheck passes after its required Health Commons, Prisma, changelog, and route-stub generators. The committed tree is conflict-free and clean.

No environment, Worker binding, Web API, or SQL/application-database migration changes. This PR does advance the persisted Durable Object runner schema floor from version 15 to version 16. The transcript status marker uses the existing transcript schema: older containers ignore it, while newer containers read it only as provenance. The optional provider-effect fact is written and consumed inside the same hosted-runner artifact; new readers accept all older records, and records without the fact fail closed only for ambiguous multi-message media projection.

This is a Cloudflare/runner hard-floor deploy. Deploy with `container_rollout=immediate` and require managed-container smoke to report the exact new bundle fingerprint before admitting image or voice delivery. `RunnerStateStore` version 16 is established before runner or workspace access, so version 15 cannot read or quarantine a marked record. After a member's Durable Object reaches version 16, do not roll Worker or runner below that floor; recover with a forward fix on version 16 or newer. No tandem Web deploy or compatibility subsystem is required.


rocketman-21:

Original requirement: after Murph delivers a generated image, a later explicit native reply must identify and reuse that exact capture for the existing group-avatar action. Native provider resume is an optimization; the committed transcript and sent outbox must preserve exact identity without granting effect authority.

Shape comparison:
- First-reviewed head `ba55cd37241d`: authored source +223/-25 (248 lines churn).
- Current reviewed head `e453a66203`: authored source +581/-53 (634 lines churn).
- Round-1 remediation delta through `1ac8b9c497`: authored source +372/-42 (414 lines churn).
- The remediation did not cross the 500-added-source trigger, but round 3 and the repeated continuity mechanism independently require this retrospective.

Repeated mechanism and root cause: the accepted round-1 correction correctly chose the existing transcript marker plus sent outbox as the ownership boundary, but it used `delivery.message === null` as a proxy for media identity. Production Linq image delivery appends accessibility alt text and persists that text in `providerMessageEffects`, so the exact generated-image branch was unreachable on the ordinary path. The regression test copied the same false assumption by using empty messages and no provider effects instead of tracing the real Linq send shape.

Requirement reset:
1. One provider message may truthfully carry both visible text and exact media provenance.
2. Text or accessibility fallback must never erase the exact media selected by a native reply.
3. The canonical mapping remains replied-to provider message id -> exact sent outbox delivery -> exact vault image -> matching runtime transcript marker.
4. The marker remains provenance only. The later accepted input, existing group preflight, vault reload/validation, and publication remain the only mutation authorities.

Decision: continue with the smallest owner-correct fix. Delete message nullability as the discriminator, derive generated-image reply context from the exact sent media plus matching marker regardless of accompanying text, and preserve visible text as context when present. Do not add a queue, database, protocol, index, lifecycle, or authority owner. Reverting or splitting would leave the original exact-reply requirement unresolved, while the correction fits the owners already introduced for that requirement.

Required proof before round 3: exercise the production Linq send path so its real `providerMessageEffects` feed the outbox shape; send two generated captures with the same normal/default alt; native-reply to the first provider message; prove only the first ref/hash reaches planning; and retain the completion-only no-effect-authority, later accepted-input, preflight, vault-byte, and publication assertions. Re-audit the full original requirement against this mapping before the next pushed review round.

rocketman-21:
Round 3 retrospective (repeated exact-media continuity mechanism)

The accepted round-3 finding exposed the remaining requirement gap: the prior reset separated visible text from media, but still treated one outbox intent as one physical reply target. Linq can split one intent into independently replyable primary and rich-link provider messages, so intent-level media cannot be copied to every provider effect.

Shape: the round-3 remediation is +214/-60 overall and +42/-19 authored source. It adds no persistence schema, service, queue, or authority owner.

Root cause and escape: coverage proved two separate image intents and ordinary text/link splits independently, but did not compose image media with a one-intent/two-effect physical split. The resolver also reordered the legacy scalar ahead of the persisted ordered id list, hiding the only existing media-owner signal.

Reset invariant:
1. One outbox intent may produce multiple independently replyable provider messages.
2. For current Linq split sends, only index zero of the persisted ordered provider ids owns the intent media.
3. A matched sibling effect must carry no parent media and a null-text sibling must receive neutral context, never a generated-image or generic-media claim.
4. Multi-effect legacy records without ordered ids fail closed rather than guessing from the scalar id.
5. The later accepted input, group preflight, vault validation, publication, and provider mutation remain the only effect authorities.

Architecture decision: keep the existing outbox, provider-effect, transcript-marker, and reply-resolver owners. Use the already persisted id order locally; do not add a schema, database field, protocol, or compatibility service.

Required proof now landed: one real vault image plus terminal link traverses attachment reservation/upload, primary send, link send, effect persistence, outbox reload, and replies to both physical ids. The primary exposes exact ref/hash and bounded text; the link sibling exposes neither image identity nor a media claim. Created-chat coverage proves the same ordered physical shape. The correction also stayed inside the existing vault CLI bundle budget.

rocketman-21:
ReviewGPT round 4 retrospective (physical media ownership)

Original requirement: after Murph delivers a generated image, a later explicit native reply must identify and reuse that exact capture for the existing group-avatar action. Native provider resume is an optimization; the committed transcript and sent outbox must preserve exact identity without granting effect authority.

Shape comparison:
- First-reviewed head `ba55cd37241d`: authored source +223/-25 (248 lines churn).
- Current reviewed head `ed2f90680c`: authored source +636/-69 (705 lines churn).
- Relative to the first-reviewed shape, review remediation has grown source additions by 413 and total authored-source churn by 457. The next substantive run is round 5, so the later-round trigger independently requires this retrospective.

Repeated mechanism and root cause: round 3 correctly proved that one intent can produce independently replyable physical messages, but its correction treated ordered provider-message position as a generic media-owner fact. That is true for Linq image/text plus rich-link delivery, but false for successful text-plus-voice delivery, where the voice media is intentionally the second provider message. Logical intent media and physical-message ownership are different facts; ordering alone cannot derive the latter across delivery shapes.

Reset invariant:
1. The physical sender that knows which provider message actually carried the intent media must attest that fact on the existing provider-message effect.
2. A true-only media-owner fact preserves intent media only for that exact matched effect; sibling effects receive no parent media.
3. A single-provider-message legacy media delivery remains unambiguous.
4. A multi-message legacy delivery without explicit ownership evidence fails closed for media instead of guessing from either the first or final provider id.
5. Generated-image transcript markers remain provenance only. A later accepted input, group preflight, exact vault validation, publication, and provider mutation remain the only effect authorities.

Decision: continue with the smallest owner-correct redesign. Delete the positional ownership heuristic; add one optional true-only fact to the existing provider-effect record; set it where the Linq primary body includes media and where the dedicated voice endpoint succeeds; and resolve exact media from that attestation. This adds no database migration, queue, service, state machine, compatibility subsystem, or authority owner. Reverting or splitting would leave the original exact-reply requirement unresolved, while this correction fixes the abstraction at the sender boundary already responsible for physical provider effects.

Required proof before round 5: run a production-shaped successful text-plus-voice journey through the real channel adapter, outbox persistence/reload, and explicit-reply resolver. Replying to the text id must quote only text; replying to the voice id must receive the attested media/no-text context. Retain voice-only single-id proof, split voice-fallback sibling neutrality, generated-image existing-chat primary/link ownership, created-chat image/link ordering, focused typechecks, exact hosted-local assembly, and all exact-head CI.

rocketman-21:
## ReviewGPT round 5 retrospective

Accepted finding: the round-4 media-owner correction writes `carriesIntentMedia` inside a strict persisted provider-effect object. The immediately previous runner rejects that unknown nested key, and a normal vault-backed outbox read quarantines the entire sent intent. Idle snapshot maintenance can trigger that read without a member interaction, and the portable snapshot excludes the quarantine subtree, so a rollback followed by a forward upgrade can permanently lose the provider-message-to-media binding.

How this escaped the prior correction: the parent review proved the forward writer, current reader, delivery equality, generated artifact, exact assembly, and physical image/link plus text/voice behavior. It also disclosed that an older strict reader would not understand the fact, but treated that as a temporary fail-closed projection caveat. It did not trace the exact previous parser through vault-backed quarantine, idle residue inventory, and portable snapshot exclusion. The current-version regression suite therefore proved the new semantic invariant without proving the persisted-state rollback invariant.

Root cause: a new writer was enabled in the same release that introduced its reader, even though the field sits below a strict cross-version persistence boundary. The schema location turns an ordinary emergency rollback into destructive record quarantine rather than an ignorable or retained unknown field.

Corrected invariant: a persisted runtime writer may activate only after every supported rollback target can parse and retain its records. If an existing permissive boundary cannot carry the fact without changing old semantics, deployment must establish and enforce a reader-capable rollback floor before writer activation. A PR-body warning is not an enforcement mechanism.

Remediation direction: reuse the existing hosted runner rollout and fingerprint/deployment owners to make the accepting reader the enforced floor before any media-owner writer can activate. Add focused compatibility proof for marked image and voice intents through the supported floor reader, idle snapshot maintenance, portable archive, restore, forward read, and exact sibling projections. Do not add a compatibility service, queue, or second persistence owner.

rocketman-21:
Round 6 retrospective (recorded before remediation)

The finding is accepted. The generated-image completion added exact physical-message ownership for fully sent multi-effect deliveries, but the review-induced ownership field was not propagated through Linq’s existing accepted-primary/rich-link-sibling failure checkpoint. That checkpoint remains retryable with `deliveryConfirmationPending: false`; because exact reply lookup currently admits only `sent` intents, a native reply to the already-visible image before the link retry loses an otherwise exact provider-message-to-media join.

Why the prior rounds missed it: the regressions covered successful image+link and text+voice deliveries plus retry convergence, but did not cross the two mechanisms in one production-shaped journey: accepted primary media, failed terminal link, persisted/reloaded retryable checkpoint, and an inbound reply before `nextAttemptAt`. The Round 4 explicit-effect correction narrowed the evidence contract without auditing every partial-delivery serialization path that carries provider IDs.

Correction discipline for round 7: preserve the existing primary provider effect on the partial-delivery error and reconstructed delivery; admit only a narrowly attested Linq retryable checkpoint (`deliveryConfirmationPending: false` plus an exact matching `carriesIntentMedia: true` effect), never generic retryable intents or positional inference. Add an endpoint-shaped persisted/reloaded regression with a second generated image, provider-resume rejection, pre-retry native reply, and same-idempotency convergence proof. Update the PR affected-surfaces disclosure, then run a fresh full-patch ReviewGPT round because this PR is sensitive and large.

rocketman-21:
## ReviewGPT round 7 retrospective (accepted physical-delivery lifecycle)

### Original requirement

After Murph generates and visibly delivers an image, a later accepted member request must be able to bind and reuse that exact capture through the existing authorized group-avatar path. Generation history and provider continuity are provenance only; the foreground input, group preflight, vault validation, private publication, backend access check, and Linq mutation remain the effect owners.

### First-reviewed shape versus current shape

The first-reviewed head relied on native provider continuity plus a bounded runtime-authored transcript marker for fresh-history recovery. Review rounds then separated visible text from media identity, separated one logical intent into independently replyable physical effects, replaced positional ownership with the sender-authored `carriesIntentMedia` fact, established the schema-v16 rollback floor, preserved that fact through accepted-primary rich-link partial delivery, and required delivered evidence before a generated capture can be reused as an avatar.

The current shared delivery-evidence predicate correctly accepts `sent` delivery and the narrowly attested non-confirmation-pending Linq `retryable` checkpoint. It still couples retained physical acceptance evidence to that one logical retry status. The existing outbox owner preserves the accepted provider ID and exact marked effect when a retry becomes `sending` and when a later terminal failure persists `failed`, but both the explicit native-reply reader and the generated-ref avatar boundary discard it solely because the logical status changed.

### Repeated mechanism and decision

This repeats the round-6 mechanism: a provider-accepted physical effect becomes visible and interactive before the logical multi-message intent is terminally successful. The prior correction covered only the pre-retry `retryable` moment; it did not cover the same evidence across the full existing partial-delivery lifecycle.

Continue this PR with the smallest owner-boundary correction. Reuse the persisted outbox delivery and sender-authored effect; add no state owner, schema, queue, lifecycle, reconciliation pass, compatibility path, or positional inference. The shared evidence predicate will admit non-sent evidence only for the existing `retryable`, `sending`, or `failed` states when confirmation is not pending, the channel is Linq, and exactly one accepted provider effect carries the intent media. `pending`, `awaiting_approval`, `abandoned`, malformed, duplicate-owner, unknown-ID, unmarked, confirmation-pending, and adjacent sibling records remain excluded. Generic outbox readers remain sent-only; only the explicit reply path and generated-image delivery gate opt into this evidence.

Focused proof will reproduce the production ordering where an accepted-primary checkpoint reaches its retry time, the retry becomes terminal before the same pass scans the accepted native reply, and fresh planning still receives only that image's exact ref/hash with another generated image present. Direct generated-ref avatar reuse will use the same shared predicate, and negative coverage will prove the excluded states and ambiguous effect shapes remain fail-closed.

rocketman-21:
## Seven-round hard-cap retrospective and continuation decision

Original requirement: after Murph visibly delivers a generated image, a later accepted member request must be able to identify and reuse only that exact capture through the existing authorized group-avatar path. Provider continuity and transcript history remain provenance; accepted input, group preflight, vault validation, private publication, backend access, and Linq mutation remain the effect owners.

Shape comparison:

- First-reviewed head `ba55cd37241d`: authored source +223/-25 (248 lines churn).
- Current behavior head `73c5453bab7f`: authored source +799/-150 (949 lines churn).
- Review remediation therefore added 576 source additions and 125 source deletions relative to the first-reviewed shape. The growth is concentrated in preserving exact physical-effect identity across split delivery, persistence, retry, rollback-floor, and explicit-reply boundaries, plus production-shaped regressions; it did not add a queue, state machine, reconciliation service, database owner, or compatibility subsystem.

Why continuation is justified: reverting would restore the original lost generated-image continuity or reintroduce a proven authority/delivery defect. Splitting the now-coupled provider-effect, outbox, reply, and group-boundary corrections would leave an intermediate branch that cannot safely satisfy the user-visible journey. The current correction uses the existing transcript, outbox, provider-effect, vault, group-preflight, publication, and deployment-floor owners; the round-7 remediation broadened only an opt-in evidence predicate across the existing `retryable`, `sending`, and `failed` lifecycle when exactly one Linq effect is physically accepted. Generic readers remain sent-only.

Gate state before continuation: the preliminary specialist findings are resolved; the required final-gate lane means no separate local deep-review applies; focused local proof, package typecheck, bundle policy, exact runner assembly, privacy/static checks, and parent source review are clean; exact-head GitHub CI is fully green. The round-7 repeated-mechanism retrospective was recorded before remediation.

Decision: **continue** with substantive round 8. This explicit continuation was requested in the active task after the round-7 fix. Preserve the immutable first-reviewed baseline and require a later exact-current-patch `ROUND_OUTCOME: PASS` before the PR can become ready.

Tooling note: the first round-8 invocation was invalid and does not advance the counter. A second local session advanced the docs-only active-plan head while the packager was building the prior behavior head, so the ZIP combined the old `pr.diff` with the new plan file and correctly failed closed. Repackage the stable pushed head and retry substantive round 8.

rocketman-21:
## ReviewGPT round 8 retrospective — markerless trusted completion

The finding is accepted as `REVIEW_INDUCED`.

Original requirement: provider continuity and transcript history may retain an exact generated capture, but a later explicit group-avatar request may reuse it only after the existing outbox proves that exact media was visibly delivered.

Root cause: the delivery-eligibility correction needs to preserve the ordinary `raw/captures/**` path, so it distinguishes generated captures by the runtime-authored transcript marker. The marker writer, however, was conditioned on the exact media appearing in an actual response segment. A trusted ready completion can legally call `finish_without_reply` or otherwise omit the attachment while still persisting provider resume continuity and the exact saved ref. In that state the capture is generated but markerless, so the verifier's ordinary-capture fallback returns true without any outbox evidence. The correction therefore conflated generated provenance with successful delivery at the writer, then treated missing provenance as proof that the delivery guard did not apply.

This repeats the prior visibility mechanism but does not require another owner. The hosted completion restriction already carries the trusted exact ref/hash/type/size, the transcript already owns data-only provenance, and the outbox already owns physical visibility.

Decision: continue with the smallest owner-boundary correction. Persist the existing data-only generated marker for every trusted ready completion that carries exact canonical capture media. When that exact media was actually attached, retain its real delivery ordinal; otherwise use the completion turn's response ordinal. The marker continues to authorize nothing: absent exact matching outbox evidence, generated-avatar reuse fails before private publication or mutation. Ordinary canonical captures that never came through a trusted generated completion remain markerless and keep the existing path. Add no queue, state machine, schema, index, lifecycle, reconciliation pass, or alternate delivery truth.

Required proof before round 9:

- A trusted ready completion with a non-null provider thread id calls `finish_without_reply`, persists resume continuity and the exact generated marker, but has no media outbox intent.
- Resolving that retained ref against the persisted marker and empty outbox returns false.
- The existing group effect boundary still stops false delivery evidence before private publication or mutation.
- Exact sent and narrowly attested accepted Linq primary delivery remain eligible.
- An unrelated ordinary `raw/captures/**` ref with no generated marker remains eligible.
- Mismatched response media cannot turn the marker into delivery evidence because the outbox join still requires the exact ref/hash/type/size and turn.

Shape before this correction remains authored source +799/-150 versus +223/-25 at the immutable first-reviewed head. The correction changes one existing marker owner and focused regressions only.

rocketman-21:
Round 9 repeated-mechanism retrospective

Original invariant: a generated image may become group-action input only after the existing delivery owner proves that the exact generated media was made visible in chat. Ordinary member-supplied capture references remain eligible without generated-delivery provenance.

Round 8 fixed the case where a trusted completion restriction existed but a ready completion produced no attachment: the finalizer now records generated provenance even for the no-reply outcome.

Round 9 exposed the same provenance invariant at two earlier/later boundaries. A trusted completion batched with a later accepted input is currently discarded before the effect restriction is constructed, so an in-flight group call has no generated-origin fact. Separately, transcript retention can remove an older delivery marker while the existing generated-capture lookup still recognizes the reference. Both paths let a generated reference be mistaken for an ordinary capture.

Decision: keep the existing completion restriction/capture lookup as generated-origin evidence and the outbox as the sole delivery owner. Add no queue, database, state machine, or alternate delivery truth. The smallest correction will carry one trusted completion restriction through a compound completion turn and make group verification fail closed whenever the requested reference is known-generated from the in-flight restriction or the existing capture lookup but lacks exact outbox-backed delivery proof. Ordinary capture references keep their current path. Focused failing tests will precede the production change, and the resulting exact head will receive a fresh full-patch ReviewGPT round.

rocketman-21:
## ReviewGPT round 10 repeated-mechanism retrospective

The original requirement is unchanged: a generated image may be reused for a group avatar only after the existing delivery owner proves that the exact immutable media became visible in the same assistant session. A trusted completion may share a turn with the later foreground request, but it must never inherit the foreground request's unrelated action authority.

The immutable first-reviewed head contained authored source +223/-25. Before this correction, the current pushed head contains authored source +869/-152. Review-driven growth separated provider continuity from delivery identity, attached media identity to the exact accepted physical effect, established the schema-v16 rollback floor, and enforced exact generated-origin plus outbox-backed visibility at the group mutation boundary. It did not add a queue, state machine, reconciliation service, database owner, or alternate delivery truth.

Round 9 corrected a completion-first compound batch, but left cardinality ambiguous at the same boundary: the restriction builder returns no restriction for multiple trusted completions, while the existing selector and adjacency grouping can place multiple completions in one batch. The result fails open into the later foreground input's generic group-action scope. Round 10 also found that a retained no-attachment provenance marker tied exact delivery evidence to the original completion turn even when a later turn in the same session visibly delivered the same exact immutable media.

Decision: continue with deletion and owner-boundary tightening only. Isolate adjacent trusted completions in the existing grouping predicate, while preserving one completion plus its later same-route foreground input. For delivery provenance, remove only the original-turn equality from the existing marker-to-outbox join; retain same-session and exact ref/hash/type/size matching plus accepted delivery evidence. Add no multi-completion restriction shape, queue, manager, lifecycle, compatibility path, new state owner, or alternate fallback. Focused regressions cover two adjacent completions followed by foreground input, later-turn delivery of the exact marked media, and same-session/hash negatives. The corrected exact head will receive a fresh full-patch ReviewGPT round.

rocketman-21:
Round 11 recovery note: the captured response does not count as a substantive outcome because it emitted the unsupported token ROUND_OUTCOME: FAIL, but its single diagnostic finding is accepted after parent code-path proof. A current trusted image-completion scope blocks Murph dynamic effects while the provider attempt still inherits the foreground Codex native sandbox and feature set. The correction stays inside the existing Codex turn runner: use the existing read-only sandbox and native-capabilities-disabled thread config only while that completion scope is current, retain native resume plus the existing hosted tool context/materializer and exact-media or physical-note dynamic paths, and restore ordinary native capabilities only for a current accepted foreground input. This adds no owner, queue, state machine, lifecycle, or compatibility path. Focused provider-boundary proof will assert the restricted completion attempt and the later foreground restoration. The next valid audit remains substantive round 11 because the malformed outcome cannot advance the counter.

rocketman-21:
## ReviewGPT round 11 repeated-mechanism retrospective

The original requirement is unchanged: trusted image completion may resume the foreground provider history so the generated capture remains available, but the completion itself may produce only no reply, its exact validated media attachment, or the already-owned exact-origin physical-note continuation. Provider continuity must not become generic effect authority.

The immutable first-reviewed head contained authored source +223/-25. Before this correction, the current pushed head contains authored source +874/-152. Review-driven growth separated provider continuity from dynamic-tool authority, exact physical delivery, generated-origin provenance, retry lifecycle, and rollback-floor safety. It reused existing transcript, outbox, provider-effect, grouping, and tool owners and added no queue, reconciliation service, database owner, session manager, or compatibility subsystem.

Round 10 closed a cardinality hole inside the dynamic-tool restriction path, but round 11 exposed the same authority invariant at the parallel native-provider boundary. `hostedImageCompletionEffectRestriction` reaches the provider execution input and runtime-denies Murph dynamic RPCs, yet the native runner gate recognizes only output-only and creative-notification profiles. A trusted completion therefore retains the hosted route sandbox plus native shell, file-change, app, browser, web, plugin, and multi-agent surfaces. Native resume also does not resend per-thread config, so a thread-only restriction would be insufficient.

Decision: continue by tightening the existing provider-attempt boundary only. Derive native-host restriction from the existing trusted-completion restriction, reuse the canonical read-only sandbox and native-capability disable configuration, and apply equivalent process config overrides so resumed threads are covered. Remove completion access to provider/public fetch adapters and injected tool environments. Preserve native thread resume, the existing hosted tool context, workspace materialization, dynamic tool catalog, private attachment delivery, and the current exact-media/no-reply/physical-note runtime checks. Add no new authority contract, profile, state owner, queue, lifecycle, manager, or output-only compatibility branch. Focused provider-boundary proof will fail before the correction and then prove native effects are disabled while exact hosted completion paths remain wired; the corrected exact head will receive another sensitive full-patch ReviewGPT round.

rocketman-21:
## Round 11 repeated-mechanism retrospective

The valid round-11 full audit accepted one review-induced authority finding.
The completion-native restriction is fixed at provider launch, but the live
hosted effect scope can clear after a later foreground input is admitted through
`turn/steer`. The current test proved restoration only by starting a second
provider invocation, so it did not exercise the immutable configuration of the
already-running restricted attempt.

Corrected invariant: an accepted foreground input must not be admitted into a
provider attempt whose immutable sandbox, process features, environments, and
fetch adapters still belong to a trusted completion. A foreground input already
present before launch may keep the existing fold and start under ordinary
configuration. Once a restricted completion attempt has launched, later input
must remain with the existing mailbox owner and execute exactly once in the next
unrestricted turn.

The smallest correction stays at the provider-attempt boundary: suppress live
provider steering for a completion-native-restricted attempt. The existing
mailbox/admission owner retains the input and schedules the following ordinary
turn. This rejects dynamic capability mutation, a new queue, state machine,
authority profile, lifecycle manager, compatibility branch, or alternate input
owner.

Round 12 will add a production-shaped late-steering regression, clarify that the
PR changes the persisted runner schema floor without adding an SQL/application
database migration, include the existing redacted desktop/mobile changelog
captures in the guarded package, and review the fresh full patch. The decision
is to continue because the correction closes a proven accepted-input invariant
inside the existing owners and splitting would leave the reviewed native
restriction incomplete.

rocketman-21:
## Round 12 repeated-mechanism retrospective

Round 12 found one remaining visibility-boundary gap: the direct
`image_ref` avatar route verifies a generated capture before private
publication or group mutation, while the sibling `avatarSource: "generate"`
route could consume generated `referenceImageRefs` before that same check.

### Corrected invariant

Every `raw/captures/**` input that can influence a group-avatar publication
must pass the existing generated-image delivery verifier before image
generation, private publication, or group mutation. Origin classification may
come from the retained marker, current completion restriction, or generated
capture lookup, but only accepted physical outbox evidence can authorize reuse
of a generated capture. A mixed reference set fails closed if any generated
input is not visible.

### Smallest correction

Keep the existing group-tool effect boundary and verifier as the sole owner.
At that boundary, collect the direct `imageRef` or all generated
`referenceImageRefs`, deduplicate them, and run the existing verifier for every
candidate before preparing the runtime request. Reuse the existing visible-
first failure response if any candidate is rejected.

Focused proof will cover an unseen generated reference blocking the image
provider, private publisher, and group mutation; the same request succeeding
after exact accepted delivery; a mixed set failing closed; and an ordinary
non-generated capture remaining eligible.

### Rejected complexity

No new queue, state owner, authority profile, verifier, database, lifecycle,
compatibility branch, or dynamic capability mechanism is warranted. The gap is
one sibling branch bypassing an existing pre-publication guard.

rocketman-21:
## Round 13 repeated-mechanism retrospective

Round 13 found that the new visible-first verifier correctly requires outbox
delivery evidence, but the existing terminal-outbox count cap can delete that
evidence while the generated capture is still live under its 14-day retention
window. A visibly delivered image can therefore become falsely ineligible after
100 newer terminal intents.

### Corrected invariant

Generated-image delivery evidence must remain available for the same bounded
period in which the retained generated capture can drive explicit reuse. Count-
based inventory control may prune ordinary terminal intents, but it must not
shorten the existing age-based authority window for exact generated-image
delivery evidence. The transcript marker, provider continuity, and capture
lookup remain provenance only and never substitute for accepted delivery.

### Smallest correction

Keep the existing outbox store and maintenance owner. Derive protected delivery
evidence directly from the intent's existing singleton `gpt-image-2`
`vault_image` under `raw/captures/**`; exclude only those intents from the
100-record count eviction while retaining the existing 14-day terminal-age
cutoff. Ordinary terminal inventory remains capped at 100.

Focused proof will persist accepted generated delivery plus more than 100 newer
ordinary terminal intents, run the real prune path, and verify delivery remains
eligible before 14 days. It will then advance beyond the age cutoff and prove
the generated delivery is pruned and no longer authorizes reuse.

### Rejected complexity

No new state, lookup, queue, lifecycle, database, delivery owner, or alternate
authority source is warranted. The correction is one derived exception inside
the existing outbox pruning owner.

rocketman-21:
## Round 14 repeated-mechanism retrospective

Round 14 found that the visible-first verifier joined accepted delivery to a
canonical capture ref before either avatar route loaded the bytes it would
actually consume. Because a later foreground turn can legitimately write
within `raw/captures/**`, the same ref could retain delivery evidence for image
A while resolving to different image B bytes at publication or provider time.

### Corrected invariant

Every generated capture that can influence a group avatar must be resolved
once at its consuming boundary, and that exact resolved descriptor—ref,
SHA-256, media type, and byte length—must match accepted physical outbox
delivery evidence before the same byte snapshot reaches the image provider or
private publisher. Origin classification may still come from the transcript
marker, current completion restriction, or generated-capture lookup. Ordinary
captures absent from those provenance sources keep their existing path. A
newly generated avatar output is authorized by its explicit generation action
and is not mistaken for a reused source reference.

### Smallest correction

Delete the ref-only pre-load loop. Extend the existing hosted delivery verifier
to receive the already resolved descriptor and compare it with the existing
outbox media fields. The direct route verifies after its one source read and
publishes those same bytes. The generated-reference route verifies immediately
after its existing reference resolution and before its existing provider
request. No new hash, persistence, lookup, queue, lifecycle, authority, or
reconciliation owner is needed.

Focused proof will visibly deliver image A, overwrite the same ref with valid
image B, and prove both direct publication and generated-reference provider
egress stop before any effect. It will also prove unchanged A succeeds and an
ordinary canonical capture remains eligible.

### Decision

Continue with the byte-bound correction. Reverting would restore the original
lost generated-image journey, while splitting at this point would leave either
visible-first delivery or exact-byte reuse incomplete. Broad filesystem
restrictions, immutable capture storage, a second delivery ledger, or a new
authorization service are rejected as unnecessary complexity.

PR #1597, opened 2026-08-10
Run device sync from system mailbox mode

rocketman-21:
## Why

Device-sync recovery work can arrive through hosted runtime system-mailbox state, but `system_mailbox` invocations previously only imported and projected it. Deterministic Junction work could remain behind a model-capable wake and require an assistant-capable pass.

## User-visible Goal And UX

When device sync needs recovery, the hosted assistant catches up without admitting an assistant turn or invoking Codex. A system-mode child runs at most one deterministic device-sync item, checkpoints successful canonical updates, then yields so fresh foreground input can proceed. It never consumes operator maintenance receipts.

Immediate feedback is unchanged because this is background recovery. Follow-up and retry wakes remain typed and are rederived after each checkpoint.

## ReviewGPT Preliminary Lenses

- Product-experience lens: applicable. Background recovery progresses without assistant admission, while foreground work wakes the system child and proceeds after its bounded checkpoint.
- Prompt lens: not applicable. No assistant/provider prompt behavior changes.
- Frontend lens: not applicable. The existing changelog renderer receives one registry entry; no component, layout, state, or interaction changed.
- Coverage lens: applicable. Regression coverage exercises import-and-run, stale projection rederivation, retry, dirty acknowledgement, continuation wake, abort propagation, foreground handoff, canonical publication, and maintenance-owner isolation.

## ReviewGPT Final Context

ReviewGPT context sensitivity: sensitive

Reason: hosted runtime wake ownership, canonical checkpoint ordering, and system-mailbox execution affect production reliability.

ReviewGPT first-reviewed head: 5ae11d90cc3d1b346a5c4727dc0647d1646b80fa

ReviewGPT current candidate head: 3597c083887152470240b1f071ff5dcbca71a12c

[...]

PR #1656, opened 2026-08-11
Fix newsletter integration regressions

rocketman-21:
## Why this PR exists

The generic group-email cutover exposed two failures in the production-faithful public/private integration matrix. A generated-image completion could deliver its text without the requested attachment, and a Junction replay assertion could fail after later device-import logs rotated the earlier system-mailbox receipt out of the bounded status window.

Generated-image completions keep their exact trusted response media even when model-relayed capture metadata is stale. Wearable replay verification continues to require a clean, durably settled system-mailbox lane without depending on an ephemeral diagnostic entry still being among the latest 50 logs.

## Root cause

- The trusted hosted-image completion gate compared the model-relayed vault descriptor before the runtime rehydrated its ref from the vault. A stale hash was rejected before the existing canonical read and full exact-media comparison, so the model continued with text alone.
- The device-sync scenario treated a recent per-invocation counter or log line as the only receipt evidence. New import timing logs can legitimately evict that line, but a merely settled system lane could also describe older work. The durable fallback therefore needs a pre-trigger frontier to prove the tested replay advanced the lane.

## Invariants

- A hosted image-completion turn may request only the exact trusted vault ref.
- The runtime rehydrates that ref and requires the complete trusted media descriptor to match before creating a response patch; a different ref, altered authored metadata, or changed bytes still fails closed.
- A device-sync receipt is accepted only when there is explicit receipt evidence or the system imported watermark advanced from the pre-trigger baseline, is handled through exactly, and has zero lag.
- Retryable or record failures still fail the wearable assertion.

## Architecture and reuse

- Existing systems reused: canonical vault-image rehydration, the existing full exact-media comparison, and the system mailbox's durable imported/handled-through watermarks.
- New logic: the early completion authority gate admits only the exact trusted ref and delegates mutable metadata proof to the canonical vault read; the wearable proof captures the system frontier before each trigger and requires the post-trigger frontier to advance and settle fully when bounded receipt logs have rotated.
- New abstractions: none are needed because both corrections stay within the existing assistant-engine and mailbox-test ownership boundaries.
- Complexity intentionally avoided: no new runtime owner, compatibility path, status retention, or diagnostic-log dependency.

## ReviewGPT later-round context

- Classification reason: trusted private-image effect authority and durable mailbox settlement evidence are security/reliability boundaries even though the correction is small.

ReviewGPT first-reviewed head: 3bf3124ca84dbdbae13f17f144a6cd648279107b

ReviewGPT current candidate head: 575aaf34c71f252600ee14dfa1f52c27e819b853

- Final round 1 passed at `3bf3124ca84dbdbae13f17f144a6cd648279107b` with no findings.
- The independent completion-specialists pass found that a missing or malformed pre-trigger imported watermark was coerced to `"0"`, which could manufacture evidence of advancement. The accepted remediation now requires an explicit non-negative decimal baseline, proves the failure cases directly, and makes the high-volume replay require the durable frontier rather than a bounded log fallback.
- Final round 2 passed the exact remediation head with no findings; model verification confirmed GPT-5.6 Pro.
- Post-review integration triage proved that the route-authority test observed completion on the human owner even though the durable route binds work to its synthetic group container. The test now resolves the canonical route and waits on `containerMemberId`; this is the same narrow correction already reviewed independently in public PR #1597.
- The mandatory round-3 retrospective records an explicitly justified continuation: production shape is unchanged from round 1, both review-driven corrections are test-only proof required by this paired cutover gate, and no new production owner or lifecycle machinery is permitted.
- Final round 3 passed this exact head with no findings. It reverified generated-image authority, durable wearable-frontier proof, the canonical synthetic-container route observer, and the retrospective scope boundary; response-model evidence resolved to GPT-5.6 Pro.

## Review lenses

- Product experience: applicable — preserves generated-image attachment delivery.
- Prompt: not applicable.
- Frontend: not applicable.
- Coverage: applicable — existing scripted runtime, media, group-email terminal, and combined hosted-local scenarios exercise the paths.

## Candidate shape

| --- | ---: | ---: |
| Source | 12 | 4 |
| Tests / fixtures | 191 | 5 |
| Docs | 0 | 0 |
| Config / tooling | 0 | 0 |
| Generated / other | 0 | 0 |
| **Total** | **203** | **9** |

## Changelog

- Changelog: not applicable
- Reason: This restores pre-cutover response-media delivery before the generic group-email rollout and introduces no new member-facing behavior.

## Verification

- Hosted image-completion authority and group-email terminal runtime: 10 tests passed.
- Full hosted-local Codex image-media delivery: 3 tests passed, including background completion and saved-image reuse.
- Runner bundle/package build passed as part of the hosted-local scenario.
- Cloudflare package typecheck passed.
- Focused hosted-system mailbox frontier proof: 4 tests passed, including missing, malformed, negative, and valid baselines.
- Full local Linq group-route drift E2E passed against the paired private Temporal worker: 2 tests passed in 51 seconds, including the corrected synthetic-container completion proof.
- Exact current-head public CI passed at `575aaf34c71f252600ee14dfa1f52c27e819b853`.
- Exact current-head public/private integration passed all 12 hosted-local scenario groups against public `575aaf34c71f252600ee14dfa1f52c27e819b853` and private `d4ff0dd77fd0ffdaece12f86b97bd6771451b185`, including group-email newsletter, route authority, generated-media/provider, trigger-bound wearable frontier, restart/retry, and checkpoint durability ([run](https://github.com/cobuildwithus/murph-cloud/actions/runs/31489870482)).
- `git diff --check`

The exact combined public/private hosted-local workflow is the production-faithful remote gate for both reported failures.

### DIFF
--- a/apps/cloudflare/test/hosted-local-device-sync-junction-wearable-direct-resource-replay-e2e.test.ts
+++ b/apps/cloudflare/test/hosted-local-device-sync-junction-wearable-direct-resource-replay-e2e.test.ts
@@ -100,6 +100,26 @@ let scenario: HostedLocalFullStackScenario | null = null;
 let linqStub: HostedLocalLinqStub | null = null;
 let browserVaultSummary: JunctionWearableBrowserVaultReplicaSummary | null = null;
 
+describe("hosted system mailbox frontier proof", () => {
+  it.each([
+    null,
+    "not-a-sequence",
+    "-1",
+  ])("rejects a %s pre-trigger imported sequence", (value) => {
+    expect(() => requireNonNegativeDecimalSequence(
+      value,
+      "pre-trigger hosted system mailbox imported sequence",
+    )).toThrow("pre-trigger hosted system mailbox imported sequence");
+  });
+
+  it("accepts an explicit non-negative pre-trigger imported sequence", () => {
+    expect(requireNonNegativeDecimalSequence(
+      "42",
+      "pre-trigger hosted system mailbox imported sequence",
+    )).toBe("42");
+  });
+});
+
 describe("hosted local Junction wearable direct-resource replay e2e", () => {
   beforeAll(async () => {
     plan = await buildJunctionWearableHostedReplayPlan({ replaySize: "smoke" });
@@ -475,6 +495,7 @@ describe("hosted local Junction wearable direct-resource replay e2e", () => {
     }
 
     await assertHostedDeviceSyncReplayReceiptAccepted({
+      requireAdvancedDurableFrontier: true,
       scenario: activeScenario,
       systemImportedSeqBefore: systemImportedSeqBeforeDeviceSync,
       userId,
@@ -1232,6 +1253,7 @@ async function readBrowserVaultReplica(input: {
 }
 
 async function assertHostedDeviceSyncReplayReceiptAccepted(input: {
+  requireAdvancedDurableFrontier?: boolean;
   scenario: HostedLocalFullStackScenario;
   systemImportedSeqBefore: string;
   userId: string;
@@ -1278,7 +1300,10 @@ async function assertHostedDeviceSyncReplayReceiptAccepted(input: {
       lane.lane === "system" && lane.lag === "0"
     );
   const receiptObserved = durableSystemLaneAdvancedAndSettled
-    || recordedDeviceSyncLog !== undefined;
+    || (
+      input.requireAdvancedDurableFrontier !== true
+      && recordedDeviceSyncLog !== undefined
+    );
 
   if (
     !receiptObserved
@@ -1299,6 +1324,8 @@ async function assertHostedDeviceSyncReplayReceiptAccepted(input: {
         systemHandledThroughSeq,
         systemImportedSeq,
         systemImportedSeqBefore: input.systemImportedSeqBefore,
+        requireAdvancedDurableFrontier:
+          input.requireAdvancedDurableFrontier === true,
       })}`,
       ...(safeErrors.length > 0
         ? [`system mailbox safe errors: ${JSON.stringify(safeErrors)}`]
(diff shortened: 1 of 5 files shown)


rocketman-21:
## ReviewGPT round 3 retrospective

**Original requirement.** Preserve the exact trusted generated-image attachment when only model-relayed derived metadata is stale, while continuing to reject changed image authority; and prove high-volume wearable replay from durable system-mailbox advancement instead of bounded diagnostic-log retention. This PR is the integration-regression follow-up required to make the newsletter runtime/skill cutover provable as one paired candidate.

**First-reviewed versus current shape.** The first-reviewed head, `3bf3124ca84dbdbae13f17f144a6cd648279107b`, contained the production image-authority correction plus its focused regression coverage and the initial wearable replay proof. The current head, `575aaf34c71f252600ee14dfa1f52c27e819b853`, is 203 additions and 9 deletions across four files. Production-source shape is unchanged since the first review: 12 additions and 4 deletions in the existing dynamic-tool admission gate. No production owner, state machine, queue, lease, fence, lifecycle, compatibility path, dependency, or public abstraction was added.

**Review-driven growth.** Since the first-reviewed head, remediation added 54 lines and deleted 6 lines, all in existing integration tests:

- 45 additions and 5 deletions make the wearable replay proof fail closed when its pre-trigger durable frontier is missing, malformed, or negative. This resolves the accepted specialist finding that `?? "0"` could fabricate a baseline and turn an absent prerequisite into apparent advancement.
- 9 additions and 1 deletion make the route-authority test read the canonical durable group route, assert its human-owner binding, and observe execution on `route.containerMemberId`. The prior observer incorrectly waited on the human owner even though canonical routing executes the group automation on its synthetic container.

These are distinct test-proof defects, not a repeated production mechanism.

**Decision: explicitly justified continuation.** Keep the unchanged production design and both test-only corrections in this PR. The wearable correction is necessary for the durable-frontier requirement to be truthful. The route correction is necessary for the exact public/private combined gate to observe the runtime member selected by the product's canonical route owner. Splitting either proof would leave this cutover PR dependent on a known-invalid aggregate test or an unproven replay invariant without reducing production complexity.

**Locked scope boundary.** Continue only with the existing production image-authority correction and the test-only proof corrections above. Add no production owner, compensating lifecycle machinery, compatibility path, or new runtime abstraction. Any further finding must be tied to the original authority or durable-proof requirements; otherwise shrink, split, or reject it as out of scope. The current exact head is the candidate for ordinary round 3 review.

PR #1688, opened 2026-08-11
Fix pending group health sharing

rocketman-21:
## Why this PR exists

A newly approved group health share could immediately be reported as having no usable records because the read boundary collapsed an unmaterialized first snapshot and a completed empty snapshot into the same state. Projection processing also waited behind unrelated join-confirmation recovery, the first fast-wake revision had no durable recovery owner, explicit reaffirmation could reuse a materialized older date window, and maintenance could be consumed before projection actually succeeded.

Once a member approves an exact group share, Murph starts preparing that share immediately and recoverably. Until the first authorized snapshot is ready, the group sees a truthful preparing state instead of a false absence, repeat permission pressure, a zero score, or a device diagnosis.

## User experience

The entry point remains the existing server-owned reaction or first-party permission page. Approval commits the exact grants and the existing durable maintenance control row in the same database transaction, then signals that exact mailbox checkpoint concurrently with confirmation recovery. Caller cancellation cannot suppress the committed work. Explicitly reaffirming a recent-data permission starts a fresh null-snapshot generation immediately, so the group sees pending rather than any older materialized window. If the first Temporal signal fails, the existing bounded scheduled mailbox-handoff sweep re-signals the unconsumed row. After import, the row stays retained through source checkpoint and projection; a projection failure keeps the same recording item retryable and holds the mailbox watermark behind it.

A shared read during that asynchronous window says the approved recent data is still preparing and invites a later retry; it does not promise a completion time. Pending participants stay unranked, are not treated as zero or disconnected, and do not receive another permission request. The final data remains visible only through a later authorized group read. No unrelated outbound message or new production screen was added.

Every touched consent surface now matches the producer: sleep timing, sleep duration, selector-based activity distance, and selector-based session count disclose the current member-local date and prior six civil dates, while the private projection owner emits at most seven authorized date records.

- Web remains the sole grant authority; the private member runtime remains the sole projection producer.
- Pending reveals only that an active readable exact-scope grant lacks its first snapshot. It reveals no health value, source, private sync state, or suspended-access reason.
- Encrypted empty snapshots remain missing; inactive-access withholding remains opaque missing; revoked or absent grants remain not granted.
- Consent and durable projection-work admission commit atomically; bounded post-commit signaling is only a latency hint and scheduled recovery owns missed signals.
- No touched consent surface authorizes more than the seven-date producer bound.
- A stale producer cannot write a later grant generation. Active-scope discovery returns a fixed-width opaque digest of the active exact-scope row set; Web re-derives it before replacement. A mismatch remains generically retryable while an exact granted null snapshot exists and becomes ordinary no-active only after revocation or materialization, so records read before consent rotation cannot enter the new generation. Reaffirmation and the bounded legacy rollout rotate recent-date generations before sharing resumes.
- A group-share maintenance row remains unhandled until every current granted runtime-projectable generation is materialized or revoked. Temporarily inactive, capability-deferred, generation-raced, missing-port, and projection-error outcomes remain retryable; only one opaque deferred-work bit crosses into the private runtime.

- Hosted Web-to-runner wire parsing: the additive pending value is validated and invalid status/record combinations still fail closed.
- Destination authority: projection discovery and delivery now resolve owner-backed and current-participant-backed active group runtimes through the same bounded set resolver before hashing or replacement, while each replacement transaction still revalidates live authority.
- Assistant model input: the group prompt and model adapter preserve four states and prohibit ranking, private-source inference, or re-consent pressure for pending participants.
- Temporal/mailbox orchestration: grants and the existing durable `runtime.maintenance-requested` row enter one transaction. Group-share maintenance creates an additive post-checkpoint projection record, remains unhandled through source checkpoint and delivery, and uses the existing recording retry path on failure. Post-commit signaling uses the exact saved checkpoint, and the existing scheduled handoff query includes unconsumed maintenance rows without increasing its candidate or concurrency bounds.
- Private projection runtime: recent sleep and daily owners derive the current civil date from the validated member-vault timezone, reject future and pre-window dates, and cap the authorized date set at seven records.
- Native and Web consent copy: exact seven-day windows are disclosed for sleep and selector-based activity scopes. The real Web permission-card component is captured in the design catalog at desktop and mobile viewports.

## Architecture and reuse

- **Existing systems reused:** The grant transaction, nullable first-snapshot marker, encrypted empty snapshot representation, maintenance control mailbox, scheduled mailbox-handoff sweep, Temporal processing adapter, private projection worker, exact-scope shared read, and Web permission-card component remain the owners.
- **New logic:** Web maps a readable null snapshot to pending, rotates explicitly reaffirmed recent-date permissions to fresh generations, atomically admits projection work with the grant, signals the exact checkpoint after commit, includes maintenance rows in the existing recovery sweep, carries pending through the parser/model guidance, narrows recent projections to the member-local seven-date boundary, and provides a bounded count-only rollout command for materialized legacy group generations. The runtime retains group-share maintenance through checkpoint and projection, delivers currently active scopes immediately, and consumes the obligation only when every approved runtime-projectable generation is materialized or revoked. Durable first materialization is source-filtered by the existing null snapshot cursor and divided into deterministic complete-scope pages totaling at most 25 destinations.
- **New abstractions:** One extracted transaction helper shares the existing generation-stable maintenance append between interactive acceptance and the finite rollout command. One additive local post-checkpoint record routes group-share maintenance through the existing system-mailbox recording retry lifecycle. One fixed-width deferred-work bit, one exact runner-capability query parameter, and one acknowledged ephemeral first-materialization mode preserve that lifecycle across temporarily inactive grantors or destinations, bounded continuation, and mixed-version rollout. The existing null snapshot is the only progress cursor; none creates persisted product state or a new owner.
- **Complexity intentionally avoided:** No new queue, scheduler, retry manager, status column, database migration, second projection owner, cached fallback, side-effectful read, or proactive group message was introduced.

## ReviewGPT findings and remediation

- Preliminary specialists and final round 1 independently found that seven-day consent copy could authorize an eighth record and that a one-shot payload-free wake could leave pending without a recovery owner. Both findings were accepted.
- Final round 2 found that the UTC cutoff still violated the member-local promise at timezone extremes and that post-commit maintenance admission retained a grant-without-work crash window. Both findings were accepted.
- The producer now derives the current civil date from the validated member-vault timezone and permits only that date plus the prior six dates. Approval now admits the grant and its stable-generation maintenance row atomically, then signals the exact checkpoint as a best-effort latency hint; the existing bounded sweep owns missed signals.
- Final round 3 found that explicit reaffirmation could reuse a materialized pre-fix recent-date generation and that maintenance was removed before the later best-effort projection offer succeeded. Both findings were accepted. Reaffirmation now rotates and clears materialized recent-date grants in the acceptance transaction; the runtime retains only generation-stable group-share maintenance through checkpoint and projection, retries failures on the existing recording path, and advances the system watermark only after terminal completion.
- A bounded, dry-run-by-default rollout command rotates already-materialized pre-cutoff group health generations and appends their durable maintenance row in the same transaction. It outputs counts only, uses a stable cutoff for idempotency, and excludes current-state, email, device-status, non-group, and post-cutoff grants; orphaned pending recent-date generations whose original wake is absent are included.
- Final round 4 found that a not-yet-due projection retry could block unrelated runtime controls and that backfilling before the pending-aware Web reader would recreate the false-missing UX. Both findings were accepted. Projection work now has a derived local serialization key within the existing system mailbox, and failure recomputes the earliest wake so account disconnect and other runtime controls proceed while projection FIFO and the unresolved watermark are preserved. The rollout now converges Runtime first, deploys the complete pending-aware Web release second, and only then runs the stable-cutoff backfill.
- Round 4 also identified a discovery-to-delivery consent race. The runtime now returns Web's opaque active-generation digest with each offer; Web compares it to the current active destination set before any replacement. A rotated generation produces a generic retryable result while exact null work remains and a fresh durable retry without exposing raw share IDs or recipient cardinality to the private runtime.
- Final round 5 found that tokenless legacy delivery could still bypass that fence during mixed-version operation. The finding was accepted. Generation proof is now mandatory for every delivery: new Runtime fails before reading private vault data when old Web omits proof, and new Web returns a retryable error before share lookup or replacement when old Runtime omits proof. The token-capable Runtime fleet is the hard rollback floor before Web promotion.
- The specialist frontend lens found both rendered permission-card states legible, responsive, and coherent. It returned no separate frontend finding.
- Final round 6 found that discovery and delivery used an owner-only access approximation, so an approved share to a participant-backed active group could remain pending after the durable obligation was consumed. The finding was accepted. Both seams now use one bounded participant-aware owner-or-current-participant resolver, hash the identical active destination set, and retain the final transaction-local authority check. Exact-scope reads reuse the existing 25-destination admission bound; all-scope reads compose it with the finite projection registry; current-participant reads compose it with the 32-person roster bound; every overflow fails closed.
- Final round 7 found that temporary destination inactivity could still consume the only durable obligation. The finding was accepted. Web now derives one opaque deferred-work bit from granted runtime-projectable null snapshots omitted by current access or runner capability. Runtime delivers active scopes immediately but retains the same mailbox item on a five-minute deferred retry. Exact generation mismatches and mid-fan-out access changes return generic retryable failures while null approved work remains. A previous runner without the exact capability receives retryable discovery failure during convergence instead of consuming the row. Round 7 returned findings, not PASS; the user explicitly approved continuation to round 8.
- Final round 8 found that targetless maintenance could replay every materialized scope and destination before foreground work regained control. The finding was accepted. Maintenance now requests an acknowledged first-materialization mode; Web source-selects only null snapshots, pages complete exact scopes to at most 25 destinations, and delivery conditionally replaces only exact rows still null. Materialized rows never enter the page or generation digest. A page that made progress continues after one second, while inaccessible-only work retains the five-minute retry, leaving foreground and independent controls runnable between passes.
- Final round 9 found that temporary grantor inactivity could acknowledge first materialization as terminal and consume the only durable obligation while approved snapshots remained null. The finding was accepted. An inactive grantor now returns the acknowledged mode plus the same opaque deferred-work bit without reading grants, destinations, or private vault data. Deferred-capable Runtime retains the existing item and five-minute retry; an older Runtime receives the existing retryable capability error. When authoritative access returns, the retry materializes current null grants or terminates normally if they were revoked. Generic idle refresh remains terminal.

## Round-three architecture retrospective

Decision: justified continuation without a new persisted owner. The first-reviewed head changed 25 files / 512 lines; the current head changes 49 files / 2,549 lines. The review-driven growth closes four reproduced privacy/reliability gaps, adds the finite legacy rollout required to make the consent correction universal, and is primarily executable proof: 1,141 added test lines versus 805 added source lines.

- Member-local current date plus prior six civil dates remains the universal existing “last 7 days” contract for all recent date-based projection families. Per-scope exceptions would preserve the same disclosure bug elsewhere; the separate global-close workout projection remains unchanged.
- Atomic admission stays in the existing group acceptance transaction coordinator, which already owns membership, exact grants, and confirmation admission. The low-level grant store remains mailbox-agnostic; a new service would add an owner.
- Pending and atomic work admission remain indivisible because pending is truthful only when approval durably admits the work that resolves it. The date correction stays in the same PR because splitting would temporarily make consent and producer bounds disagree.
- Retained concepts and owners: ephemeral grant-store requiresProjection; ephemeral group-acceptance checkpoint DTO; local generation-derived idempotency fingerprint; existing exact-checkpoint signal API; existing bounded sweep predicate; one assistant-runtime vault date context; the derived pending wire/model value; one shared transaction append helper; and one additive local post-checkpoint record carried by the existing system-mailbox state. No new status column, database migration, queue, scheduler, or persisted product-state owner exists.

The full retrospective and concept-by-concept disposition are recorded in the PR discussion.

## Round-five architecture retrospective

Decision: justified continuation with no new persisted owner. Review-driven source growth since the immutable first-reviewed head is above the automatic retrospective threshold, but the round-4 and round-5 remediation stays inside existing owners: one derived in-memory serialization key in the system-mailbox selector, one stateless mandatory digest in the existing active-scope/delivery wire, the existing row-generation IDs as the authority, and the existing bounded backfill. No queue, status column, feature flag, compatibility database state, or second retry lifecycle was added. The deployment correction also deletes the need for a partial Web compatibility release: converge Runtime, deploy complete Web, then backfill and drain. The generation fence is indivisible from that order because ordinary approval, reaffirmation, and the backfill can rotate consent while runtimes may be reading.

## Round-seven architecture retrospective

Decision: justified remediation with no new persisted owner. The accepted failure was reproduced in the existing terminality path: access-safe omission plus terminal `no-active-share` removed the only mailbox item while the Web-owned snapshot remained null. The correction adds one derived boolean to the existing active-scope response, one exact capability parameter for mixed-version safety, one bounded null-snapshot existence query only on delivery absence or token mismatch, and one outcome on the existing system-mailbox recording state. Active destinations still materialize in the same pass; permanently inactive destinations use a five-minute retry rather than a hot loop. No queue, scheduler, table, status column, recipient identifier, fan-out count, or second retry owner was added.

## Round-eight architecture retrospective

Decision: justified correction by narrowing existing work, with no new persisted owner. The accepted path replayed all materialized rows because a targetless durable record reused the generic all-active refresh. The correction gives that existing record one ephemeral acknowledged mode, uses the existing null snapshot as its progress cursor at the source query, selects deterministic complete-scope pages totaling at most 25 destinations, and uses the existing recording retry time as the page continuation. Delivery adds the same mode only to require an exact null-row compare-and-set. Generic abortable idle refresh remains unchanged. No cursor table, queue, scheduler, manager, or lifecycle owner was added.

## Round-nine architecture retrospective

Decision: justified symmetric correction with no new persisted owner. The inactive-grantor branch now uses the same existing deferred bit and recording retry already required for inactive destinations. It performs only the grantor access check, returns no scopes, and reads neither grant rows nor private vault state. The existing five-minute recording retry remains the sole recovery owner; restored authoritative discovery either resumes null-only materialization or proves revocation and terminates. Generic idle refresh is unchanged. No queue, wake producer, state field, status column, or second lifecycle was added.

## Changelog

- Changelog: updated
- Items: 2026-08-11 · group-sharing-starts-after-approval

## Hot reply path impact

Not applicable. This changes the post-commit tail of an explicit reaction or link approval, not durable acceptance of an ordinary current conversation message through provider start and durable reply handoff.

## Database collection fanout impact

The existing scheduled mailbox-handoff query widens its runtime-control predicate inside the same bounded candidate set: it retains `LIMIT handoffLimit + 1` (26 by default, 251 at the hard maximum), chooses at most one row per user, and keeps five-way signal concurrency. Approval adds the existing workspace target check/upsert and one mailbox append inside the already-short database-only grant transaction; exact-pointer signaling remains after commit.

A first-materialization discovery query now selects only granted null snapshots at the source. Its fail-closed read admits at most 2,500 rows: 25 destinations multiplied by the current 100-scope finite registry, with `take=2,501` detecting overflow. The current runtime-projected subset is 98 scopes, so its composed maximum is 2,450 null rows. Access resolution then uses one set-based member query and at most one current-participant query; at the absolute 2,500-container candidate maximum, the existing 32-participant roster bound permits at most 80,000 participant rows and reads 80,001 only to fail closed. A redundant maintenance pass after all snapshots materialize selects zero rows and performs no member or participant access query.

From that bounded control read, Web chooses complete exact-scope generations totaling at most 25 active null destinations. One durable pass therefore performs at most 25 private scope reads, 25 Runtime-to-Web delivery requests, and 25 sequential replacement transactions. Across those delivery requests the exact-scope candidate total is still at most 25; each request enforces the existing 26-row fail-closed read, and their participant-access bounds compose to at most 825 returned-or-overflow-probe rows. Each replacement transaction performs only the two canonical member access validations, encryption, and one exact null-row conditional update; no provider call, sleep, or other network work occurs inside a transaction. Inactive or unsupported null rows consume no page capacity and retain the opaque deferred bit. An inactive grantor first-materialization request performs the existing grantor access check, then reads zero grants and zero destinations and retains the same five-minute retry.

The finite rollout fetches at most 101 candidate grantors and processes at most 100 per invocation (26 and 25 by default), uses one connection, runs grantors sequentially, and caps each short database-only transaction at 25 qualifying grant rotations plus one mailbox append; signaling remains after commit. Delivery absence or generation mismatch adds at most one exact-scope `findFirst` only while checking for remaining granted null work. No per-group network fanout occurs inside a transaction.

Complete first-provider request capture compared base `a4b76776c11dfe1ca045bd40bdacdeea600510df` with candidate `1ca9ee3ae5160f62954ff746dce2c283bbf2d5c1` using the pinned real Codex App Server, the repository scripted Responses endpoint, `gpt-5.6-terra`, low reasoning, production code mode, identical synthetic direct/group requests, and `gpt-tokenizer` 3.4.0 `o200k_harmony`. The branch was subsequently rebased once. The only later prompt correction removed 18 instruction bytes while preserving the pending rules, so it cannot increase the measured byte footprint; the exact-head resident prompt-ratchet test passes. The later ReviewGPT remediation changes projection/orchestration code and consent copy only; it does not enter the initial provider request. The synthetic design-catalog study also does not enter provider input.

- Individual Murph: 23,841 tokens / 108,952 bytes at base; 23,837 / 108,961 at candidate; -4 tokens (-0.0168%) and +9 bytes (+0.0083%).
- Group Murph: 20,277 tokens / 93,171 bytes at base; 20,348 / 93,549 at candidate; +71 tokens (+0.3502%) and +378 bytes (+0.4057%).
- Attribution: the entire delta is assembled instruction input from the group-state and freshness wording. Tool schemas, generated guidance, and all other included fields are unchanged. Temporary capture code and payloads were removed.

- Product experience: applicable. The pass found and drove the exact-window and durable-continuation remediations.
- Prompt: applicable. Resident group guidance adds pending-specific response, retry, scoring, diagnosis, and repeat-consent rules; the prompt-budget ratchet passes.
- Frontend: applicable. The real production selector permission cards are rendered with synthetic props and captured at both required viewports; the specialist found no visual defect.
- Coverage: applicable. Earlier focused proof includes 235 Web behavior tests, 107 projection-owner tests, 154 group-tool tests, 66 hosted parser tests, 104 Assistant Engine tests, and the exact 73-test prompt-ratchet file. Round-three remediation additionally passes 135 focused Web acceptance/reaction/grant/backfill tests, 378 runtime mailbox/workspace tests, 81 hosted-execution contract tests, 37 changelog tests, all affected typechecks, scoped lint, docs drift, and privacy/diff checks. Round-six remediation passes 137 focused Web tests, 9 Hosted Execution tests, both affected typechecks, scenario-manifest integrity, scoped lint, and diff checks. Round-seven remediation passes 441 focused Hosted Execution, Assistant Runtime, Web, and Cloudflare tests; all four package typechecks; the 204-scenario integrity guard; scoped Web lint; and diff/privacy checks. Round-eight remediation passes 451 focused tests across the same four owners, all four package typechecks, the 204-scenario integrity guard, docs drift, scoped Web lint, and diff/privacy checks. Its direct proofs cover a 2,450-row maximum materialized fan-out yielding only the one null cursor row, inactive-destination reactivation, complete-scope page bounds, redundant-maintenance exclusion, mode acknowledgment, prompt continuation versus deferred retry, and foreground admission between pages. Round-nine remediation passes 455 focused tests across the same four owners, all four package typechecks, the 204-scenario integrity guard, docs drift, scoped Web lint, and diff/privacy checks. It directly proves inactive-grantor opaque deferral without share/private reads, old-runtime failure, retained item and watermark, input-free retry, revocation termination after access restoration, and unchanged generic idle terminality.

Health-sharing persisted state, privacy semantics, cross-owner wire compatibility, projection disclosure bounds, and post-commit ordering make this a sensitive change.

## Change-shape breakdown

Classification: production TypeScript, including the synthetic design study, is source; Vitest files are tests/fixtures; durable owner docs, public changelog content, the index, execution plan, and public-safe Frog entry are docs; the package script entry and bounded operator script are config/tooling; no generated files changed. Binary files: none.

| --- | ---: | ---: |
| Source | 1,326 | 226 |
| Tests / fixtures | 2,791 | 171 |
| Docs | 523 | 33 |
| Config / tooling | 147 | 0 |
| Generated / other | 0 | 0 |
| **Total** | **4,787** | **430** |

## Design proof

- Catalog surface: [Group join permission studies](https://app.withmurph.ai/design?tab=sections#group-join), state `group-join-selector-permissions`
- Desktop screenshot: Desktop selector-based group-share consent cards
- Mobile screenshot: Mobile selector-based group-share consent cards
- Local and delivered captures were inspected at native resolution; ordinary body copy is immediately legible and no member data appears.
- Claude Code UI double-check: attempted after the surface stabilized; Fable reported explicit usage-credit exhaustion, so the repository-required gap is recorded and no alternate Claude request was made.

## Deployment concerns

1. Deploy the Cloudflare consumer/parser and runner bundle first with immediate container rollout. It must include the additive pending parser/model status, member-local seven-date projection bound, retryable post-checkpoint consumer, projection-only mailbox serialization, opaque generation-token client, exact deferred-work capability, and first-materialization mode acknowledgment check. Prove fleet and warm-container convergence before continuing.
2. Verify every live Runtime active-scope request receives a fixed-width generation token and every delivery sends it. Declare that converged Runtime bundle the hard rollback floor. Then deploy the complete Web release, including the pending-aware reader, exact seven-day consent copy, atomic grant/work admission, reaffirmation rotation, null-only bounded page selector, and mandatory generation-token delivery boundary. Verify a tokenless synthetic delivery is retryable with no share lookup or replacement, an active null snapshot reads as pending, a first-materialization request receives the exact mode acknowledgment and no materialized row, a deferred-capable runner receives the opaque deferred bit, an older runner receives generic retryable discovery while deferred work exists, and inactive-grantor first materialization retains the item without share discovery before a restored-access retry completes or terminates. This deployment precedes every backfill write.
3. Capture one stable UTC cutoff. Under the production Web environment, run the count-only dry run and then bounded apply batches with that exact cutoff until `selectedGrantors=0` and `hasMore=false`; wait for the durable maintenance backlog to drain. Do not change the cutoff or skip the drain proof. As soon as the complete Web writer begins accepting approvals, the pending-aware Web reader is the Web rollback floor; forward-fix Web rather than rolling it back to false-missing semantics.

ReviewGPT first-reviewed head: 52e37b84bd2eb1b58b9d0d033ab4b4750c4a35c6

ReviewGPT first-review change shape: source +105/-46; tests/fixtures +126/-36; docs +174/-25; config/tooling +0/-0; generated/other +0/-0.

Current candidate head: 59d87cfa3778e0c3f5bc2c27c7a33a1b66456618


rocketman-21:
## Round-three architecture retrospective

**Original requirement.** Once an exact group share is approved, preparation should begin immediately and the room should see a truthful pending state until the first authorized snapshot exists. Approval must not create a durable grant that can remain pending forever, and the disclosed seven-day boundary must match what every affected producer can emit.

**Shape comparison.** The first-reviewed head changed 25 files / 512 lines (source +105/-46, tests +126/-36, docs +174/-25). The current head changes 34 files / 1,609 lines (source +379/-153, tests +675/-97, docs +273/-32). The 1,097-line increase is review-driven: round 1 required durable recovery and exact seven-date production; round 2 proved that UTC dates violated member-local consent at timezone extremes and that post-commit admission retained a grant-without-work crash window. Most growth is boundary/failure coverage across the existing projection families and grant/reaction paths; authored-source churn is 532 lines, below the large-feature threshold.

**Decision: justified continuation, with no further machinery.**

- The member-local current date plus prior six civil dates is the universal contract for recent date-based share projections, not only the four newly surfaced consent options. Existing policy copy already says “last 7 days” for daily metrics, workout summaries, activity, heart-rate zones, sleep, and meal totals. A shared UTC cutoff could disclose an eighth local date or omit the oldest authorized local date anywhere that common policy is used. Keeping one vault-timezone date context in the existing assistant-runtime projection owner is smaller and safer than per-scope exceptions. The separate `workouts.v0` global close rule remains unchanged.
- The group acceptance store remains the application transaction coordinator: it already owns membership acceptance, exact-scope grants, and join-confirmation admission. It constructs one existing `runtime.maintenance-requested` event only after knowing which grant generations actually need projection, and appends it through the existing mailbox owner in the same database transaction. Moving this into the low-level grant store would couple that owner to mailbox protocol and emit one row per scope; adding a new service would add an owner without reducing coupling. The current shape keeps the grant store mailbox-agnostic.
- Pending readiness and atomic projection admission are indivisible for safe shipping: pending is only truthful if approval durably admits the work that resolves it. The member-local date correction is logically separable code, but not safely separable from this consent change: splitting permits a deployment where exact seven-day copy and producer output disagree and creates a cross-PR tandem dependency. Keeping the consumer/parser, producer bound, and Web approval boundary together preserves one explicit consumer-first deployment contract.

**Review-added concept inventory.**

- Keep `requiresProjection`: an ephemeral grant result owned by the grant store; it prevents wakes for ready snapshots and identifies null/regranted generations. It is not persisted state.
- Keep the projection-maintenance signal DTO: an ephemeral post-commit handoff owned by group acceptance, parallel to the existing confirmation signal. It exposes only the exact saved mailbox pointer needed by the existing signal API.
- Keep deterministic grant-set fingerprinting: a local idempotency key derived from existing grant-generation IDs, so request retries deduplicate while a regrant creates new work. Random identity would admit redundant maintenance rows.
- Keep exact-checkpoint signaling: an existing orchestration capability that avoids a second lookup and minimizes approval-to-preparation latency. The signal remains best effort.
- Keep maintenance-row sweep inclusion: a widened predicate in the existing bounded sweep, not a new queue, scheduler, query pass, or reconciliation owner.
- Keep shared vault date context: one validated timezone read in the existing projection owner, enforcing the universal recent-date policy consistently.
- Keep pending wire/model state: derived from the existing nullable encrypted snapshot; no status column, migration, or new state owner.

Nothing is added to production state beyond the existing grant and mailbox row, and no reviewed concept can be deleted without reopening one of the reproduced failures. Splitting or extracting further would add deployment or ownership seams rather than simplify the architecture.

PR #1864, opened 2026-08-15
fix(hosted): import initial activation before conversation turn

rocketman-21:
## Why

The first hosted owner can begin from a conversation signal while the initial `member.activated` continuation is present in the same immutable mailbox prefetch. Two special-case paths moved activation outside the existing workspace-runner boundary: the dirty live-wake path replaced the conversation import, and the cold path combined activation with the uncaught initial conversation import. Either could suppress the exact fresh-input batch. Canonical checkpoints could also publish a stale conversation cursor after local mailbox state had advanced.

## User-visible goal

Let a member message Murph immediately after joining and receive the normal first reply without leaving account setup work or mailbox progress behind.

## Product journey

- Initiating actor: a newly activated member sending their first conversation message.
- Entry point and timing: the foreground conversation signal can arrive before the activation continuation signal.
- Feedback and destination: the existing assistant reply remains prompt; activation and both mailbox lanes converge in the same live owner.
- Continuation owner: the existing hosted workspace runner and canonical checkpoint writer.
- Permission boundary: unchanged authenticated mailbox and workspace ownership.
- Recovery contract: conversation import retains the exact assistant input; the runner attempts activation before provider planning, and a non-abort activation failure leaves system progress unchanged for the existing mailbox/system retry path.
- Direct evidence: 663 focused runtime tests cover restored first owner, cold bootstrap, dirty live wake, successful and failing activation decode, deep prefetch, transient refetch, no-op activation completion, post-reply maintenance, and canonical cursor projection. The first current-base matrix exposed actor-less direct Telegram routes being rejected by new base route partitioning; 114 focused assistant-engine tests now prove that direct welcomes remain next-turn context while group and unknown routes still fail closed. Exact attached run 31895428855 exercises the corrected head.

## Invariants

- The accepted current conversation remains the initial import and foreground owner.
- The existing workspace runner is the only pre-assistant activation import owner and failure boundary.
- Successful activation is applied before provider planning; non-abort activation failure cannot suppress the conversation and leaves the system watermark retryable.
- Ordinary system work remains behind foreground conversation priority.
- An activation with no room setup or welcome delivery completes during import instead of creating a trailing no-op queue item.
- Canonical mailbox-progress checkpoints project lane cursors from local mailbox state, while a genuine pre-import sequence-zero checkpoint continues to omit the conversation field.
- No new queue, retry owner, scheduler, state store, or dependency is introduced.

## Architecture and reuse

[...]

PR #1886, opened 2026-08-15
fix(assistant): retain direct welcome context

rocketman-21:
## Why

The cross-session route-state change merged immediately before #1864 requires an exact actor partition for non-Linq routes. Direct Telegram conversations intentionally have no actor ID because the direct thread already proves the single participant. As a result, a delivered signup welcome was present in the outbox but excluded from the member's first inbound provider context.

## User-visible goal

Keep an already-delivered welcome available as context when a member sends the first message in the same direct conversation.

## Product journey

- Initiating actor: a newly activated member replying in their direct Telegram conversation.
- Entry point and timing: the welcome is delivered in the activation wake; the member's first inbound arrives in a later wake.
- Feedback and destination: the normal first reply can interpret the welcome that Murph already sent.
- Continuation owner: the existing exact-route state and assistant reply path.
- Permission boundary: missing actor scope is accepted only when the persisted route proves `threadIsDirect === true`.
- Recovery contract: group and unknown routes with no actor remain fail-closed; actor-bearing exact-route digests are unchanged.

## Invariants

- A proven direct thread can be exact without an actor partition.
- Non-direct and unknown routes still require an actor ID.
- Actor-bearing direct and group routes retain their existing exact digest.
- Route-state reads, claims, migration, and retry ownership remain unchanged.
- No new queue, retry owner, state store, service, or dependency is introduced.

## Architecture and reuse

- Existing systems reused: the input and outbox exact-route resolvers, route digest, watermark, claim, and migration machinery remain the sole owners.
- New logic: one shared actor-scope derivation keeps an existing actor value or substitutes the reserved direct-thread scope only when `threadIsDirect === true`.
- New abstractions: one narrow pure helper; no new service, state owner, persisted shape, or dependency.
- Complexity intentionally avoided: no compatibility path, queue, retry loop, route manager, or second source of truth.

- Signup welcome continuity for direct Telegram first contact.
- Actor-less direct email routes can use the same proven-direct rule.
- Mixed actor partitions, groups, and legacy wildcard matches remain isolated.

## Review remediation ledger

- #1864's exact-head matrix exposed the failure after the activation-order correction itself had passed focused review.
- Root cause proof: direct Telegram import and welcome delivery both intentionally carry `actorId: null`; the new exact-route resolver returned no route before reading the direct-thread proof.
- Correction: preserve actor partitions when present and substitute a reserved scope only for a proven direct thread.

## Review lenses

- Product experience: applicable; restores first-contact conversation continuity, with the direct Telegram cross-repository scenario as journey evidence.
- Prompt: not applicable; no prompt or tool-description changes.
- Frontend: not applicable; no user-facing frontend UI changed and no rendered evidence is required.
- Coverage: applicable; 114 focused exact-route and reply-path tests pass, and exact-head CI is in progress.

## Line breakdown

Classification rule: executable production code is source; regression cases are tests; the changelog entry is docs/process; no generated files are included.

| --- | ---: | ---: |
| Source | 22 | 8 |
| Tests/fixtures | 72 | 1 |
| Docs/process | 1 | 1 |
| Config/tooling | 0 | 0 |
| Generated/other | 0 | 0 |
| Total | 95 | 10 |

## Verification

- `pnpm exec vitest run --config vitest.config.ts --no-coverage test/assistant-auto-reply-route-state.test.ts test/assistant-automation-reply-event-path.test.ts` from `packages/assistant-engine` (114 passed)
- `pnpm --filter @murphai/assistant-engine typecheck`
- `pnpm --dir apps/web typecheck:prepared`
- `pnpm exec vitest run --config apps/web/vitest.config.ts --no-coverage apps/web/test/changelog-fragments.test.ts` (7 passed)
- `git diff --check`
- Exact cross-repository correction proof: the direct Telegram scenario passed in run 31895428855; a later full retry is being evaluated separately.

## Changelog

- Changelog: updated
- Items: 2026-08-15 · first-replies-finish-account-setup
- Member outcome: a delivered welcome remains available as direct-conversation context for the first reply.

This PR changes exact-route privacy partitioning, so every substantive final review uses a full snapshot.

ReviewGPT first-reviewed head: 533cd3786608298b3111c53065ef860beae0a1cd

### DIFF
--- a/apps/web/changelog/entries/2026-08-15/first-replies-finish-account-setup.json
+++ b/apps/web/changelog/entries/2026-08-15/first-replies-finish-account-setup.json
@@ -7,7 +7,7 @@
     "priority": 3,
     "title": "First replies now finish account setup cleanly",
     "summary": "When someone messages Murph immediately after joining, Murph can answer without leaving the new conversation's setup work behind.",
-    "details": "This improves the first-conversation handoff while keeping welcome messages and shared-room setup on their existing durable delivery path.",
+    "details": "Welcome messages stay on their durable delivery path and remain available as context for the member's first direct reply, while shared-room setup keeps its existing isolated path.",
     "relevanceTags": ["assistant", "messaging", "onboarding", "reliability"],
     "sourcePullRequests": [1864]
   }
--- a/packages/assistant-engine/src/assistant/automation/cross-session-route-state.ts
+++ b/packages/assistant-engine/src/assistant/automation/cross-session-route-state.ts
@@ -137,9 +137,12 @@ export function resolveAssistantAutoReplyInputExactRoute(input: {
       : null
   }
 
-  const actorId = normalizeNullableString(input.conversation.actorId)
+  const actorScope = resolveAssistantAutoReplyActorScope({
+    actorId: input.conversation.actorId,
+    threadIsDirect: input.conversation.threadIsDirect,
+  })
   const threadId = normalizeNullableString(input.conversation.threadId)
-  if (!actorId || !threadId) {
+  if (!actorScope || !threadId) {
     return null
   }
 
@@ -149,7 +152,7 @@ export function resolveAssistantAutoReplyInputExactRoute(input: {
       ? buildAssistantAutoReplyExactRoute([
           'email',
           identityId,
-          actorId,
+          actorScope,
           threadId,
         ])
       : null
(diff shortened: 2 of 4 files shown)


rocketman-21:
Closing because proactive Telegram signup outreach is intentionally disabled on main, the original route fix already landed through #1890, and the remaining unpublished migration only targets aging historical state.

PR #1981, opened 2026-08-18
Add canonical context to scheduled reminders

rocketman-21:
## Why and outcome

Scheduled reminders previously carried delivery facts but not the canonical records they concerned. A later ordinary direct-chat reply could therefore see a stale active workout without knowing that the new reminder targeted a different saved routine.

This adds bounded generic `contextReferences` to canonical automations and carries them through the existing outbox into scheduled and next-message assistant context. Workout and experiment reminders attach their exact canonical IDs; the assistant still uses the existing domain reads, mutation locks, and commands.

## Product UX

- Effort: Medium product change
- Walkthrough: Ready
- Affected: members replying in ordinary private chat after workout, experiment, or device-activity reminders, including text, text-plus-media, and media-only presentations.
- Recovery: one message may both log the exact set and unmistakably close that session. Merely exhausting a saved plan leaves the workout active for targetless extra sets. A stale different workout requires a member-supplied exact end time or total duration before Murph closes it, starts the referenced routine, and logs the preserved new set.
- Response: ordinary workout mutations return the verified workout card. Missing stale-workout timing gets one text-bearing narrow question that preserves the exact workout-format ID, exact active-workout event ID, proposed routine, exercise, set, and result across committed-transcript replay. Its immediate answer may continue only when exact rereads prove that the same event remains the sole active workout.
- Exclusions: ambiguous routines or sets, multiple active workouts, conflicting references, changed state, or missing finish timing fail closed. Native iMessage Reply is not required, and explicit historical corrections keep their exact targeting.

## Architecture and invariants

- The canonical automation remains the only owner of reminder metadata; no new service, queue, authority object, composite workout command, or reply-attestation mechanism is introduced.
- References are bounded exact IDs for routing and interpretation only. They grant no read or mutation authority and must come from a successful current canonical read or create result.
- Existing one-active-workout and mutation-lock invariants remain the write owners.
- One prior-delivery classifier independently projects quoteable text and canonical references. Cross-session text may be projected; same-session text stays transcript-owned; non-empty references survive either session shape and media-only presentation.
- The exact route digest, causal cutoff, native-reply anchor, claim, and settled-through watermark still order and consume eligible context once.
- Empty references are omitted from persisted outbox JSON.
- Device-activity reminders reuse the existing parent-authority read, authority hash, prompt renderer, outbox fields, and delivery idempotency metadata.
- No database, network, provider, or awaited call was added to the foreground reply path.

ReviewGPT first-reviewed head: a45ae0d4b81a7dcb0350e95015aa66c87a5a6a85

## Review remediation ledger

- Preliminary specialists: accepted compound-transition response guidance, one canonical finish rule, and stronger evidence language for model-authored references. The deterministic owned-boundary suite was retained instead of adding a nondeterministic real-model test.
- Round 1: fixed same-live-session context loss at the existing eligibility boundary and omitted empty persisted fields for old-reader compatibility.
- Round 2: replaced repeated text-only gates with one projection owner so media-only references survive while reference-free media remains excluded.
- Round 3: projected device-activity automation identity and references from the existing verified parent read through the common authority hash, prompt, and outbox owners.
- Round 4: deleted inferred stale-workout end timing; only member-supplied exact end time or total duration can close the earlier workout.
- Round 5: deleted plan-exhaustion closure inference; explicit or unmistakable closure intent remains independent of saved placeholders.
- Round 6: made the exact workout-format proposal transcript-durable for immediate timing answers after native-resume rejection.
- Round 7: added the exact stale active-workout event ID to that proposal and requires exact format, exact event, and sole-active rereads before mutation.

All accepted findings were corrected at existing owners. No service, state owner, compatibility manager, queue, lease, or reconciliation lifecycle was added.

## Evidence

- Full focused implementation suite: 16 Vitest files, 1,209 tests passed.
- Round 2 remediation suite: 5 Vitest files, 317 tests passed.
- Round 3 device-activity suite: 3 Vitest files, 342 tests passed.
- Workout guidance and CLI owner checks: 2 Vitest files, 11 tests passed.
- Changelog fragment, registry, route, and catalog tests: 4 files, 50 tests passed.
- Affected workspace typechecks and root `pnpm typecheck` passed on the reviewed feature head.
- Contract generation, CLI package-shape verification, and `git diff --check` passed.
- After base reconciliation, the generated skill-hash test passed 2 tests, CLI package-shape verification passed, `git diff --check` passed, and `git merge-tree --write-tree` was clean.

## Round-eight continuation

The seven-round cap retrospective concluded that the accepted fixes corrected projection and workout-transition assumptions while retaining the same owners and shrinking unsafe inference. The requester explicitly authorized continuing through conflict resolution and merge on 2026-08-18, satisfying the required continuation decision.

Current `main` was merged once. The only conflict was `packages/cli/src/vault-cli-skill-hash.generated.ts`; the canonical generator derived the combined-tree value. No runtime logic, schema contract, or product behavior was hand-authored in the resolution.

Round 8 returned `ROUND_OUTCOME: PASS` with zero accepted findings after a full audit of `e87f3c1b6afa06a446aecc01d419304cf8f02372` and explicitly verified the round-7 identity correction.

That reviewed head’s exact CI exposed one current-`main` test mismatch: the Web build command had removed its wrapper while the build-order assertion still required the old syntax. The production build path was unchanged; a test-only follow-up now asserts the actual invariant that the WAF preflight is first. The focused file passes all 17 tests. Under the isolated-proof exception, this does not create another substantive ReviewGPT round.

## Deployment concerns

- Deploy the reference-aware Worker and runner together with `container_rollout=immediate` before any reference-bearing reminder can fire, then prove the exact runner-bundle fingerprint.
- The first canonical automation or outbox intent with non-empty references establishes that bundle as the hard rollback floor.
- Rollback below the floor is safe only after current code removes all canonical references and affected intents/checkpoints drain; otherwise forward-fix with the compatible reader. Local CLI downgrade below the same reader floor is unsupported while references remain.

## Changelog

- Changelog: updated
- Items: 2026-08-18 · workout-reminders-keep-the-right-routine


rocketman-21:
### ReviewGPT repeated-mechanism retrospective

Original requirement: every provider-accepted scheduled reminder that carries canonical references must make those references available to the next eligible ordinary direct message, independent of whether the reminder presentation contains text, text plus media, or media only. Native reply is optional.

First-reviewed shape: the outbox reader retained media-only deliveries, but the prior-context resolver and context builder each required quoteable text. The round-1 remediation added a same-session reference exception after the first text gate, so it could not fix reference-bearing media-only reminders. This repeated the same mistake: treating transcript recoverability as eligibility for canonical relationship metadata.

Decision: replace the scattered gates with one projection at the existing prior-delivery selection owner. Quoteable text and canonical references are independent: cross-session deliveries may project text, same-session deliveries do not duplicate transcript text, and either path may project non-empty references. A delivery is eligible only when at least one projected component remains. The context builder will consume that already-classified projection without a second eligibility filter.

Preserved owners: persisted outbox, exact route digest, causal cutoff, native-reply anchor, claim, and settled-through watermark. No service, queue, state machine, new durable field, compatibility path, or reconciliation owner is added. The earlier local same-session exception and both text-only assumptions are deleted/replaced.

Required deterministic proof: same/cross session × text/media-only × absent/non-empty references × ordinary/native reply, plus route ordering and one-time settlement. Reference-free media remains excluded; reference-bearing provider-accepted media reaches exactly one eligible turn.

rocketman-21:
Round 4 anomaly retrospective (before remediation)

- Intended invariant: every supported scheduled-reminder family carries its canonical automation ID and context references into both the scheduled turn and the next ordinary reply context.
- The previously reviewed shape covered canonical time-based jobs. Device-activity automations use a separate local scheduler translation, so their already-verified parent record was not projected into the prompt/outbox context.
- Root cause: the implementation was reviewed by downstream mechanism (prompt/outbox projection) without enumerating every upstream schedule family that reaches it.
- Correction: extend the existing device-activity parent-authority read to return the parent ID and references; reuse the existing prompt renderer, outbox field, authority hash, and delivery idempotency metadata. This adds no new state owner, queue, authority object, or compatibility layer.
- Verification will compose the device-activity owner path end to end: live parent read -> execution prompt/outbox input, reference-sensitive stale-authority rejection, and prior-delivery automation-ID recovery from existing metadata.

rocketman-21:
Round 5 anomaly retrospective (before remediation)

- Original requirement: the reminder must identify its exact automation and canonical workout/experiment so an ordinary reply is not misapplied to yesterday's active session; completing a planned workout should not require a separate ceremonial finish message.
- First-reviewed shape added the generic relationship context. Later remediation also allowed a stale workout to be closed from startedAt plus active durationMinutes.
- Round 4 proved that active durationMinutes is only elapsed time at the latest mutation, not an observed end. Treating it as exact would trade the UX fix for silent canonical timing corruption.
- Decision: delete that derived-end rollover rule, its compound-response exception, and the misleading lifecycle test. Prevent the stale state at its source in prompt ownership: when the member's current set completion fills the final planned placeholder, finish that same workout immediately and return the completed card—no separate finish command. If an already-stale different workout remains and the member has not supplied its end time or total duration, preserve the exact new-set details and ask one narrow timing question before any mutation.
- This reuses the existing set-log, finish, start, card, and mutation-lock owners. It adds no field, command, state owner, scheduler, authority layer, migration, or reconciliation path, and should reduce the PR's authored surface.

rocketman-21:
Round 6 anomaly retrospective (before remediation)

- Original requirement: exact reminder metadata must disambiguate the intended saved routine, and a member should not need a separate ceremonial command when the same message clearly completes both a set and the session.
- Round 4 removed the unsafe stale-workout end-time inference. Round 5 shows that the attempted source prevention—treating the final planned placeholder as session closure—repeats the same mistake with different non-end evidence. A saved plan can be exhausted while the live session continues with targetless extra sets.
- Decision: delete placeholder-count authority. Restore closure intent independent of set count. One message may contain both the exact set result and unmistakable closure language (for example, “last set … I’m done”); Murph then reuses set-log and finish in that same turn and returns the completed card, so no separate finish command is needed. A plain final planned-set result leaves the event active and supports another targetless set.
- Keep the corrected already-stale path unchanged: exact reminder context identifies the new routine/set, but missing end time or total duration causes one narrow no-mutation timing question.
- This is another deletion-only ownership correction. It adds no persisted field, command, branch owner, lifecycle, scheduler, migration, repair path, or compatibility mechanism.

rocketman-21:
Round 7 anomaly retrospective (before remediation)

- Original requirement: the exact automation and canonical record IDs should travel with reminder context so an ordinary reply can act without guessing or native Reply.
- The safe stale-workout path consumes that hidden one-time reminder context while asking for missing end timing. Committed transcript replay preserves the visible question but not hidden turnContext or prior tool arguments, so a supported fresh provider thread could lose the exact routine ID before the member answers.
- Decision: make this one bounded no-mutation question transcript-durable by including the already verified workout_format ID in its text alongside the routine, exercise, set, and result. The immediate timing answer must exact-read that ID and re-read active state; changed, missing, ambiguous, or unrelated continuation fails closed.
- This directly reuses the existing transcript fallback and ordinary domain reads. It adds no context store, watermark, propagation lifecycle, authority object, reply attestation, service, queue, lease, migration, repair, or reconciliation mechanism. Voice may accompany the question, but the exact proposal remains text-bearing for replay.
- Existing generic context transport and one-time settlement stay unchanged.

PR #2004, opened 2026-08-19
test(assistant): cover remapped Linq thread context

rocketman-21:
## Why and outcome

Cover the remaining regression gap identified while comparing the closed outbox-projection work with #1887: a Linq delivery can start with a participant target and later materialize into a different provider thread. This adds real-boundary proof that an unanchored inbound on the materialized thread recovers the prior outbound context through the warm projection, queues the response to that thread, and keeps foreground reads bounded as unrelated outbox history grows.

## Product UX

- Effort: Not applicable — test-only coverage with no member-visible behavior change.
- Result: Not applicable
- Walkthrough: Not applicable — existing Linq behavior is exercised without changing it.

## Evidence

- Direct: `MURPH_TEST_STATE_CARDINALITY=1 NODE_OPTIONS=--max-old-space-size=6144 pnpm --dir packages/assistant-engine exec vitest run --config vitest.config.ts --no-coverage test/assistant-hot-reply-state-cardinality.test.ts --maxWorkers=1` (5 tests passed).
- Coverage: The new probe runs the production queue-only auto-reply boundary at 1, 8, 128, and 256 unrelated outbox intents; it asserts prior context reaches the provider, the queued reply binds to the materialized Linq thread, and the production history reader reports that no canonical outbox scan occurred.
- Type safety: `pnpm --dir packages/assistant-engine typecheck`.

## Deployment concerns

- Deployment: not applicable
- Reason: Test-only change; no runtime or deployment artifact changes.

## Changelog

- Changelog: not applicable
- Reason: Internal regression coverage only; no member-visible change.

### DIFF
--- a/packages/assistant-engine/test/assistant-hot-reply-state-cardinality.test.ts
+++ b/packages/assistant-engine/test/assistant-hot-reply-state-cardinality.test.ts
@@ -605,12 +605,16 @@ async function prepareLinqMaterializedThreadReply(
     async loadOperation() {
       const {
         createAssistantAutoReplyGroupContext,
+        createAssistantAutoReplyHistoryReader,
         processAssistantAutoReplyGroup,
       } = await import('../src/assistant/automation/reply.ts')
       const context = createAssistantAutoReplyGroupContext([contextItem])
       if (!context) {
         throw new Error('Expected one unanchored Linq context.')
       }
+      const historyReader = createAssistantAutoReplyHistoryReader({
+        vault: vaultRoot,
+      })
 
       return async () => {
         boundaries.executeProvider.mockClear()
@@ -626,6 +630,7 @@ async function prepareLinqMaterializedThreadReply(
               userEnvKeys: [],
             },
           },
+          historyReader,
           inboxServices: createInboxServices(),
           requestId: null,
           sessionMaxAgeMs: null,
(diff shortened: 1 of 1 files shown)


rocketman-21:
Coverage specialist disposition: accepted one finding. The test now passes a real production history reader into the reply boundary and asserts `outboxScanPerformed` is false, closing the fixed-cap fallback loophole while retaining the cardinality, prior-context, and materialized-thread assertions. Focused test file: 5/5 passed. Assistant-engine typecheck: passed.

rocketman-21:
CI status:

- The corrected exact head could not start GitHub Actions because each job was rejected with `The job was not started because your account is locked due to a billing issue.` No new-head CI job executed.
- On the prior runnable head, the dedicated foreground state-cardinality job was green. Its only red job was an unrelated current-`main` CLI coverage mismatch: `release-script-coverage-audit.test.ts` expects `manual line-count table`, but the completion-workflow document on `main` no longer contains it. This PR changes neither file.
- After the eight-line coverage correction, the full touched test file (5/5) and assistant-engine typecheck both passed locally.

Please restore Actions billing and rerun the failed workflows. If `main` has not independently repaired the CLI coverage mismatch by then, that base regression will also need resolution outside this Linq-only PR.

PR #2096, opened 2026-08-20
Migrate stable Junction profile timestamps

rocketman-21:
## Why and outcome

Junction profile records written before timestamp stabilization can replay at the same provider revision with `created_at` replacing `updated_at` as canonical occurrence time. The generic conflict guard correctly rejected that shape, so otherwise unchanged stable-ID profiles could not converge.

This adds one exact, one-way stable-ID migration and advances the existing one-shot profile normalization revision from 1 to 2 so already-checked accounts receive one bounded refresh. Real same-revision profile changes remain fail-closed, and member edits, chronology, and deletions are preserved.

Round-2 review found that the account-wide refresh also re-admits supported no-ID profiles whose timestamp-derived identity can move when an earlier `created_at` appears while `updated_at` is unchanged. The accepted retrospective keeps the account-wide refresh because current account metadata cannot distinguish that cohort. Core now binds only the exact equal-revision timestamp replay to its stored provider baseline and retains the existing identity as a canonical no-op. It appends no event-spine revision, preserves member edits and tombstones, and rejects semantic or ambiguous matches before writes. Retrospective: https://github.com/cobuildwithus/murph/pull/2096#issuecomment-5361717138

## Product UX

- Effort: Patch
- Walkthrough: Ready
- Outcome: Eligible older connected-health profiles refresh without losing corrections, restoring deleted facts, or duplicating no-ID facts.
- Reaches: Stable-ID Junction profiles in the released updated-at shape migrate to created-at chronology. Re-admitted no-ID profiles keep their current canonical identity when the provider revision and semantic content are unchanged. Unrelated device facts are unchanged.
- Proof: Provider-shaped integration coverage exercises the real revision-1 scheduled reconcile, stable-ID migration, current-v1 no-ID replay, member edits, member deletion, ambiguity, replay, malformed chronology, and same-revision semantic conflicts.

## Evidence

- Focused stable/no-ID importer proof: 1 file, 8 passed, 240 skipped.
- `pnpm --filter @murphai/importers test`: 19 files, 553 passed.
- Focused equal-revision no-ID ambiguity proof: 1 file, 1 passed, 184 skipped.
- `pnpm --filter @murphai/core test`: 46 files, 810 passed.
- Focused production-shaped revision-2 reconcile proof: 1 file, 1 passed, 321 skipped.
- `pnpm --filter @murphai/device-syncd test`: 49 files, 1,256 passed.
- Core, importer, and device-syncd typechecks passed.
- Changelog coverage: 2 files, 45 passed.
- `pnpm --dir apps/web typecheck`, `pnpm docs:drift`, `git diff --check`, and task privacy scan passed.
- Changelog archive proof: existing 320px and 1280px artifacts show the card without horizontal overflow.

The production-shaped regression starts with revision-1 account metadata and a persisted current-v1 no-ID profile, applies a member height correction and demographics deletion, runs the real bounded scheduled reconcile, proves no event-spine append, advances revision 2 only after import, skips the marked replay, and rejects a same-revision height change atomically.

`pnpm test:diff` passed every affected package typecheck but found two unrelated pre-existing workspace-boundary violations in `packages/cli/test/junction-body-composition-e2e.test.ts` and `packages/vault-usecases/test/junction-workout-features-query.test.ts`. This PR changes neither file; the run was stopped before its broad affected-test phase after the owning-package full suites passed.

Production-reviewed head `478cf6f8712a508676b939b1dcc08a1dcab8a6a1` passed sensitive full-snapshot final round 3. Current PR head `cb46140b0a6b` differs only by the completed-plan archive, which does not require another review round.

- Canonical event-spine reconciliation in `@murphai/core`: migrates the exact stable-ID predecessor and retains exact equal-revision no-ID replays on their existing spine.
- Junction profile normalization: marks explicit and retained stable resource IDs without broadening generic reconciliation.
- Device-sync profile admission: revision 2 invalidates revision-1 profile checks once, then returns to one-shot behavior.
- Public changelog: describes stable-ID recovery and no-ID retention without promising a general identity rewrite.

## Architecture and reuse

- Existing systems reused: Junction normalizer markers, the existing scoped no-ID predecessor index, external-reference version ordering, canonical event-spine reconciliation, member-revision retention, and the existing profile-admission scalar.
- New logic: Adds one shared exact timestamp-replay proof with stable-ID migration and no-ID equal-revision wrappers.
- New abstractions: None; the compatibility rule stays inside the existing canonical reconciliation owner and scoped predecessor index.
- Complexity intentionally avoided: No new database state, queue, service, provider preflight, generic conflict relaxation, dual write, repair scan, or second migration state machine. The existing strictly newer no-ID identity migration remains the only adoption path.

## Hot reply path impact

Not applicable. Device profile import/reconciliation is outside the Foreground Reply Critical Path and adds no reply-path database, network, or provider calls.

- Murph: Not applicable; no prompt, instruction, tool schema, or provider-visible request assembly changed.
- Codex: Not applicable; no Codex prompt, instruction, tool schema, or provider-visible request assembly changed.

## Change-shape breakdown

Classification rule: production TypeScript is source; Vitest coverage is tests/fixtures; Markdown and changelog JSON are docs; no config/tooling, generated, or binary changes.

| --- | ---: | ---: |
| Source | 276 | 40 |
| Tests/fixtures | 902 | 18 |
| Docs | 253 | 2 |
| Config/tooling | 0 | 0 |
| Generated/other | 0 | 0 |
| Total | 1,431 | 60 |

## Changelog

- Changelog: updated
- Items: 2026-08-20 · stable-junction-profile-replays

## Deployment concerns

- Deployment: applicable
- Deploy unit: Core, importer, and device-syncd ship together in one Cloudflare runner artifact. No Vercel or schema tandem deploy is required.
- Expected exposure: Each active Junction account with a profile and a marker other than 2 receives one non-paginated profile GET, one existing canonical import, and one continuation on normal reconcile cadence. There is no fleet job. The provider contract is one current profile snapshot per source; response bytes remain capped, but profile record count has no separate explicit cap.
- Supported skew: Each runner process bundles matching Core/importer/device-syncd code. Old warm processes stay on the old bundle until drained.
- Safe order: Deploy the single compatible runner artifact and let old warm processes drain.
- Rollback floor: Prefer roll-forward. A revision-1 binary treats marker 2 as stale and can write 1 again, causing a 2→1→2 refetch bounce; pause Junction profile pulls before any emergency rollback.
- Reversibility: Pause Junction profile-producing pulls before reverting; resume only after the compatible runner artifact is restored.
- Convergence proof: Confirm the active runner artifact contains the paired versions and an unchanged released-shape profile refreshes once before later scheduled reconciles skip it.
- Post-deploy checks: Confirm the active runner version, one-time marker advancement, bounded profile request volume, and absence of repeated profile source-revision conflicts or duplicate profile spines.

ReviewGPT context sensitivity: sensitive — this changes persisted health-data migration, conflict, and member-revision behavior.

ReviewGPT first-reviewed head: b174c7722a62f8a75e54b8f52df2f213deadbbd0


rocketman-21:
## ReviewGPT anomaly retrospective — round 2

Original requirement: recover the released stable-ID Junction profile cohort whose unchanged equal-revision replay moves canonical time from `updated_at` to `created_at`, while retaining member edits and deletions and keeping real semantic conflicts fail-closed.

First-reviewed shape: one exact Core timestamp-migration predicate plus a stable-profile normalizer marker (125 authored-source lines). Round 1 correctly found that already-checked accounts could not reach it. Advancing the existing account-wide profile normalization scalar from revision 1 to 2 added two authored-source lines of churn and reused the one-shot admission owner, but round 2 showed that this broad admission also replays the supported no-ID cohort. Its created-at-derived external reference differs at the same provider revision, outside both the stable-ID predicate and the existing strictly-newer no-ID identity migration.

Decision: continue with a narrowed, single-owner correction. Keep revision 2 account-wide so the intended stable-ID cohort is reachable. In Core's existing no-ID predecessor scope, classify only an exact equal-revision updated-at-to-created-at replay as a canonical no-op: retain the old canonical identity and its latest member edit or deletion, append no new event-spine revision, and reject a same-revision semantic change atomically. The existing strictly-newer no-ID path remains the only path that migrates a no-ID identity. This closes the admitted cohort without duplicating identity classification in device-syncd.

Complexity and load: add no flag, queue, repair job, state owner, scan, or provider call. Reuse the already-built vault-local external-reference index and bounded profile scope; the revision-2 rollout still performs at most the existing one bounded profile refresh per previously checked account, and a profile emits only its bounded canonical facets. Align the PR outcome/changelog to the stable-ID migration promise and disclose the no-ID equal-revision no-op.

Required proof before another round: a production-shaped revision-1 scheduled reconcile with a persisted legacy no-ID profile must prove one live spine, retained edit and tombstone, revision 2 only after successful import, replay no-op behavior, and atomic rejection of a real same-revision semantic change. The hosted bounded-continuation path must remain covered. No first-reviewed baseline reset.

PR #2125, opened 2026-08-21
perf(assistant): compact automation tool schema

rocketman-21:
## Outcome

- Reduce the deferred `murph.automation` model-facing schema while preserving the full canonical runtime validator, action set, defaults, authority, and behavior.
- Derive strict inline per-action discovery contracts from the canonical JSON schema and fail visibly back to the full schema if the supported structure changes.

## Product UX result

Internal Patch with no changed member capability or control. Existing automation-enabled sessions intentionally take one fresh-thread transition after rollout. Retained ordinary history starts on a member boundary; omitted context is marked explicitly; and the first turn must inspect authoritative state or clarify instead of guessing unavailable intent. Independently initiated reminders, Assistant Ask continuations, and imported private completions remain available when the member immediately replies even if native resume is unavailable. The following successful turn resumes normally.

## Direct evidence

- Eight focused automation-schema, hosted-domain, planning, automation-continuity, private-completion, notification, finalizer, and store suites: 299 passed.\n- The exact welcome-continuity integration suite passes 16/16 with the additive standalone-context assertion.
- Assistant-engine and operator-config typechecks: passed.
- Canonical/model schema: 15,353/8,427 minified bytes and 4,533/2,565 o200k tokens. The model advertisement is 54.89% of canonical bytes.
- Parity tests mechanically cover every canonical action, root property, per-action allowed root, and per-action required root; synthetic differing variants prove action-local value constraints remain scoped. Representative cross-action fields are rejected, local refs are absent, and unsupported shapes fall back to the full schema.
- Exact old-full-schema to compact-schema regressions prove one fresh thread, an incomplete marker, coherent retained-member boundaries, persistence of the compact fingerprint, and native resume on the following turn. Count and byte bounds prove marker reservation cannot leave an ordinary assistant-only suffix.
- A production-path private-completion regression runs real retention and reconciliation, proves the current writer stamps generic standalone provenance, then removes that new field to prove already-persisted outbox-provenance entries remain compatible.
- A composed scheduled-reminder regression runs real retention and persistence, uses an incompatible turn-scoped model so native resume is null, and proves an immediate `Done.` receives the incomplete marker plus the exact reminder while excluding the older ordinary assistant answer.
- Parent Product UX re-review: Ready. Missing historical intent is not represented as current intent, while the visible independently initiated message a member answers remains available.
- Round-3 retrospective selected shrinking continuation and reduced the projector from 294 to 197 lines. Round-6 retrospective replaced the repeated producer-specific replay classifier with one generic transcript fact; no service, queue, state machine, lifecycle, migration, registry, or new owner was added.\n- Final ReviewGPT round 7: PASS with no qualifying findings on the last substantive head; the final docs-only plan archival does not change reviewed behavior.

- Deferred tool search and provider discovery metadata change; canonical parsing, validation, diagnostics, writes, scheduling, routes, and execution semantics do not.
- Top-level field descriptions are not globalized from one action. Nested value guidance remains, while action-specific allowed fields and differing value schemas live in the matching action branch.
- The complete dynamic-tool catalog is hashed into the native-thread contract. Existing resumable automation-enabled sessions therefore start one fresh provider thread on their first eligible turn after rollout. Replay remains bounded to 24 messages, 4,000 bytes per message, and 12,000 bytes total.
- When retention or marker reservation leaves no member message, replay discards ordinary dependent assistant output and retains only transcript entries with generic standalone-assistant provenance. Existing `sourceOutboxIntentId` is read only as compatibility for private completions persisted before the additive field.

## Architecture and reuse

- Existing systems reused: the existing Zod automation union remains the sole runtime and diagnostic owner; the existing contract fingerprint, fresh-thread fallback, bounded transcript replay, session persistence, notification finalizer, and private-completion reconciliation own rollout recovery.
- New logic: one dependency-free projection derives strict model discovery from the canonical automation schema. One optional true-only field inside the existing transcript schema identifies assistant output whose meaning does not depend on a preceding member transcript entry in that session.
- New abstractions: one focused pure schema-projection helper; no new service or state owner.
- Complexity intentionally avoided: no local-reference resolver, second handwritten schema, compatibility fingerprint, producer exception table, registry, generator, transport fork, dependency, cache, queue, or lifecycle machinery.

## Hot reply path impact

The PR adds no database, network, provider, transaction, retry, or awaited operation. On a fresh-thread fallback, existing bounded transcript work performs at most two linear scans/splices over the capped 24-message suffix. The replay candidate now carries one boolean instead of an outbox producer id. Persistence adds one optional true-only property to independently initiated assistant transcript entries. Ordinary replies remain unchanged.

- Individual: the identical complete serialized `thread/start` plus first `turn/start` fixture using production base instructions, ordinary resolved direct tools with automation enabled, a fixed reminder prompt, no images, and `gpt-5.6-sol` drops from 31,007 bytes / 7,739 o200k tokens to 24,081 bytes / 5,771 tokens. Delta: -6,926 bytes (-22.34%) and -1,968 tokens (-25.43%).
- Group: the matching complete fixture with the group tool enabled and automation unavailable is unchanged at 27,033 bytes / 6,072 o200k tokens. Delta: 0 bytes / 0 tokens.
- Attribution: all measured change is the deferred automation input schema. Base was reconstructed by replacing only that descriptor schema with the canonical runtime schema; instructions, tool descriptions, other tools, generated guidance, turn input, and `turn/start` params were byte-identical. Measurement used minified JSON and `gpt-tokenizer` 3.4.0 `o200k_base`; provider-internal framing and HTTP headers are excluded.

## Deployment concerns

- Deployment: applicable
- Supported skew: old and new runner processes use the same session and transcript files. The new transcript property is additive; old runners retain their pre-change permissive replay behavior, while new runners use the property. No tandem Web/Worker protocol rollout is required.
- Safe order: use the ordinary assistant-runner rollout and let warm old processes finish. New writers stamp standalone context before relying on the stricter replay filter.
- Rollback floor: either descriptor version can read persisted session state. A rollback restores the previous permissive assistant-only replay behavior; re-upgrade restores the stricter semantic filter. Avoid running old retention maintenance between a rollback and re-upgrade if preserving newly stamped optional properties is operationally important.
- Expected exposure: previously resumable automation-enabled sessions encounter one descriptor transition. Already-persisted private completions remain compatible through outbox provenance. Pre-deploy scheduled transcript entries do not have the new semantic field; a rare post-deploy fresh-thread reply to one may conservatively require clarification.
- Reversibility: reverting restores the full provider schema without changing canonical automation records; already-transitioned sessions take one reverse fresh-thread transition.
- Convergence proof: focused regressions prove the exact old fingerprint transition, bounded replay, marker-only ordinary replay after member-text retirement, generic standalone replay for current reminders and private completions, compact fingerprint persistence, and following-turn native resume.
- Post-deploy checks: smoke two consecutive automation-enabled turns on one existing session; confirm the first uses bounded fresh-thread replay and the second resumes. Also send one scheduled reminder with an incompatible override and confirm a short direct reply retains that reminder. Monitor contract-mismatch fresh starts and reply failures during rollout.

## Changelog

- Changelog: not applicable
- Reason: internal provider-input and rollout-continuity work with no new or changed member-facing capability, promise, or control.

## Change-shape breakdown

Classification uses base-to-head numstat; the completed plan and architecture note are docs, derivation/descriptor/replay/transcript code is source, and focused parity/continuity files are tests. No binaries or generated files changed.

| --- | ---: | ---: |
| Source | 297 | 20 |
| Tests/fixtures | 938 | 14 |
| Docs | 90 | 1 |
| Config/tooling | 0 | 0 |
| Generated/other | 0 | 0 |
| Total | 1325 | 35 |

Reason: this changes the provider-visible schema and one fresh-thread replay safeguard at a runtime contract boundary, so later substantive rounds receive a full snapshot.

ReviewGPT first-reviewed head: 61f721c2cdfdb8d712e6d6d3e2ccc52023788948


rocketman-21:
## Round 6 requirement-level retrospective

The original requirement remains compacting the model-facing `murph.automation` schema from the canonical runtime schema without changing automation behavior. The first-reviewed source shape was +277/-2; the current source shape is +265/-20. The compactor itself shrank, while review remediation expanded a shared replay correction and exposed the same classification mechanism in rounds 4–6.

Round 6 proves the current `sourceOutboxIntentId` exception is producer-specific: it preserves imported private completions but not scheduled assistant-initiated reminders when an incompatible continuation target forces transcript fallback. A short reply such as “Done” can therefore lose the reminder it answers after older member text is retired.

Decision: continue this PR with a bounded redesign inside the existing transcript owner. Represent one generic semantic fact for independently initiated assistant context; ordinary assistant replies remain dependent on a retained member message. Private completions and scheduled reminders must use the same fact. Avoid a producer exception matrix, new service, queue, lifecycle, or state owner. Existing delivery/outbox provenance may be used only as a narrow compatibility proof if current persisted data requires it.

Reset concept budget: one optional transcript provenance field, existing transcript persistence/parsing, existing replay planner, and focused composed tests. Delete the producer-specific classification once the semantic owner covers it. Required proof covers ordinary dependent answers, real private completions, scheduled automation with an incompatible target and null resume, mixed standalone/dependent output, marker byte/count reservation, persistence, and the following resume. The next final ReviewGPT round remains required and must return PASS.

PR #2127, opened 2026-08-21
Make private-to-group handoff replay exact

rocketman-21:
## Why and outcome

Exact retries of an accepted private-to-group context handoff could be rejected because the retry rebuilt time-dependent envelope fields, while the Cloudflare group-tool transport did not retry `handoff` after a lost successful response or retryable 5xx. A concurrent duplicate could also perform retained-root provider work while holding the handoff transaction, and untrusted context could contain the raw prompt delimiter.

This change exact-replays the same hidden Cloudflare-to-Web request once, reuses and revalidates the persisted handoff, escapes delimiter characters inside the JSON payload, and prepares both private-origin and target mailbox roots before cache-only selection, append, and replay transactions using the existing bounded preparation retry.

## Product UX

- Effort: Patch.
- Result: Ready.
- Affected people: A member whose first successful handoff response is lost now receives the original accepted group target instead of a false failure. The same private request remains pinned after joining another group; changed context or another selected group conflicts, and lost membership or route authority remains unavailable. Group participants still receive one target-authored message through the existing delivery path.
- Exclusions: No hosted Web UI or delivery-copy change. The behavior stays within the existing explicit private-to-group consent and queue-only delivery contract.

## Evidence

- Direct journey: 62 focused tests pass—19 Cloudflare transport tests and 43 Web/mailbox tests. They cover byte-identical handoff replay after lost successful bodies and retryable 5xx responses on proxy and direct transports, later-time replay, membership-count drift, committed and concurrent duplicates, changed context/target conflict, membership and route loss, delimiter-safe JSON round-trip, distinct private-origin and target roots, and a retained root that becomes visible only after the first transaction begins.
- Composed boundary proof: The transport sends exactly two identical requests and returns the accepted second response; two Web admissions return the same mailbox item while resolving the destination and appending only once. A real secure-box composition verifies two distinct signed roots and four KMS unwraps across fresh admission and replay, all before transaction entry, with one mailbox insert total. The cache-only race rolls back its first transaction, prewarms the winning retained root outside the transaction, and replays locally on the bounded second attempt.
- Focused checks: Exact pushed-head Web and Cloudflare typechecks, docs drift, and `git diff --check` pass. Web ESLint reports zero errors and 46 known unrelated warnings.
- Broader proof: The original candidate patch passed the full Web verifier before its final unrelated base fast-forward: 10,890 tests passed, 511 skipped; dev smoke, lint, and production build passed. Exact-head PR CI remains authoritative after remediation.
- Review remediation: The specialist retained-root race, round-one missing transport replay, and round-two split private-origin/target root-preparation findings are accepted and fixed using existing owners. Round two correctly required the recorded requirement-level retrospective before this remediation; no ReviewGPT patch artifact was returned or applied.

- Cloudflare group-tool transport: `handoff` joins the existing exact-replay-safe action set, adding at most one serial retry within the original total deadline for retryable response failures; direct/proxy focused tests prove byte identity and the single-retry bound.
- Mailbox payload crypto: Fresh admission, concurrent convergence, and persisted replay share one request-scoped cache. The private-origin conversation root and target mailbox root are both prepared before selection/append/replay transactions, whose provider fence fails into the existing bounded fresh-cache retry; real secure-box and retained-root race tests prove unwrap, rollback, prewarm, and retry ordering.
- Deterministic event identity and replay: The stored wake remains the source of truth for timestamps, expiry, target runtime, delivery identifiers, prompt profile, response policy, route, and context; focused admission tests prove exact and conflicting retries.
- Membership and route authority: Replay remains pinned to the persisted membership and target while current membership generation and route egress authority are revalidated; focused tests cover drift and revocation.
- Provider-visible prompt assembly: JSON remains inside the explicit untrusted-data wrapper while `<`, `>`, and `&` are Unicode-escaped so private context cannot close the wrapper; a round-trip regression test proves the original context is preserved.
- Mailbox poll wake-up: Successful exact replay returns and re-signals the persisted mailbox item, using the existing poll path without a second append.

## Architecture and reuse

- Existing systems reused: Cloudflare’s one-shot exact-replay transport, deterministic handoff event IDs, hosted mailbox prepared append crypto, request-scoped domain-root unwrap caching, membership authority, route egress authority, and the existing notification wake contract.
- New logic: One additional replay-safe action case, exact persisted-wake validation, and one shared root-preparation/cache-only boundary for fresh admission and replay; existing owners supply the total retry deadline and bounded fresh-cache retry.
- New abstractions: None. Existing transport, mailbox, and crypto ownership boundaries are sufficient.
- Complexity intentionally avoided: No new queue, retry store, schema, capability negotiation, or delivery protocol; the persisted mailbox wake remains the single replay record.

## Hot reply path impact

Not applicable. This changes a user-invoked private-to-group handoff control path, not the Foreground Reply Critical Path from current-conversation acceptance through provider start and durable reply handoff. On this handoff path only, a retryable Cloudflare-to-Web response failure can add one serial byte-identical request within the original total timeout; cancellation, non-retryable errors, or an exhausted deadline do not retry. No new provider call or fallback is added, and private-origin plus target-root provider work finishes before every handoff database transaction.

| Runtime | Base | Head | Delta |
| --- | ---: | ---: | ---: |
| Individual | Not applicable | Not applicable | This path does not change the individual runtime’s initial provider-visible request. |
| Group handoff changed prompt field | 129 tokens / 604 UTF-8 bytes | 142 tokens / 629 UTF-8 bytes | +13 tokens (+10.1%) / +25 bytes |

Measured with `gpt-tokenizer` 3.4.0 using `o200k_base` and identical delimiter-bearing context fixtures. The measured delta isolates the changed provider-visible handoff instruction field; unchanged system instructions, conversation history, tool schemas, and generated guidance are excluded because this PR does not alter their assembly. The change is entirely the JSON Unicode escaping needed to preserve the untrusted-data boundary.

Classification reason: This changes privacy-bearing untrusted prompt handling, idempotent membership and route authorization, transaction-local crypto behavior, and the Cloudflare-to-Web replay boundary.
ReviewGPT first-reviewed head: 5e369af9acd4395e0c80ca8b048f29e1e8ac0577

## Deployment concerns

- Deployment: applicable
- Supported skew: The request, response, and persisted wake shapes are unchanged. New Web with an old runner is safe but does not recover lost handoff responses; a new runner with preceding Web fails closed on its retry but is not a supported steady state because truthful recovery is absent. A new Worker may temporarily serve warm old runner containers safely, with recovery unavailable until container convergence.
- Safe order: Deploy Web first, then deploy the Cloudflare Worker and runner bundle with `container_rollout=immediate`.
- Rollback floor: Roll back the runner bundle before Web. If Web must roll back first, treat handoff replay as degraded until runner rollback convergence; no duplicate or widened delivery is authorized.
- Expected exposure: During rollout, old runners can still report a false failure after a committed handoff. During the 10-minute handoff TTL, an old unescaped row containing `<`, `>`, or `&` can fail closed as `request_conflict` on new Web. Neither case creates a fresh target selection or append.
- Reversibility: No schema or protocol migration exists. Revert runner first and then Web; already escaped prompt strings remain valid data.
- Convergence proof: Managed-container smoke must report the new runner-bundle fingerprint, and a lost-first-response handoff smoke must return the already accepted mailbox item on the byte-identical retry.
- Post-deploy checks: Confirm Web health, runner bundle convergence, one direct handoff replay smoke, and bounded handoff conflict/transport error aggregates through the 10-minute compatibility window.

## Changelog

- Changelog: not applicable
- Reason: The existing private-to-group consultation changelog item already promises pinned, bounded handoff behavior; this PR makes retries honor that published contract without adding a new member-visible capability.

## Change-shape breakdown

Classification rule: TypeScript production code is Source, the Vitest files are Tests / fixtures, Markdown owner docs are Docs; there are no binary or generated changes.

| --- | ---: | ---: |
| Source | 398 | 115 |
| Tests / fixtures | 1251 | 39 |
| Docs | 34 | 9 |
| Config / tooling | 0 | 0 |
| Generated / other | 0 | 0 |
| Total | 1683 | 163 |


rocketman-21:
ReviewGPT anomaly retrospective before remediation

- Original requirement: one private-to-group context handoff must durably append exactly once, revalidate live origin/membership/route authority, and exact-replay the committed admission without provider/KMS work inside the locked database transaction.
- First-reviewed shape (`5e369af9acd4395e0c80ca8b048f29e1e8ac0577`): the handoff transaction could decrypt both the target handoff row and private origin row while locked.
- Current reviewed shape (`839c0a9ff6154fd2b996516e8d40ccf3a14a25ce`): the accepted round-one remediation fenced the append transaction and prepared the target mailbox root, but did not prepare the distinct private-origin conversation root; fresh admission can therefore fail cache-only, while direct persisted replay still permits the same origin-root provider work under lock. This is the repeated retained-root ownership mechanism, not a new product requirement. Authored production-source churn is 352 lines, below the numeric threshold.
- Decision: continue in this PR, collapse fresh/concurrent/direct replay onto the existing bounded mailbox-crypto preparation owner, prewarm every encrypted row that locked authority revalidation may read, and keep the transaction cache-only. Add no state, queue, retry owner, migration, or compatibility path. Add a production-faithful composed regression proving provider work completes before transaction entry, fresh append occurs once, and exact replay reuses the same item.
- Scope discipline: preserve the existing origin, membership, target, and route owners; remove the split preparation boundary instead of adding another abstraction. The separate base-drift test-mock repair remains test-only.

PR #2148, opened 2026-08-22
Preserve exact workout context across follow-ups

rocketman-21:
## Why and outcome

Murph could ask a workout-specific next-set question without carrying the exact
workout record into the following turn. A natural reply could then be impossible
to save without asking the member for Murph's internal workout identity.

Murph now appends the exact canonical workout id to an ordinary private
follow-up question. The same visible text is delivered and persisted. A later
reply exact-reads and mutates that workout through the existing canonical owner.

## Product UX

- Effort: Patch.
- Result: Ready.
- Walkthrough: In an ordinary private conversation, Murph asks about one exact
  workout and visibly carries its id. The member replies naturally without
  supplying or retyping the id, and Murph updates only that workout. Ambiguous
  questions still clarify and never select by recency.
- Material exclusions: No active-workout singleton, focused-workout state,
  recency fallback, new persisted state, hidden/public text split, parser,
  scrubber, eligibility flag, notification integration, or card change.

## Evidence

- Hosted evidence: The feedback-associated runtime attempt completed three
  assistant passes and three deliveries with zero warning or error rows. The
  turn invoked workout commands and submitted feedback successfully, isolating
  the failure to continuation context rather than runtime or delivery.
- Direct runtime: 270 assistant Codex runtime tests pass. The affected
  automation, scheduled-notification, and tracked-table skill suites pass.
  Reopened-chat rendering passes with the same visible marker, and self-authored
  echo suppression passes when delivery and transcript carry identical text.
- Model provenance: The opt-in real-Codex journey sources the positive id from a
  runtime-generated tracked-workout marker. A valid-looking id found only in
  quoted member text produces no marker or mutation. The default file passes 6
  tests with 74 live-provider cases skipped; the live case is locally blocked
  because `OPENAI_API_KEY` is not configured.
- Checks: Assistant Engine and Assistant CLI typechecks pass; Assistant CLI
  helper tests pass 6/6; changelog tests pass 57/57; `git diff --check` and the
  identifier scan pass.

## Review findings

- Preliminary specialists: accepted the scheduled structured-output conflict,
  malformed hidden-marker handling, and insufficient real-model provenance.
  The provenance test remains. The parser/scrubber and eligibility fixes were
  later deleted when the hidden-id requirement was removed.
- Final round 1: accepted the scheduled-notification conflict. Its intermediate
  runtime gate was superseded by the simpler same-text design.
- Final round 2: accepted that divergent delivered/transcript text could expose
  the hidden marker in reopened chat and break id-less self-echo matching. The
  required retrospective identified the hidden/public split as the repeated
  mechanism; the user explicitly accepted visible ids, so the correction
  deletes that mechanism instead of adding consumer-specific branches.
- Rejected: None.

- The visible marker appears in delivered conversation text, durable transcript
  history, reopened local chat, and exact-text self-echo comparison.
- Scheduled notifications, structured output, groups, output-only turns, and
  maintenance turns retain their existing contracts and never emit the marker.
- The marker carries context only. Every mutation still requires an exact
  canonical workout read and the existing guarded write path.

## Architecture and reuse

- **Existing systems reused:** Deferred tracked-workout guidance, ordinary
  message delivery/transcript persistence, canonical workout ids, and exact
  workout reads and writes.
- **Production runtime changes:** None relative to the PR merge base.
- **New abstractions:** None.
- **Complexity removed from the previous candidate:** The public/private split,
  suffix parser, malformed-fragment scrubber, per-turn eligibility boolean, and
  consumer-specific transformation tests. The correction deletes 268 lines and
  adds 74, for a net deletion of 194 lines.

## Hot reply path impact

No runtime code, state, branch, database operation, provider call, retry, or
fallback is added. The model authors one extra visible line, and the existing
generic message path delivers and persists it unchanged.

- Individual: Not applicable. The rule lives in the deferred tracked-table
  skill and does not enter the initial provider request.
- Group: Not applicable. Group turns are explicitly excluded.
- A later routed private workout turn receives the narrow follow-up rule only
  when the existing tracked-table skill is loaded.

## Design proof

Not applicable. No hosted Web UI or production component changed.

## Change-shape breakdown

Classification treats the deferred skill as source, test files as
tests/fixtures, and the product spec, plan, Frog entry, and changelog as docs.
No config, generated, or binary files changed.

| --- | ---: | ---: |
| Source | 25 | 1 |
| Tests/fixtures | 214 | 7 |
| Docs | 134 | 3 |
| Config/tooling | 0 | 0 |
| Generated/other | 0 | 0 |
| Total | 373 | 11 |

Classification reason: The change affects hosted assistant continuity and the
context used before an exact canonical workout mutation.

ReviewGPT first-reviewed head: 1a275d38da6d6367add00d30a6e044e41757c3f3

## Deployment concerns

- Deployment: applicable
- Safe order: Deploy the updated Cloudflare runner bundle through the normal
  hosted release path. The Web changelog may deploy independently.
- Supported skew: Old and new runners share the same workout, transcript, and
  tool contracts. No tandem deployment or migration is required.
- Rollback: No schema, state floor, or new runtime representation is introduced.
- Post-deploy checks: Confirm the deployed runner fingerprint, run one
  controlled next-set question and natural reply, and verify the exact workout
  was read and updated with no bounded assistant/outbox errors.

## Changelog

- Changelog: updated
- Items: 2026-08-22 · workout-replies-keep-their-place

### DIFF
--- a/.agents/friction-log/20260822030752-hosted-runtime-triage/friction.md
+++ b/.agents/friction-log/20260822030752-hosted-runtime-triage/friction.md
@@ -1,0 +1,34 @@
+---
+title: 'Hosted runtime triage skill references retired repo paths'
+severity: 'minor'
+---
+
+## Expected Behavior
+
+The hosted-runtime triage skill should route investigators only to current
+repository-owned runtime and coordination documents.
+
+## Current Behavior
+
+The skill requires `packages/hosted-orchestrator-temporal/README.md` and
+`agent-docs/exec-plans/active/COORDINATION_LEDGER.md`, but neither path exists
+in the current repository. The investigator must stop and search for replacement
+owners before continuing.
+
+## Possible Solution
+
+Update the skill to reference the current Temporal ownership document and
+current coordination mechanism, or mark those reads conditional when no
+repository-owned replacement exists.
+
+## Minimal Reproducible Example
+
+1. Start a hosted-runtime failure investigation.
+2. Follow the skill source order.
+3. Attempt to read both required paths.
+4. Observe that both are absent from the checkout.
+
+## Context
+
+This interrupted a time-sensitive hosted assistant reply investigation but did
+not block production root-cause proof.
--- a/agent-docs/exec-plans/active/2026-08-22-workout-followup-context.md
+++ b/agent-docs/exec-plans/active/2026-08-22-workout-followup-context.md
@@ -7,9 +7,9 @@ Updated: 2026-08-22
 ## Goal
 
 - Let Murph carry the exact canonical workout id it already resolved into the
-  durable context of a plain-text follow-up question.
-- Keep the id hidden from the member and reuse the existing transcript as the
-  only continuation owner.
+  visible and durable text of a plain-text follow-up question.
+- Deliver and persist one identical response, reusing the existing transcript
+  as the only continuation owner.
 
 ## Product UX Patch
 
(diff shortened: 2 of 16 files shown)


rocketman-21:

Original requirement: preserve the exact canonical workout id across a plain-text follow-up so a natural reply can exact-read and mutate the intended workout without asking the member to supply Murph's internal identity.

Requirement correction: identifier visibility was an inferred polish preference, not a security, authorization, or user requirement. The member may see the canonical id. The product invariant is only that Murph carries it and never requires the member to know or retype it.

Shape comparison: the first-reviewed head added a private/public text split and suffix parser. Round-one remediation expanded that mechanism with per-turn eligibility and malformed-fragment scrubbing. The same split then reached scheduled-notification and ordinary transcript consumers differently, producing successive findings from one public-versus-private ownership assumption.

Decision: complexity collapse by deletion. Remove the runtime split, parser, scrubber, eligibility boolean, and their consumer-boundary tests. Murph will emit one marker-bearing question whose delivered and persisted text are identical. The deferred workout instruction continues to exclude structured, scheduled, group, output-only, and maintenance response contracts. No new state owner, field, queue, selector, or consumer-specific scrub branch is added.

Required preserved behavior: a fresh provider turn receives the exact id from the preceding marker, exact-reads before mutation, and never treats a valid-looking id in unrelated quoted text as authority. Scheduled structured-notification output remains unchanged because no runtime transform applies to any response.

Continuation decision: proceed in this PR. The user explicitly accepted visible ids, and the correction deletes the repeated mechanism instead of repairing another consumer.

PR #2160, opened 2026-08-22
fix(device-sync): expose bounded pass queue diagnostics

rocketman-21:
## Why and outcome

Hosted device-sync pass logs showed elapsed time and processed jobs but could not distinguish a shrinking queue from replenishing or repeatedly stuck work. This adds a bounded before/after local queue snapshot to the existing `device-sync.pass_finished` event without changing timeout, scheduling, retry, or provider behavior.

Each snapshot exposes only capped counts, truncation, closed-vocabulary job-kind counts, sampled age/attempt aggregates, and the existing 100-job-limit result. It never emits job IDs, account IDs, payloads, resources, provider prose, or member identifiers.

## Product UX

Not applicable. This is internal background-runtime observability and does not change a member journey, visible state, delivery, permission, or recovery contract.

## Evidence

- `pnpm --dir packages/assistant-runtime exec vitest run --config vitest.config.ts --isolate=true --no-coverage test/hosted-runtime-maintenance.test.ts` — 95 tests passed.
- `pnpm --dir packages/assistant-runtime typecheck` — passed.
- Production-parser proof runs ordinary and timeout-yield lifecycle requests through `parseHostedRuntimeLogRequest` before asserting stored entries.
- Boundedness proof supplies 101 pending rows across every current label, the supported legacy deauthorization value, and a parser-dangerous unknown; it verifies a 100-row sample, closed category counts, and paired lifecycle acceptance.

- Hosted runner maintenance: when a runtime log port is active, the selected local account gets two additional indexed SQLite reads capped at 101 rows each. Regression proof covers completed and timeout-yielded passes.
- Hosted runtime logs: only additive parser-native scalar metadata on the existing accepted event code changes; no Web event parser, schema, persistence owner, or alerting owner changes.

## Architecture and reuse

- Existing systems reused: the current account-scoped pending-job reader, pass lifecycle event, production log parser, and best-effort runtime-log queue.
- New logic: one private bounded aggregator derives before/after counts, truncation, readable scalar `kind=count` summaries, oldest sampled age, maximum sampled attempts, status counts, and whether the existing job limit was reached. The existing device-sync logging boundary maps jobs to seven fixed safe labels.
- New abstractions: one file-local snapshot type and callback carry diagnostic aggregates from the worker drain to the existing lifecycle logger; no package or persisted-state abstraction was added.
- Complexity intentionally avoided: no arbitrary kind encoding, shared compactor, distinct-kind truncation fields, object-array parser allowlist, Web-first rollout, new event code, database query owner, remote callback, scheduler, retry path, provider branch, persisted metric, or timeout policy.

## Hot reply path impact

Not applicable. The reads occur only in the background device-sync worker-drain stage, and the existing info log remains buffered and best-effort off the Foreground Reply Critical Path.

- Individual runtime: Not applicable; no prompt, instruction, tool schema, provider-visible request, or model assembly changed.
- Group runtime: Not applicable for the same reason.

## Risks

- Database fanout: at most two local SQLite reads per logged pass, each account-scoped and capped at 101 rows; zero Postgres queries, pooled connections, transactions, network calls, or crypto operations are added.
- Privacy: aggregates omit identifiers and payloads; current and supported legacy job kinds map to a closed parser-safe vocabulary and unknowns aggregate under `other`.
- Diagnostic failure: queue inspection is best-effort and returns an absent snapshot instead of changing device-sync control flow.

## Change-shape breakdown

Classification rule: authored runtime code is source, Vitest coverage is tests/fixtures, and the execution plan is docs. No binary or generated files changed.

| --- | ---: | ---: |
| Source | 162 | 5 |
| Tests/fixtures | 178 | 2 |
| Docs | 109 | 0 |
| Config/tooling | 0 | 0 |
| Generated/other | 0 | 0 |
| **Total** | **449** | **7** |

## Deployment concerns

- Deployment: applicable
- Supported skew: Existing Web and Worker readers accept the unchanged event code and additive parser-native redacted JSON; old runners safely omit the new fields while new runners emit them.
- Safe order: Deploy the Cloudflare Worker/runner bundle only; no Web deployment is required. Use immediate container convergence for prompt fleetwide diagnostic availability, not for schema compatibility.
- Rollback floor: None; the change writes no protocol, schema, or durable runtime state and an older runner can safely resume while omitting the fields.
- Expected exposure: During rollout, a bounded set of old warm runners may omit queue fields; immediate convergence can interrupt active background runner work, which remains checkpoint/retry safe.
- Reversibility: Revert or redeploy the prior runner bundle; stored additive log rows remain readable.
- Convergence proof: Protected deploy workflow plus managed-container smoke must report the exact reviewed runner-bundle fingerprint.
- Post-deploy checks: Confirm live `device-sync.pass_finished` rows carry bounded queue fields, then compare timeout queue deltas and job-kind/attempt aggregates before changing the 90-second budget.

## Changelog

- Changelog: not applicable
- Reason: Internal diagnostic metadata is not member-visible and does not itself change sync or recovery behavior.

Reason: Production hosted-runtime and deploy behavior changes, although the event taxonomy and control flow remain unchanged.

ReviewGPT first-reviewed head: 268bf370cb1f2b8379767dbf55e04c8843bf67b5


rocketman-21:
Anomaly retrospective: The original requirement is bounded before/after queue evidence that distinguishes recovering from stuck device-sync passes; per-kind attribution remains required because it separates the work category driving a persistent backlog. The first-reviewed head used object arrays, while the current correction uses scalar kind/count strings and added four source lines; both exposed the same producer/parser contract mismatch because the general log-code sanitizer accepts values the Web redacted-scalar parser can reject. Decision: redesign and continue narrowly. The existing device-sync pass logging boundary will own a closed seven-label vocabulary: backfill, deauth, delete, push_source_recovery, reconcile, resource, and other. Current deauthorize and supported legacy deauthorization both map to deauth; unknown values map truthfully to other. The shared compactor and distinct-kind truncation fields will be removed because the closed vocabulary is below the parser cap. No Web parser change, allowlist, compatibility state, migration, or new owner is added. Production-boundary tests will cover every current label, the supported legacy value, a parser-dangerous unknown, and paired pass-started/pass-finished acceptance.

PR #2216, opened 2026-08-24
Use generated identities for new reminders

rocketman-21:
## Why and outcome

Reminder titles were accidentally acting as identities because ordinary saves derived the canonical Markdown path from the title. A valid new reminder could therefore fail when an older reminder had the same name.

Ordinary hosted saves now omit `slug` and receive a generated `automationId`, so titles may repeat. A current loaded skill may still supply only its exact stable, non-title-derived recipe key for a singleton product recipe. Inspection resolves that key to the authoritative ID; every patch uses the ID and cannot rename the key.

## Product UX

- Effort: Patch
- Result: Ready
- Outcome: Members can create reminders with the same name and later change exactly one by its returned ID, while existing skill-owned recipes remain discoverable and editable.
- Reaches: Members creating ordinary reminders and members using group newsletter, experiment-support, or onboarding recipe flows.
- Walkthrough: Save two ordinary reminders with the same title and verify distinct IDs; inspect and patch one ID without changing the other; then save, inspect, and patch the documented group newsletter recipe through its stable key.

## Evidence

- Simplification baseline: Core 814 tests and Assistant Runtime 2,490 tests passed after replay fingerprinting was deleted.
- Final correction: 102 focused Assistant Engine schema/parser/prompt tests passed; the production-owner group automation recipe test passed with 304 unrelated cases skipped.
- Typechecks: Core, Assistant Engine, and Assistant Runtime passed on the final owners.
- `git diff --check` and privacy-marker scan passed.
- Clean current-main merge-tree: `42487fa688156f716bf66485b1d6fbf1d9b9bd28` against `origin/main` `c10b9b0fc25d28c7a6d7749292caeec69b816653`.
- Exact pushed head: `3212f9d11983259dc252143aa12a11673def062c`.

- Canonical automation storage: ordinary create-only writes use a generated ID for their internal compatibility slug and path, never the title.
- Skill-owned recipes: current group newsletter, experiment-support, and onboarding skills may pass only their explicitly defined stable recipe key on save. Parser-to-port and production-owner tests cover this existing contract.
- Mutation boundary: inspection may resolve an exact skill recipe key, returns the authoritative `automationId`, and patches remain exact-ID/version based with no key mutation.
- Compatibility: legacy slug records remain readable and fixed managed automations retain their registered IDs and slugs.
- Daylight-saving recovery: ephemeral retry correlation hashes the exact failed request rather than its title.

## Architecture and reuse

- Existing systems reused: canonical registry locking, generated contract IDs, exact inspection, optimistic versioned patching, existing stable recipe lookup, Markdown writes, and scheduler projection.
- New logic: the model save schema distinguishes ordinary no-key instances from exact stable recipe keys defined by a current loaded skill; patch never accepts key mutation.
- New abstractions: None; the existing automation owner, slug field, registry lookup, and returned automation ID are sufficient.
- Complexity intentionally avoided: no title lineage, suffix allocator, recipe allowlist service, second config owner, request fingerprint, replay ledger, persisted idempotency field, table, migration, queue, or reconciliation process.

## Hot reply path impact

No awaited operation changes. Generic saves use the existing single locked registry mutation; skill recipe lookup uses the existing inspection path. The correction adds no database call, network call, provider call, retry loop, or new service.

| Runtime | Base tokens | Head tokens | Delta | Percent | Base bytes | Head bytes | Byte delta |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| Individual Murph | 27,916 | 27,984 | +68 | +0.2436% | 127,617 | 127,957 | +340 |
| Group Murph | 23,789 | 23,857 | +68 | +0.2858% | 108,885 | 109,225 | +340 |

Measured with `gpt-tokenizer` 3.4.0 `o200k_harmony` against identical representative hosted fixtures. The delta is the assembled initial automation action guidance; the deferred automation tool schema remains outside the first request.

## Deployment concerns

- Deployment: applicable
- Surface: Cloudflare hosted runner.
- Supported skew: persisted schema is unchanged; old and new runners read explicit recipe keys and generated-ID records. Only new runners remove title-derived identity for ordinary saves.
- Safe order: merge and deploy the ordinary public runner. Existing private skills already use the restored stable-recipe contract, so no paired private change, Web coordination, or data migration is required.
- Rollback floor: older code can read every record created by the new code.
- Reversibility: reverting leaves created records valid and readable.
- Expected exposure: during rollout, an old runner can retain the prior title-collision behavior for ordinary reminders; skill recipe records remain compatible.
- Convergence proof: all versions use the same canonical automation registry and scheduler projection, with no secondary state.
- Post-deploy checks: create two synthetic same-title ordinary reminders and verify distinct IDs; inspect and patch one; then confirm the group newsletter stable recipe is accepted, inspected, and patched by its returned ID.

## Change-shape breakdown

| --- | ---: | ---: |
| Source | 40 | 46 |
| Tests/fixtures | 284 | 466 |
| Docs | 15 | 1 |
| Config/tooling | 0 | 0 |
| Generated/other | 0 | 0 |
| **Total** | **339** | **513** |

## Review retrospective

- Decision: Complexity collapse. Exactly-once creation across a durable-save/lost-ack runner restart is not a product requirement; the rare reversible duplicate risk is accepted.
- Deletion: request fingerprinting, accepted-input coupling, exact-create replay, nullable audit results, and their specialized tests were removed with no replacement state.
- Stable-recipe correction: restoring the existing optional save key prevents purpose drift in current skills without restoring title identity or adding another owner. Experiments can migrate traceability to returned IDs separately where that truly deletes state; singleton recipes still need one canonical rediscovery key.

## Changelog

- Changelog: updated
- Items: 2026-08-24/reminders-can-share-names

ReviewGPT first-reviewed head: 3e83d6090b23cf60aedb44069eb3265755e0081f


rocketman-21:
Round 2 anomaly retrospective: the original requirement is to preserve archived reminder history while allowing one fresh same-name successor and making every later retry/update resolve to that one live record. The first-reviewed shape allocated another suffix on retry; the current shape detects the live successor but loses its identity at the generic conflict boundary, so exact inspect still selects the archived root. This is the same unresolved mechanism, not a reason to add lineage state. Decision: continue with an owner-boundary correction at the existing locked registry source of truth. The conflict will carry the already-derived live successor identity; hosted recovery will inspect and patch that exact stable automation ID. Preserve create-only save semantics, fixed-ID conflicts, immutable archives, and exactly one live scheduler identity. Add the production-faithful hosted recovery proof. No schema, alias table, new state owner, state machine, queue, reconciliation loop, or compatibility path.

rocketman-21:
### Requirement-level retrospective — supersedes the suffix-lineage decision

Review round 3 proved that a readable numeric slug suffix cannot establish reminder lineage: an unrelated reminder can legitimately own the same `name-2` shape. The prior decision in this PR to scan suffixes is therefore withdrawn.

The owning contracts establish a smaller invariant:

- `automationId` is the durable scheduler and mutation identity.
- `slug` is a readable Markdown-path and compatibility lookup alias; it grants no authority and does not encode lineage.
- `archived` is a lifecycle state that ordinary authenticated automation patching may reactivate; it is not an immutable historical generation.
- Generic `save` remains create-only. Any exact alias conflict, including an archived record, returns that record’s stable ID; the assistant inspects it and applies an optimistic versioned patch when the request is to reactivate or change it.

Remediation will delete suffix allocation and all inferred-successor behavior. It will add no lineage field, tag, alias registry, or second identity. A naturally numbered reminder stays unrelated because no neighboring slug is scanned. Existing slug reads remain only for vault/CLI compatibility.

rocketman-21:
### First-principles identity retrospective — supersedes the archived-reactivation design

The earlier reviews treated same-title save failure as an archived-record recovery problem. Re-checking the actual boundary shows the smaller root cause: a user-facing title was being used to derive storage identity.

The final invariant is:

- `automationId` is the only model-visible durable identity.
- A generic save creates a fresh host-generated ID even when titles match.
- An existing automation is inspected and patched only by its exact ID.
- Titles are labels and may repeat.
- The persisted slug/path remains internal only for the existing Markdown schema, legacy record reads, and fixed managed automations.

There is no production evidence that ordinary reversible reminder saves need semantic or tool-call idempotency, so this PR adds no retry ledger or dedupe state. It deletes the title-lineage, suffix, and conflict-recovery machinery instead.

PR #2219, opened 2026-08-24
Stop staleness-only wearable outreach

rocketman-21:
## Why and outcome

A healthy connected wearable can be quiet because the member wears it only occasionally. Elapsed time cannot distinguish that choice from an outage, but two separate paths treated it as authority for proactive troubleshooting.

This removes the elapsed-time notification producer and copy, makes the weekly digest suppress stale-data gaps unless provider state explicitly requires reauthorization, refreshes the exact retired prompt on already-due digest occurrences, and durably retires legacy notification state without another provider or model call.

## Product UX

- Effort: Product change.
- Walkthrough: Ready.
- Entry points: hosted device-sync result apply and the member-owned Monday weekly digest.
- Affected people: members who use a wearable occasionally, members whose provider explicitly requires reauthorization, and members with an already-queued retired notification or digest occurrence.
- Outcome: ambiguous source silence ends with no message; explicit auth failure retains one verified reconnect-link flow; queued legacy work advances without delivery and reaches a terminal outbox state at the ordinary runner checkpoint; an exact old queued digest seed is refreshed in place.
- Exclusions: provider ingestion, operator source-health alerts, UI, consent, and historical source-data repair are unchanged.

## Evidence

- Post-merge Assistant Runtime owner suite: 815 passed, including all 358 workspace-entrypoint cases.
- Final ReviewGPT substantive round 3 passed its full sensitive audit on behavior head d4164a8c027ef78e599f4b569674b9af47f3f8e9 with no findings; the only later commit archives the completed execution plan.
- Restore proof covers pending, retryable, and provider-entry-ambiguous legacy intents through restore → ordinary `assistant_delivery` wake → checkpoint → restore; every case reaches the expected terminal state with zero provider fetches and zero Codex-process preparation.
- Assistant Engine managed automations: 101 passed; hosted Web focused owners: 188 passed; device-syncd: 7 passed; hosted-execution: 2 passed.
- Assistant Engine, Assistant Runtime, device-syncd, hosted-execution, and hosted Web typechecks passed on current `main` plus the candidate.
- `git diff --check` and the pre-commit privacy scan passed.
- Exact retired system-mailbox cases cover absent, pending, retryable, and sending outbox state: the mailbox advances without delivery, preserves the ordinary delivery wake when cleanup is pending, and never enters model or provider delivery.
- The managed-automation test preserves the due schedule, tags, and occurrence context while replacing only the exact retired digest seed; user-authored or unrelated queued payloads remain untouched.
- The production-shaped real-Codex journey asserts zero connect calls and no reply for stale-only data, then exactly one connect call and one returned URL for explicit reauthorization. Its live local call was attempted but could not start because neither supported provider credential is configured; deterministic owner proofs remain green.
- The specialist coverage artifact was inspected and rejected as a duplicate of the smaller identity-mismatch regression restored directly.

- Hosted delivery collection is now a pure suppression fence for the retired namespace; it never mutates durable outbox state while building effects.
- The model-free system-mailbox owner advances an exact retired item without delivery and retains the ordinary `assistant_delivery` wake when nonterminal outbox state exists.
- The existing workspace runner is the sole production caller that terminalizes retired outbox state. It runs only for an incoming or computed delivery wake, after a successful assistant phase, and marks the phase progressed so the ordinary checkpoint persists the mutation.
- Pre-provider outbox states become abandoned; a `sending` state becomes failed as delivery-ambiguous and is never retried.
- Managed automation reconciliation patches only the exact old weekly-digest seed on a due one-shot and preserves its occurrence-owned state.
- Hosted Web result apply, provider-entry revalidation, message templates, and path aliases lose the old producer/materializer/copy surface.
- device-syncd still evaluates and reports push-source staleness for operators; only member-facing authority is removed.
- The independently requested completion-workflow policy is isolated in #2234; this product candidate contains no workflow-policy or CLI contract changes.
- The prior public changelog item is replaced by the shipped outcome in this PR.

## Architecture and reuse

- **Existing systems reused:** Existing source-health observation remains the operational owner; existing account/source auth status remains reconnect authority; existing managed-automation patching preserves queued occurrence state; the system mailbox, delivery collector, workspace runner, and ordinary checkpoint retain their existing responsibilities.
- **New logic:** The collector filters the retired identity without mutation, the exact mailbox path advances without delivery, and the workspace runner reconciles matching outbox state only at the existing assistant-delivery checkpoint boundary.
- **New abstractions:** None. No new queue, durable state, lifecycle enum, migration, background reconciliation pass, service, or dependency was added.
- **Complexity intentionally avoided:** The change deletes the 72-hour episode, materializer, copy catalog, and provider-entry machinery instead of adding cadence inference or preference state. Legacy compatibility is bounded to the shipped identity and one checkpoint-owning reconciliation call site.

## Hot reply path impact

Not applicable. The foreground reply critical path is unchanged. Retired-outbox reconciliation is gated to an incoming or computed `assistant_delivery` wake and runs after the assistant phase; it adds no database, network, provider, or other awaited operation before provider start for a current conversation.

- Murph runtime: not applicable to the first foreground individual or group provider request. No initial system-prompt, tool-schema, or request-assembly path changed. The altered managed seed is provider-visible only on a due member-owned weekly-digest occurrence.
- Codex runtime: not applicable to the first foreground individual or group provider request for the same reason; the seed is occurrence input, not initial Codex request assembly. Group delivery is excluded by the seed's member ownership.
- Measurement method: call-path and complete diff classification plus exact managed-seed readback. This does not claim a measured zero-token delta; a complete initial-request capture was omitted because those initial requests do not contain the changed seed.

## Change-shape breakdown

Classification uses repository path and runtime role; changelog fragments are generated/other, execution plans and compatibility text are docs, and no binary files changed.

| --- | ---: | ---: |
| Source | 246 | 536 |
| Tests/fixtures | 752 | 536 |
| Docs | 154 | 103 |
| Config/tooling | 0 | 3 |
| Generated/other | 14 | 14 |
| **Total** | **1166** | **1192** |

## Risks

- Old warm runner containers can still execute already-queued work until the corrected bundle converges. The deployment order below closes that window before the producer is removed.
- A legacy intent already in `sending` may have crossed provider entry before the corrected runner observes it. The owner marks it delivery-ambiguous and never retries it; the earlier provider outcome cannot be recalled.
- A real silent push outage no longer asks the member to troubleshoot automatically. Operator staleness alerts and explicit provider reauthorization state remain available.

## Deployment concerns

- Deployment: applicable
- Supported skew: the corrected runner consumes old queued mailbox/outbox identities as terminal no-send and refreshes the exact old queued digest seed. Old and new Web versions share the same persisted shapes.
- Safe order: deploy the hosted runner/container bundle first, wait until old warm containers have drained or been replaced, then deploy Web so no new elapsed-time notifications are produced and managed seeds are refreshed.
- Rollback floor: there is no schema migration. Rolling Web back while the corrected runner remains deployed is safe but keeps restored producers suppressed; remove the tombstone last only when intentionally restoring the retired behavior.
- Expected exposure: before runner convergence, an already-queued item could still enter the old send path. After convergence, mixed Web versions can only create items that the corrected runner terminally drains.
- Reversibility: revert Web independently while retaining the tombstone, then revert the runner only if restoring proactive staleness outreach is deliberate.
- Convergence proof: confirm all runner versions contain durable retired-identity terminalization and that old warm containers are gone before Web deploy.
- Post-deploy checks: inspect bounded counts for `assistant.notification.retired_source_delivery_stall_terminal_no_send`, verify no new delivery-stall notification requests appear after Web convergence, and confirm refreshed weekly occurrences contain the stale-data suppression rule.

## Changelog

- Changelog: updated
- Items: 2026-08-24 · quiet-wearables-stay-quiet
- Source PR: #2219

ReviewGPT context sensitivity: sensitive — this changes hosted messaging egress and system-mailbox behavior across Web and runner owners.
ReviewGPT first-reviewed head: 16e109beb9df6a3d0c1875c9f52fc8700b8576a1


rocketman-21:
Superseded by a fresh implementation with one adaptive no-data outreach owner, a member-controlled interval, and transcript-aware follow-up.

PR #2318, opened 2026-08-26
Fix broad private food search timeouts

rocketman-21:
## Outcome

- Bounds private food-name search before ranking so broad catalog matches no longer require a whole-match sort.
- Keeps one bounded GiST admission branch active per query: strict-word-nearest names for full-text searches, or whole-name-nearest names for false-FTS typo recovery.
- Uses the same whole-name trigram metric for typo ordering and eligibility, so closer-but-ineligible names cannot starve valid matches at the 10,000-row cap.
- Keeps exact-name, brand/search-text, generic-only, typo fallback, source filtering, source-priority representative selection, and large-alias canonical diversity within the admitted set.
- Removes the canonical-rank index from the deploy contract because the corrected query has no consumer for it.

## Product UX result

Members should see food searches complete instead of intermittently failing on broad terms, while typo recovery still returns valid foods even when more than 10,000 misleading names rank closer under the full-text relevance metric. Search inputs, result fields, limits, and routes are unchanged; this is a reliability/performance correction with no visual UI change.

## Direct evidence

- Read-only production diagnosis proved the labels database and installed indexes healthy, then reproduced PostgreSQL SQLSTATE 57014 under the real 8-second timeout because the old plan sorted the full FTS match set before LIMIT.
- The accepted review finding was reproduced before the correction: 10,050 names ranked closer under strict-word similarity but failed whole-name trigram eligibility, causing the old shared-candidate typo path to return zero rows instead of 50.
- The corrected focused PostgreSQL regression passes with those same 10,050 misleading candidates, preserves canonical diversity and source-priority representative selection, returns 50 rows, and is deterministic.
- Current SQL on the read-only production catalog completed a broad query in 2.21 seconds and a proven false-FTS typo query in 2.41 seconds under the unchanged 8-second timeout.
- Sanitized production plans completed in 2.35 seconds broad and 1.05 seconds typo. The broad path realized only the 10,000-row strict-word GiST scan; the typo path realized only the 10,000-row whole-name GiST scan; each unused branch had zero loops. GIN FTS was used on both paths.
- Maximum direct sort-child input was 18,688 rows broad and 172 rows typo, below the 20,250 bound. No second typo-similarity base-table scan occurred.
- 34 focused SQL-shape/runtime-env tests pass.
- The focused adversarial PostgreSQL regression passes (1 test, 133 skipped).
- `pnpm --dir apps/web typecheck` passes.
- The all-in-one 250,000-row local plan test reached its 120-second test-harness ceiling on the resource-saturated local executor; it did not produce a query timeout failure. Exact-head CI owns that full regression, while the production-scale executions and plans above passed.

- Both GET and batch POST `/api/foods` share this query owner.
- Supplements and public food search stay on their existing query path.
- Exact ID and UPC lookup paths are unchanged.

## Architecture and reuse

- Existing systems reused: The existing product-label query builder plus current GIN, GiST, and exact-name btree indexes.
- New logic: The existing candidate pipeline conditionally realizes one bounded GiST branch; typo recovery orders and filters with the same whole-name trigram similarity.
- New abstractions: None; the existing product-label query layer remains the sole owner.
- Complexity intentionally avoided: No new service, cache, dependency, state owner, canonical scan, duplicate typo scan, second active nearest-name scan, or timeout increase.

## Hot reply path impact

Food label lookup can sit on the assistant reply path. The change reduces database work and retains the current route and batch concurrency bounds; it adds no network call or transaction.

## Provider-input impact

Food search text remains parameterized and private. No query term, result row, member identifier, or provider payload is added to logs or durable artifacts.

## Deployment concerns

- Deployment: applicable
- Safe order: Deploy Web normally because both required production indexes are already valid, ready, and live.
- Supported skew: Current and new Web builds are both compatible with the two required indexes during rollout; the unused canonical-rank index remains harmless.
- Rollback floor: Reverting Web code remains compatible with all retained production indexes.
- Expected exposure: Web query-shape and preflight changes only; no Cloudflare change, data migration, row mutation, or destructive DDL.
- Reversibility: Revert the Web change and leave the compatible indexes in place.
- Convergence proof: The Vercel production deployment activates the corrected query owner.
- Post-deploy checks: Rerun bounded read-only broad private and generic probes under the 8-second timeout, confirm GiST/GIN admission, and inspect `/api/foods` timeout and 500 logs.

## Changelog

- Changelog: updated
- Items: 2026-08-25 · food-search-finishes-reliably

## Change shape

- Source: +60 / -65
- Tests/fixtures: +407 / -49
- Docs: +125 / -20
- Config/tooling: +0 / -11
- Generated/other: +0 / -0

Reason: Changes the production private-food SQL and required index deployment contract.
ReviewGPT first-reviewed head: 19e238e98e0a94a14fd42a0bca1f25e13cadd3fb


rocketman-21:
Retrospective decision (Round 2)

Original requirement: private food-name search must complete within the existing 8-second database deadline while preserving exact, full-text/brand, generic/source-filtered, typo-recovery, canonical-diversity, and source-priority behavior.

First-reviewed shape: separate 5,000-row FTS, nearest-name, and typo admissions followed by canonical deduplication. That bounded work but let duplicate aliases consume a lane before canonical identity was owned.

Current shape: Round 2 widened only FTS admission to 10,000 rows and added stronger proof. This resolves the established 6,000-alias FTS case but repeats the same admission-before-dedup mechanism on the unchanged 5,000-row typo scan. Review-driven production growth was only +2/-1, but it introduced a second match cap without making the rule uniform.

Decision: redesign inside the existing query owner, not add another lane or exception. Use one 10,000-row match limit for both unordered GIN FTS admission and the single materialized GiST nearest-name admission; delete the independent trigram KNN scan and the second match-cap constant. Both FTS name relevance and false-FTS/true-trigram recovery will reuse that one nearest-name set, then flow through the existing shared ranking and canonical dedupe. Exact-name admission remains 250. No new table, index, service, state, transaction, timeout, or owner.

Required proof: compose the existing 6,000-alias catalog with a query proven FTS-false/trigram-true and assert 50 distinct canonical results, the priority representative, deterministic repetition, bounded sort input (20,250 maximum for the FTS union; 10,000 for typo), and completion under the real 8-second timeout. Re-probe the exact shape read-only against production before the next review.

Why continuation is justified: this removes one candidate scan and one cap concept while giving FTS and typo paths the same bounded name-admission ownership. It is smaller than the current split shape and stays entirely within the existing SQL/test boundary.

PR #2329, opened 2026-08-26
Track silent group-to-private conversions

rocketman-21:
## Outcome

Counts a person once in the existing Ops group-to-private metric when Murph observed them in a Linq group roster before their first private activation, even if they never messaged in the group.

## Product UX

- Affected person: the internal Ops operator interpreting the group-to-private metric.
- Ready state: the definition names both live roster observations and retained group messages, and describes the result as sequence-based exposure attribution rather than engagement or causation.
- No member-facing journey, permission, interaction, layout, loading, empty, error, or recovery state changes.

## Design proof

- Design page: [Ops group-to-private component](https://www.withmurph.ai/design?tab=components#growth-group-private-conversions)
- Evidence: Exact static-render assertions verify the revised explanatory copy in the real production component. The change is text-only, so an image adds no proof.
- Coverage: The design section renders populated and empty states. Both share the revised definition, and no style, interaction, or responsive behavior changed.

## Evidence

- 506 focused deterministic Vitest tests passed across roster reconciliation, attribution, growth snapshots, retention, migration, and privacy schema guards.
- Three opt-in local PostgreSQL tests passed against production SQL: roster observation, production writer-to-attributor composition with idempotent marking, and two-connection reversed-order reconciliation.
- Web typecheck, focused ESLint, Prisma generation, and `git diff --check` passed.
- Review corrected the metric copy and strengthened snapshot and PostgreSQL proof. The temporary synchronous account-deletion path was deleted after the 14-day lifecycle decision. The remaining concurrent-writer finding was fixed with one SQL ordering clause and no new runtime machinery.

- Existing post-response Linq roster reconciliation stores blinded evidence for current non-self handles, including unresolved silent participants.
- Observation rows are ordered by their existing contact-key primary key before upsert, so concurrent rosters acquire shared rows consistently while participant reconciliation stays in the same statement.
- Daily growth snapshot processing joins live evidence to phone or verified-email indexes and the first private activation before setting the existing one-time member marker.
- Hourly retention deletes expired unmatched evidence in bounded serial batches.
- Group-message sender attribution remains a fallback and the existing marker prevents double counting.

## Architecture and reuse

- Existing systems reused: capped Linq roster reconciliation, versioned contact lookup keys, daily growth snapshot, the existing member conversion marker, and hourly retention.
- New logic: one ordered set-based observation upsert, one set-based activation attribution update, and one bounded expiry delete.
- New abstractions: one three-column blinded observation model and one shared 14-day policy constant.
- Complexity intentionally avoided: no synchronous account-deletion cleanup or extra lock coordination. No raw contacts, group/member/route identifiers, relation, provider call, activation-path read, queue, timer, event pipeline, per-group funnel, backfill service, or new dashboard component.

## Hot reply path impact

- The existing roster reconciliation remains post-response and adds no provider, crypto, transaction, or database round trip. It extends the same statement with at most 32 blinded keys. Daily attribution and hourly retention remain off the foreground reply path.

- Individual runtime: Not applicable; no prompt, tool schema, generated guidance, or provider-visible input changed.
- Group runtime: Not applicable; no prompt, tool schema, generated guidance, or provider-visible input changed.

## Risks

- Privacy: evidence stores only a versioned blinded contact key plus first-observed and expiry timestamps. It is global group-derived analytics evidence, not member-owned identity, membership, or access authority; exports omit it and retention removes it no later than 14 days after its latest observation.
- Account recreation can match still-live evidence during the remaining retention window. This bounded behavior is accepted in exchange for avoiding account-deletion coupling.
- Database load: the existing roster path remains one provider read and one SQL statement, with the existing 32-handle cap and one pooled connection. Shared observation rows use their primary key as a canonical acquisition order. Daily attribution is one set statement; retention is serial and capped at four 5,000-row batches per hourly run.
- A contact-key rotation inside the 14-day window can cause a rare false negative; retaining raw or separately keyed identity was intentionally rejected for this internal metric.

## Change shape

Classification rule: production TypeScript and Prisma schema/migration are source; test files are tests/fixtures; architecture, plan, testing map, privacy contract, and Frog entry are docs; no generated or binary files changed.

| --- | ---: | ---: |
| Source | 198 | 8 |
| Tests/fixtures | 617 | 14 |
| Docs | 173 | 2 |
| Config/tooling | 0 | 0 |
| Generated/other | 0 | 0 |
| Total | 988 | 24 |

## Deployment concerns

- Deployment: applicable
- Supported skew: the additive table is safe with old Web code, which ignores it; new Web code requires the table.
- Safe order: apply the Prisma migration before deploying Web.
- Rollback floor: after migration, the prior Web version remains safe; leave the additive table in place during rollback.
- Expected exposure: a code-first deploy would make post-response roster reconciliation fail soft until the table exists, so migration-first ordering is required.
- Reversibility: revert Web to stop writes; retained rows are blinded and bounded by their expiry column.
- Convergence proof: after deployment, one routed group roster pass creates observations and the next daily growth snapshot marks eligible later activations.
- Post-deploy checks: confirm the migration, inspect bounded reconciliation/attribution error aggregates, and verify the Ops total after a synthetic group-to-private journey.

## Changelog

- Changelog: not applicable
- Reason: internal Ops analytics attribution only; no member-visible feature or reliability change.

ReviewGPT context sensitivity: sensitive — adds persisted analytics state and composes provider-roster, activation, growth, and retention owners.
ReviewGPT first-reviewed head: 93b6e090ae6b4298d38c069c323489e169ff4f7d


rocketman-21:

**Original requirement:** Count a silent Linq group participant once when that person later activates privately, without changing group authority or adding a foreground reply-path dependency.

**First-reviewed shape versus current shape:** The first-reviewed head had 188 lines of authored-source churn and already contained the three-column observation table, roster writer, daily attributor, hourly expiry, and existing Ops marker integration. The current head has 205 lines of authored-source churn. Its net runtime growth over that baseline is limited to small corrections around the existing owners; most review-driven additions are boundary tests and accurate documentation.

**Review-driven detour:** Round 1 treated the observation as member-owned deletion data. Round 2 added a synchronous account-deletion helper and focused proofs. Round 2 then proved that helper inverted lock order with roster reconciliation. The current patch deletes the helper, its invocation, deletion count, and deletion tests without replacement locks, retries, or owners. That correction removed 200 lines while adding 24 lines, mostly documentation.

**Owner inventory:**

- Retained: roster reconciliation writes bounded blinded evidence; the daily growth snapshot consumes live evidence into the existing one-time member marker; hourly retention deletes expired evidence.
- Removed: account-deletion SQL, account-deletion ownership, deletion-count plumbing, deletion-specific tests, and the lock coupling they created.
- Unchanged: participant authority, the activation path, the foreground reply path, and the existing Ops metric owner.

**Durable classification:** `docs/hosted-account-data-deletion-export.md` is the authoritative lifecycle contract. The row is global group-derived analytics evidence, not member-owned identity or access authority. Exports omit it. Account deletion does not remove it. A recreated account may match it while it remains live. Its only lifecycle owner is expiry no later than 14 days after the latest observation.

**Decision:** Continue with the current shape. A silent participant has no retained sender message, so the bounded observation is necessary to satisfy the metric. One writer, one consumer, and the existing retention owner are the smallest design that preserves the metric, truthful Ops copy, one-time attribution, bounded database work, privacy classification, and unchanged participant authority. Do not reopen synchronous deletion or lock coordination unless the product requirement changes.

**Proof retained:** Focused deterministic coverage checks SQL shape, attribution, snapshot invocation and failure isolation, retention, migration, and privacy registration. Two PostgreSQL tests prove the production roster writer composes through activation into the production attributor and remains idempotent. The Ops change is text-only with no layout or interaction change, so exact static-render assertions are the proportional UI proof.

PR #2451, opened 2026-08-27
perf(cloudflare): split thin runner boot kernel

rocketman-21:
## Why and outcome

The hosted runner currently evaluates the full workspace executor, drain stack, and deploy-smoke graph before Node can listen. This change keeps a small admission kernel on the synchronous boot path, hydrates the heavy runtime once after listen, and overlaps that hydration with the accepted invocation's existing workspace restore.

Exact production assembly reduces the static boot closure from the recorded 8,571,156-byte / 49-chunk baseline to 1,952,434 bytes across 22 chunks. A fixed byte and chunk-count gate prevents silent eager-import regression. Clean exact preemption now stops the old invocation from waiting on hydration, waits only for its existing workspace restore to settle, and reuses the same warm container.

## Product UX

- Effort: Low
- Walkthrough: Hold
- Affected people: members whose first message reaches a cold hosted runner should spend less time waiting before provider work can begin.
- Exclusions: no UI, copy, interaction, permission, or delivery-contract change. The local exact cold A/B was attempted but cross-architecture execution could not produce a valid baseline sample; rollout remains blocked on production-shaped accepted-to-provider proof.
- Plan difference: none.

## Evidence

- Exact production runner assembly on the final head: entry 67,196 bytes; static boot closure 1,972,340 bytes / 21 chunks; total 11,563,870 bytes.
- Latest Cloudflare node-platform bucket: 151 files passed; 2,719 tests passed and 2 existing skips.
- Runner bundle budget suite: 43 tests passed.
- Assistant-runtime restore/invocation suites: 36 tests passed.
- Cloudflare and assistant-runtime typechecks passed.
- Cloudflare and assistant-runtime builds passed.
- Entrypoint concurrency proof holds the heavy loader unresolved, verifies restore preparation starts, verifies the exact prepared handle reaches invocation, and verifies hydration runs once per process.
- Exact-preemption proof verifies the old request releases without waiting for hydration, retains the single-job slot until restore settlement, does not stop the warm app, and lets the replacement reuse the running container without a destroy or restart.
- Lifecycle remediation proves hydration rejection poisons admission and schedules one terminal exit, while SIGTERM remains bounded even when hydration never settles.
- Local accepted-to-provider A/B: Hold. The historical baseline returned HTTP 500 before its first sample under amd64 translation; the final-head setup also encountered unrelated fixed CLI-generation timeouts. No latency percentage is claimed.

- Fatal and shutdown drains: moved behind the cached heavy-runtime owner; hydration failure now poisons admission and schedules one terminal exit, and shutdown starts its deadline before awaiting hydration.
- Deploy smoke handlers: moved into the same heavy module; existing smoke route behavior remains covered by entrypoint and bundle boot tests.
- Warm workspace path creation: extracted into one shared owner so preparation and invocation resolve the same vault root; restore preparation tests cover the binding.
- Assistant-runtime restore phases: exposed as a one-shot prepared handle bound to the exact request and vault root; mismatch and double-consumption tests fail closed.
- Runner bundle policy: startup bytes and static chunk count are now both gated at production assembly.

## Architecture and reuse

- Existing systems reused: the existing single-job admission owner, runner-job parser, runtime fencing, workspace restore implementation, Cloudflare platform ports, abort signals, and hosted invocation owner.
- New logic: one cached heavy-runtime import and one accepted-job restore preparation run concurrently, then join before the existing invocation continues. On exact preemption, only that invocation stops waiting; the shared hydration continues and the existing restore must settle before handoff.
- New abstractions: a narrow heavy-runtime interface and a one-shot workspace-restore preparation handle; both represent current owners and add no service or durable state.
- Complexity intentionally avoided: no new queue, retry loop, persistence format, orchestration engine, snapshot format, dependency, or second runtime owner.

## Hot reply path impact

Applicable. The change moves existing work earlier and parallelizes it; it adds no database or provider calls.

- Database: zero added or moved calls.
- Network: the existing workspace read remains at most one control-plane request and is moved before heavy hydration completes; an inline workspace still avoids it. Existing snapshot and artifact reads execute once through the unchanged restore owner and retain manifest-driven cardinality.
- Provider: zero added or moved calls; provider start remains downstream of the same restored runtime.
- Other awaits: warm-root filesystem preparation and one workspace restore move earlier; one process-cached heavy import begins after listen. Heavy hydration and restore preparation begin in parallel, then the exact prepared handle is consumed once. Exact preemption stops waiting on hydration but still waits for the old restore to settle.
- Ordering and bounds: the existing commit timeout, abort signal, restore fallback, retry behavior, fencing checks, and single-job slot remain authoritative.
- Before/after proof: exact static boot closure falls from 8,571,156 bytes / 49 chunks to 1,952,434 bytes / 22 chunks. Focused tests reject a cosmetic ready-only split by proving restore starts while hydration is unresolved and the prepared work is reused by the invocation. End-to-end latency remains a rollout gate because the local cross-architecture A/B could not produce a valid baseline sample.

- Murph runtime: Not applicable; no prompt, tool/schema guidance, message assembly, or provider request code changed.
- Codex runtime: Not applicable; no Codex instructions, tool definitions, model configuration, or provider-visible request assembly changed.

## Change shape

Classification uses repository paths: runtime TypeScript is source; test files are tests/fixtures; agent-docs are docs; bundle scripts and package exports are config/tooling; generated output is separate. There are no binary files.

| --- | ---: | ---: |
| Source | 2,051 | 1,555 |
| Tests/fixtures | 1,098 | 159 |
| Docs | 111 | 1 |
| Config/tooling | 42 | 12 |
| Generated/other | 0 | 0 |
| Total | 3,302 | 1,727 |

## Risks

- Earlier health can expose a listening kernel before heavy hydration finishes. The accepted request immediately starts useful restore work, holds the single-job slot, and joins both preparations before invoking the existing runtime; production rollout remains gated on accepted-to-provider-start latency, not the health timestamp alone.
- The restore handle crosses a module boundary. It is one-shot and identity-bound to the exact request object and vault root, and it adopts the canonical runtime abort/fence guard before artifact access continues.
- A clean preemption response preserves the warm container only while the parent has the matching exact abort in flight. Old/new rollout skew and every other error path retain the existing fail-closed replacement behavior.

ReviewGPT context sensitivity: sensitive — this changes the hosted runtime entrypoint, foreground reply timing, shutdown drains, and runner deploy artifact.

ReviewGPT first-reviewed head: 9a185def9fc75112ffaccd68d0bc683f50eed910

ReviewGPT final-reviewed head: fe5a692841e39f6ccc3c661ad7062f5dba9b3692

ReviewGPT final outcome: FINDINGS. Disposition: the reported same-member detached-agent continuation edge case is accepted as a low-incidence product risk; it does not cross a user or security boundary, and the user explicitly chose not to add more lifecycle machinery. The behavior-preserving base merge at 523a3d1f41ad810b1198aef2466c23d1a1020340 and reconciliation commits e6a05032a3744ea289ae8f3c2228d7796586da96 65c0f1e91f7f1b9f7eb85b483a2db136e043bd20, and 1577cf9a8041d6f31e50ebfc2dc7dcbbf0cd077f combine current main&apos;s measured total-output ratchet with the reviewed thin-kernel startup gates, add the missing public-export coverage assertion, and preserve the existing explicit container-entrypoint ownership of the two lightweight Codex lifecycle hooks.

## Changelog

- Changelog: not applicable
- Reason: Internal hosted-runner startup performance and regression protection with no member-visible feature or copy change.

## Deployment concerns

- Deployment: applicable
- Supported skew: The Worker-to-runner job schema and runtime architecture version are unchanged; old and new immutable runner images accept the same request shape.
- Safe order: Build and validate the complete runner bundle/image, publish the immutable image, then roll the compatible Worker/container reference.
- Rollback floor: No state or schema migration exists; revert the Worker/image reference to the prior immutable runner generation.
- Expected exposure: Mixed container generations may coexist during rollout, but each invocation remains isolated inside one generation and uses the unchanged fencing contract.
- Reversibility: Stop rollout and restore the prior image reference; no data rewrite or cleanup is required.
- Convergence proof: Confirm all new containers report the expected release/image identity and the boot-closure metrics match the assembled artifact.
- Post-deploy checks: Run an exact production-cold A/B and verify fence-to-ready, Node startup, accepted-to-provider-start, p95, failure phases, fencing, delivery, and recovery before broad rollout.


rocketman-21:

Original requirement: make a cold hosted runner listen with a small safe admission kernel, begin the accepted invocation's existing workspace restore while one process-cached heavy runtime hydrates, and join both before the unchanged provider/delivery path continues. Earlier health alone is not success.

Shape comparison:

- Base: 71df1789de9d688fe351ea504bca227e4266307a
- Immutable first-reviewed head: 9a185def9fc75112ffaccd68d0bc683f50eed910
- Current head: 9a185def9fc75112ffaccd68d0bc683f50eed910
- Review-driven source growth: zero
- Authored-source churn: 1,950 additions plus 1,543 deletions. The deletions and the 1,023-line heavy module are predominantly relocation of the existing executor, drain, lifecycle, and smoke owners.

Ownership decision: continue as one coherent PR.

- The heavy-runtime dynamic boundary is the mechanism that removes the executor/drain/smoke graph from pre-listen evaluation.
- The one-shot restore-preparation handle is the minimum cross-boundary handoff that lets useful restore work start before hydration finishes without running restore twice or creating a second invocation owner.
- The Cloudflare adapter constructs the existing platform-neutral restore contract without pulling the full Cloudflare invocation factory back into the boot graph.
- The shared warm-root resolver keeps early preparation and the existing invocation fallback on one deterministic path/permission owner; folding it into either caller would introduce a dependency back-edge or duplicate filesystem authority.
- The byte/chunk gate is independent verification, not a runtime owner.

Splitting extraction/budget enforcement from accepted-invocation restore concurrency would temporarily ship the exact misleading state this task rejects: an earlier ready signal with unchanged accepted-to-provider delay. The two parts are therefore indivisible for the requested production outcome.

ReviewGPT found one grouped lifecycle defect in the cached hydration owner: terminal rejection is not yet fail-closed, and the shutdown deadline begins after hydration settles. The correction will tighten that existing owner so rejection poisons/exits once and shutdown races hydration/drain against the existing bound. It will add no retry owner, queue, persisted state, or lifecycle abstraction. The existing hosted-local cold-start benchmark will provide base/head accepted-to-provider and final-delivery evidence before the next substantive review.

PR #2600, opened 2026-08-30
Prevent non-progress runtime wake storms

rocketman-21:
## Outcome

- Keeps default-owned work independently visible while route-filtered device maintenance is active, so a due reminder cannot be hidden by a device-sync wake.
- Advances the durable progress witness only after an applied canonical device import; cold replay, machine-local queue reconstruction, and timestamp-only evidence do not manufacture progress.
- Makes background device maintenance yield between admitted imports and retain remaining generated-image work instead of monopolizing one invocation.
- Replaces the former 64-receipt hard stop with backward-compatible v1 receipt-log compaction: up to 64 active references and 512 cumulative canonical receipts, with deterministic replay and a fail-closed old-writer barrier.
- Bounds hosted invocation mailbox admission at 100 items and preserves current conversations, due default work, provider work, retention work, and canonical future device deadlines while unchanged background work backs off.

## Product UX result

A due recurring reminder remains deliverable exactly once while durable device work remains. The bounded maintenance passes yield between imports, the durable backlog drains to zero across multiple passes, and the scenario completes without failed jobs. A cold replay of unchanged provider data leaves the durable progress generation unchanged.

## Direct evidence

- Canonical receipt-log compatibility and compaction tests passed 15/15, including old-reader ordering, old-writer fail-closed behavior, cumulative restart counts, exact duplicate retry at the 512 boundary, and crash-safe clean-pointer recovery.
- The six changed assistant-runtime suites passed 235/235; hosted execution passed 35/35; device-sync service passed 147/147; importer proof passed 576/576.
- Assistant-runtime, hosted-execution, device-sync, importer, Web, and Cloudflare typechecks passed.
- Focused real-Codex proof passed: unrelated device maintenance did not suppress the independently scheduled mobility reminder, and the actual decision produced one mobility send.
- Closed-loop cold-recovery proof passed: the initial canonical import advances generation once; replay of the same stable WHOOP record emits no duplicate receipt checkpoint and does not advance generation.
- The final public candidate is `ceea6d1e4ec01c6be33438384e42a87e2bad41fb`; 112 focused mailbox/runtime tests and the concurrent receipt-restore regression pass, and `git diff --check`, typecheck, privacy scan, and parent/static audits are green.
- The private Temporal counterpart has passed its 365-test workflow suite, exact replay and patch-order proof, typecheck, and bundle build; its full exact-head verification is running. The next exact paired integration, final ReviewGPT, required CI, and parent final review remain pending.

## Affected surfaces

Hosted workspace checkpoint/read/status wire; canonical-write receipt parsing, compaction, restore, and recovery; hosted mailbox admission; background device-maintenance yielding; generated-image retention; WHOOP and Strava raw-evidence identity; runtime wake projection and handoff; device-sync progress diagnostics; private Temporal retry/admission ownership; and the cross-repository hosted-local scenario.

## Architecture and reuse

- Existing systems reused: HostedWorkspace CAS, canonical-write receipt artifacts, runtime wake ownership, Temporal carry state, structured runtime logs, and the canonical hosted-local harness.
- New logic: one optional progress epoch, one independent default-work witness, a semantic canonical-import progress marker, capped mailbox and receipt admission, and deterministic receipt-log compaction.
- New abstractions: the existing receipt parser is extracted into one shared assistant-runtime module so restore and compaction enforce the same canonical action contract; no new service or state owner is introduced.
- Complexity intentionally avoided: no queue, table, service, scheduler, rate-limit manager, dependency, or duplicated domain state.

The receipt-log wire stays on `murph.hosted-canonical-write-receipt-log.v1`. New readers use an optional cumulative count and active-prefix count. Old readers ignore those fields and replay the padded prefix in first-occurrence order; old writers encounter the pre-existing physical 64-entry boundary and fail before they can erase cumulative provenance. Compaction accepts at most 64 active receipts totaling 4 MiB, so unusually large accepted receipts can still reach the former write-65 failure boundary before the 512 cumulative ceiling; that residual is documented and does not regress the old hard limit.

Legacy omission clears the progress capability fail-open. Equal or +1 CAS transitions make retries idempotent. Attempt, signal, version, wake, and machine-local queue churn do not advance canonical progress.

## Hot reply path impact

Current-conversation and default processing bypass the private background fuse. The 65th pending foreground canonical write can now await at most 64 receipt reads in waves of eight, with at most 4 MiB of compaction input, plus bounded artifact uploads in waves of eight; previously that write failed at the hard 64-entry boundary. Background work normally consolidates at 63 pending receipts. No provider or database wait was added to the reply path.

## Provider-input impact

No initial provider-visible prompt, request, tool schema, or Murph/Codex instruction changed. WHOOP and Strava raw-evidence hashing omits the top-level execution-owned `importedAt` field; normalized import timing and receipt observation time remain intact, while provider records, deletions, source windows, and user health facts are unchanged.

## Deployment concerns

- Deployment: applicable
- Safe order: deploy the private Temporal parser and worker, then the public Web migration and projection, then the public Cloudflare writer and runtime.
- Supported skew: the private parser accepts legacy omission; older public writers leave the capability inactive and fail-open. The v1 receipt wire keeps old readers ordered and old writers fail closed at their existing boundary.
- Receipt-log skew: the wire remains v1. Old readers preserve replay order, while old warm writers fail at their existing 64-entry boundary instead of dropping new cumulative metadata. Current readers and writers continue from the compacted active prefix.
- Reversibility: the public capability writer can remain disabled or be reverted before publication. The private worker becomes fix-forward after a workflow records its non-deprecated patch marker; after capability publication, disable the public writer before reverting compatible public readers.
- Expected exposure: one rollout window may retain legacy unbounded background admission, and an old warm receipt writer may fail at its former boundary; neither path silently discards foreground/default work or receipt provenance.
- Convergence proof: confirm unchanged-start aggregates remain bounded, canonical generations advance only for applied work, compacted receipts restore in order, and a due reminder delivers exactly once during device backlog.

## Changelog

- Changelog: updated
- Items: 2026-08-30 · reminders-stay-on-time-during-syncs

## Change-shape breakdown

Classification rule: `test/**`, `**/*.test.*`, and test helpers under test trees are tests/fixtures; `.agents/friction-log/**`, `agent-docs/**`, and the authored changelog entry are docs; remaining authored runtime, schema, migration, and harness code is source. No binaries or generated files changed.

| --- | ---: | ---: |
| Source | 1,704 | 398 |
| Tests/fixtures | 3,431 | 176 |
| Docs | 369 | 12 |
| Config/tooling | 0 | 0 |
| Generated/other | 0 | 0 |
| **Total** | **5,504** | **586** |

The original requirement is to keep due default-owned work deliverable while bounded device-sync work remains, and to distinguish canonical progress from local replay churn. The first-reviewed shape was source 677 added / 34 deleted, tests 1,947 / 39, and docs 158 / 0. The first-reviewed-head-to-current authored remediation delta, excluding base-only merges, is source 1,049 / 389, tests 1,450 / 207, and docs 226 / 27.

Continue as one indivisible public contract change. Splitting would separate the persisted progress/wake contract, its readers and writers, canonical receipt/recovery semantics, and the cross-repository proof that exercises them together. The growth closes concrete review and integration findings at those existing owners; it adds no queue, service, scheduler, state machine, dependency, or new state owner.

Reason: cross-cutting runtime, persisted-wire, canonical recovery, Temporal, and cross-repository coordination with more than 500 changed lines.

ReviewGPT first-reviewed head: 3a478ba0000afa0d067d00d81aba7737cf6036b0


rocketman-21:
Review anomaly retrospective for substantive round 3: the original requirement is to keep due default-owned work deliverable while bounded device-sync work remains, and to distinguish canonical progress from local replay churn. The first-reviewed public shape was source 677 added / 34 deleted, tests 1,947 / 39, and docs 158 / 0. The current shape is source 921 / 92, tests 2,368 / 70, and docs 278 / 0; the direct first-head-to-current remediation delta is source 268 / 82, tests 437 / 47, and docs 120 / 0. The growth is attributable to keeping filtered and global wakes distinct from one read, applying progress only at the canonical receipt boundary, removing execution timestamps from WHOOP/Strava evidence identity, and exact cross-repository regression proof. Decision: continue this PR as one indivisible public contract change. Splitting would separate the persisted triple, its readers/writers, canonical progress semantics, and the proof that exercises them together. The change still reuses the existing workspace CAS, receipt, runtime, and harness owners; it adds no queue, service, scheduler, state machine, dependency, or new state owner. Further expansion beyond these existing owners is not justified.

rocketman-21:
The latest non-starvation E2E failure is a test-ordering deadlock, not evidence that the projected-default-wake fix missed the production path. The due reminder reaches the provider at the deadline. The test then holds that response and waits for the authoritative dirty-resource count to fall from 113 to 13 before releasing it, but that reduction is a deferred post-checkpoint dirty acknowledgement and cannot run until the assistant turn yields. The 90-second stream idle timeout therefore expires first; only a later invocation applies the acknowledgement, and retries find the one-shot scripted response consumed.

Minimal correction: delete the pre-release `expectPendingDirtyResourceCount(seed.connectionId, firstPassRemainingCount)` wait. Do not replace it with `pendingJobCountAfter` or `pendingQueuedJobCountAfter`; both are correctly zero because the remaining resources are staged dirty state, not queued jobs. Existing assertions still prove 113 seeded, first pass processed 100 and hit its cap, exactly one reminder provider request and no early send, one final send, eventual authoritative count zero, pass counts `[100, 13]`, exactly two passes, and no job failures. I reproduced both queue fields locally and confirmed the circular wait is the only invalid assertion.

PR #2615, opened 2026-08-30
feat(cloudflare): add one-slot warm standby

rocketman-21:
## Outcome

Adds an opt-in, one-slot, release-scoped ENAM Cloudflare runner standby. The slot boots the image, hydrates the heavy runtime, and completes a disposable content-free Codex App Server initialize/stop preflight before any member claim. The real reusable Codex process still starts after immutable member binding and encrypted-workspace restore.

The existing exact-version deployment smoke now also proves the effective standby mode. A boolean capability receipt prevents an older public smoke implementation from silently passing the paired private workflow.

## Product UX result

- Effort: Patch.
- Walkthrough: Ready.
- Affected people: an eligible cold member, a concurrent second member who must fall back without delay, and a member withdrawing consent or deleting data after a claim.
- Material exclusions: no rendered UI changes, no multi-slot or multi-region pool, and no claimed production latency gain before shadow evidence.
- Difference from the approved plan: none.
- An eligible cold member can claim the ready shell and skip the Cloudflare container/image/runtime portion of cold start.
- A concurrent second member, a slot undergoing readiness reproof, or any no-ready/error case before ownership uses the existing exact-user cold path.
- Readiness reproof atomically withdraws the slot from claims before external work starts and republishes the same slot only after proof succeeds.
- A rare ambiguous bind after the opaque stop target is durably reserved retries that exact target rather than risking two live member containers.
- A foreground standby timeout releases the existing consent mutation lock while retaining the exact cleanup target; withdrawal and account deletion then retire that target and fail closed on member mismatch or unsettled cleanup.

## Direct evidence

- `pnpm --dir apps/cloudflare typecheck` — passed on current head `6dce1a13d7883cff161e49da60d8d4d1245bcc0e`.
- `pnpm --dir apps/cloudflare build` — passed on the current head.
- `pnpm --dir packages/hosted-local-harness typecheck` — passed on the current head.
- Current-head hosted-local parity proof — Wrangler fidelity 4/4 and harness environment 94/94 tests passed.
- Reviewed production-candidate Cloudflare suites — 4 files, 259/259 tests passed across two stable shards; this includes all 14 coordinator continuation tests plus entrypoint health, binding and cleanup, and alarm paths.
- Prior exact-head directly affected suites — 8 files, 751/751 tests passed in two stable shards (201/201 and 550/550).
- Composed foreground path — claim/bind, invocation-binding timeout, `retry_later`, consent-lock release, exact-target retention, and queued withdrawal retirement passed in one production-owner journey.
- Readiness ownership race — initial readiness, atomic ready-to-provisioning withdrawal, concurrent `no_ready_slot`, same-slot republication, and exactly two preparations passed.
- Existing provider/internal-host interception, destructive cleanup, exact-mode smoke, and paired deployment-contract suites remained covered by the directly affected and prior focused exact-head proof.
- `git diff --check`, changed-file identifier scan, and secret-pattern scan — passed.
- Independent deadline, coordinator-race, and simplification reviews — passed with no remaining production finding. ReviewGPT round 7 returned `ROUND_OUTCOME: PASS` on exact production candidate `27e9ce3e3b4764c5489a7b301254f702a6b7d7bb`.
- Post-review CI remediation mirrors the already-reviewed production container, bindings, and migration into the hosted-local harness and exports those existing classes from its test entrypoint. It changes no production runtime or deployment config. Per explicit author direction, there is no round 8; parent review and exact-head CI own this local-only delta before merge.

- Container health/readiness and the content-free Codex shell probe: necessary to advertise only a fully hydrated pristine slot; covered by standby and health tests.
- Runner selection, command budgeting, write-fence retention, provider-egress interception, wake routing, consent-lock release, and destructive cleanup: necessary to keep member authority and the exact opaque stop target with `UserRunner`; covered by runtime, egress, and composed user-control tests.
- Worker scheduled bootstrap, two SQLite Durable Object classes, two bindings, Container placement, and migration `v7`: necessary to own one release-scoped slot without product truth or a generalized scheduler; covered by coordinator and deploy-artifact tests.
- Hosted-local Wrangler generation and its test Worker entrypoint: necessary to keep local runtime declarations aligned with production while the mode remains default-off; covered by exact config-fidelity, export, and harness tests.
- Rendered deploy artifacts and the paired private deployment workflow contract: necessary to fail closed on unsupported public/private skew; covered by exact-mode smoke and deploy-contract tests.
- Architecture, security, reliability, runtime protocol, activation, rollback, and cost documentation: necessary to keep the experiment and deletion gate durable and reviewable.

## Architecture and reuse

- Existing systems reused: `UserRunner` member authority, the foreground command budget, consent mutation lock, write fences, encrypted-workspace restore, provider-egress minting and host interception, Container lifecycle, Durable Object alarms, `waitUntil`, same-member idle retention, account deletion, consent withdrawal, and the existing exact-version deploy smoke.
- New logic: one release-scoped memberless coordinator keeps one opaque pristine slot; `UserRunner` durably reserves the exact stop target, binds it once, then uses the existing member runner path. Public health reports the canonical effective mode for the existing smoke to verify.
- Hosted-local parity reuses those exact class names, bindings, and `v7` migration in the existing config builder and re-exports the same concrete classes; there is no second config source, test double, state owner, or local mode enablement.
- Readiness has one mutation owner: coordinator maintenance atomically moves ready to the existing provisioning field before reproof. The container activity hook only renews platform lifetime, and only a successful claim schedules immediate replenishment.
- Stale/off cleanup has one durable continuation owner: after reconciliation and owned-slot retirement, a scalar existence read over the existing tombstone table keeps the same coordinator alarm alive until every claim resolves. Alarm-persistence failure is returned to Cloudflare for platform retry.
- Deadline ownership has one source: fresh claim/bind and immediate ambiguity recovery share `min(foreground deadline, now + 250 ms)`; retained recovery and invocation binding use the original remaining foreground budget with the existing Web-control step cap.
- New abstractions: only narrow standby coordinator/slot contracts needed to keep coordination memberless and enforce immutable slot binding; the concrete standby class reuses the existing egress registration owner and host map.
- Complexity intentionally avoided: no generalized scheduler, autoscaler, reusable cross-member process, second health probe, second readiness owner, per-RPC deadline reset, recovery queue, duplicate egress map, tombstone scheduler, durable field, or asynchronous member-binding layer; unused retirement timestamps, response echoes, and superseded branches are deleted.
- The round-3 owner/deadline simplification was net-deleting in production source (107 additions and 121 deletions across four runtime owners). The round-5 fix adds only one scalar existence read and one condition on the existing alarm, while deleting the swallowed scheduling fallback. The current complexity collapse projects two health values once from the already-frozen startup environment and extracts the existing fresh allocation sequence as one owner-local controller phase; behavior, ordering, deadlines, state, and authority are unchanged.

## Complexity impact

- Guard: pass — the current-main `pnpm complexity:diff` implementation passes against the PR base and reviewed production head.
- Post-review local parity correction: 22 source additions and 2 deletions across the existing hosted-local config/entrypoint owners, with no production hotspot or runtime path changed.
- Hotspots: The `createServer` request handler (`<anonymous>`, 125) and unchanged `runner-container.ts` owners `wakeRuntime` (74), `invokeHostedExecution` (21), `ensureContainerReady` (36), `destroyIfRunning` (26), and `classifyHostedRunnerContainerErrorResponse` (24) remain at base complexity. `runtime-invocation.ts` improves from debt 10 to 9 and maximum 26 to 25, with `prepareWithFence` (24) and `prepareWorkspaceRunnerInvocation` (25) remaining; `runtime-processing-controller.ts` improves from debt 12 to 9 with maximum 27 unchanged, with `ensureExistingRuntimeProcessing` (27) and `startRuntimeProcessing` (22) remaining. No changed-file hotspot worsens.
- Agent judgment: Two startup values are projected once from the existing frozen environment and the existing claim/reserve/bind sequence is extracted into one owner-local phase; no behavior, state, protocol, owner, or module was added, and a broader module split was rejected as unnecessary.

## Hot reply path impact

The source-controlled default is `off`, so the shipped foreground reply path is unchanged. In `shadow`, provisioning and replenishment remain off-path. In `allocate`, selection adds at most one coordinator claim RPC, one local durable stop-target reservation, and one slot bind RPC in serial under a single fresh-standby budget capped by both 250 ms and the existing foreground command deadline. Immediate ambiguous-bind resolution remains inside that same budget. Retained-target recovery and the claimed invocation's exact binding proof use only the original remaining foreground budget, each with the normal Web-control step cap; no operation restarts or extends the caller deadline. A pre-ownership miss falls back to the exact-user path, while uncertain post-reservation ownership yields for safe retry. No-ready claims do not enqueue redundant maintenance. No retry loop or replacement provisioning is added to the accepted-message path. Deterministic shared-deadline, expired-no-dispatch, late-bind, atomic reproof, and composed consent-lock tests are the direct proof; production latency remains an activation gate to measure in shadow rather than a claimed result.

- Murph: Not applicable — no instructions, tools, schemas, generated guidance, model selection, or Murph provider-request assembly changed.
- Codex: Not applicable — the content-free preclaim probe makes no provider request, and a claimed turn reuses the existing Codex/provider request assembly only after exact binding. A binding timeout occurs before provider credential mint or invocation. No provider-visible input surface changed, so token/byte comparison is not applicable.
- Credential boundary: provider and internal pseudo-host traffic from the concrete standby class uses the existing Worker-owned host interception map; this changes authorization routing, not provider-visible prompt content.

## Deployment concerns

- Deployment: applicable
- Supported skew: The public Worker contract may land first while standby mode remains off; the private workflow fails closed until that contract and its live-mode smoke receipt are available. Worker/container release or bundle mismatch fails closed through readiness fingerprints, while already bound older member containers continue the ordinary exact-member lifecycle and are never returned to ready.
- Safe order: Merge the public Worker PR first and the private workflow PR second, then deploy the additive bindings and migration with mode absent or off before progressing through shadow and allocate.
- Rollback floor: Set mode to off immediately; retain the class exports and bindings until unbound slots retire and already bound member containers drain. Old-release/off coordinators retain their own alarm only while an owned slot or claim tombstone remains, then stop after convergence.
- Expected exposure: Off changes no member path; shadow maintains one unclaimed ENAM slot; allocate exposes only eligible cold starts to the bounded claim attempt.
- Reversibility: Returning to off prevents new claims without interrupting existing exact-member containers, which continue through the ordinary lifecycle.
- Convergence proof: The pinned public contract, generated Wrangler value, exact-version health mode, capability receipt, release-scoped coordinator state, and runner smoke must all agree before activation.
- Post-deploy checks: In shadow, verify exactly one current-release ENAM slot remains pristine with zero claims; check cleanup, latency, ready coverage, reliability, and cost gates before allocate.

One continuously ready 2-vCPU/6-GiB/6-GB slot is documented at about $40.52 per 730-hour month for gross allocated memory and disk before unused shared included allotment, plus actual active CPU.

## Changelog

- Changelog: not applicable
- Reason: The source-controlled mode remains off, so this PR does not change member-visible behavior until a separately controlled operational experiment is activated.

## Change shape

Classification rule: runtime files under `apps/cloudflare/src/**` are source; `apps/cloudflare/test/**` are tests/fixtures; Markdown is docs; deploy scripts and Wrangler configuration are config/tooling; generated code is separate. Binary files: none.

| --- | ---: | ---: |
| Source | 1,866 | 48 |
| Tests/fixtures | 1,995 | 1 |
| Docs | 310 | 5 |
| Config/tooling | 119 | 3 |
| Generated/other | 0 | 0 |
| **Total** | **4,290** | **57** |

Reason: This changes runtime trust boundaries, destructive cleanup, durable coordination, and Cloudflare deployment surfaces.

ReviewGPT first-reviewed head: 1dc69d0f57259e09481e547c06cf273e62d605da


rocketman-21:
## Round 3 architecture retrospective

### Irreducible requirement

Keep one opt-in, release-scoped ENAM standby slot; keep it memberless and content-free until one immutable bind; persist the exact opaque target before that bind; retain that target through ambiguous ownership; fail closed for withdrawal and deletion; and ship/deploy with mode `off` before `shadow` or `allocate`.

### First-reviewed versus current shape

The immutable first-reviewed patch (`1dc69d0`) already contained all three durable ownership boundaries: the memberless coordinator, the physical standby Container/slot store, and the existing per-member `UserRunner`. It was 3,132 additions / 52 deletions across 32 files, with 1,880 lines of authored-source churn.

The current patch (`5eb2635`) is 3,596 additions / 55 deletions across 41 files, with 1,878 lines of authored-source churn. The review delta is 488 additions / 27 deletions across 17 files: production source 22/20, tests 398/0, deploy docs 16/4, and deploy-smoke tooling 52/3. Review therefore reduced authored-source churn by two lines; the growth is almost entirely direct proof and the paired deploy contract, not new runtime ownership.

### Remediation attribution

- Preliminary coverage finding: added layered proof in the existing withdrawal and account-deletion owner suites that exact claimed-slot retirement precedes acknowledgement and that unsettled retirement retains the target. No production mechanism was added.
- Rounds 1 and 2, exact-target loss: replaced caller-selected preservation flags with one state-store rule derived from the stored standby target. The obsolete options and two per-caller branches were deleted; transport-failure, replacement, restart, withdrawal, and deletion owners have focused proof.
- Rounds 1 and 2, provisioning wedge: changed the existing persisted-target no-op into resumption through the same prepare/reconcile/alarm path. This added no state, owner, queue, scheduler, lease, or recovery service.
- Round 2, concrete-class egress bypass: reused the existing registration helper and host map for `StandbyRunnerContainer`; no second map or interception owner was added. Static registration and production-shaped provider/internal-host routing are both proved.
- Public effective-mode reporting and the smoke capability receipt were proactive paired-deployment remediation, not a public ReviewGPT finding. They reuse the canonical mode parser, existing `/health` response, existing exact-version smoke, and GitHub step output so the private workflow cannot silently pass against an older public smoke implementation. No second probe or state owner was added.

### Collapse alternatives tested

1. Make one deterministic Container both coordinator and claimed member slot: rejected. After immutable bind it must remain member-owned and never reusable, so it cannot advertise/provision the successor. A generation or successor pointer simply recreates a separate durable coordinator.
2. Move shared allocation into `UserRunner`: rejected. Independent per-member Durable Objects cannot atomically admit one winner for one shared slot.
3. Move member binding/cleanup into the coordinator: rejected. That gives the memberless availability owner member identity and makes cleanup depend on a remote global owner; it also violates persist-target-before-bind ordering.
4. Replace the coordinator with native Container lookup: rejected. The Container namespace provides named instances, not a cross-`UserRunner` atomic ready pointer or orphan-claim ledger. The coordinator's one SQLite transaction is the minimal shared serialization point.
5. Remove the separately threaded standby namespace in favor of only the generic router: rejected after code-path inspection. The controller must prove the standby binding exists before consuming a global claim; the generic router intentionally exposes only base container RPCs and throws lazily for a missing standby binding. Unifying them would require a new availability signal/optional RPC abstraction and make the pre-claim fallback less explicit.
6. Split the deployment receipt from this PR: rejected. The private workflow consumes it as the fail-closed compatibility proof for the paired rollout, and duplicating the live-mode probe privately would create a second parser/probe owner.

### Decision

Continue the current three-owner shape: coordinator = global atomic availability only; standby Container = physical readiness, immutable binding, and terminal retirement; existing `UserRunner` = member authority, write fence, invocation, and user-control cleanup. The contract module remains the dependency seam that lets those owners share identities without importing concrete implementations.

Shrink two unused data surfaces before the next review: delete the unread `retired_at` slot column/value, because terminal `state = 'retired'` is the invariant; and remove unused echoed claim/release/region fields from `HostedStandbyClaimResult`, keeping only outcome plus `slotName` on success. Do not add a new abstraction for either change.

No other review-created mechanism is superseded. The added proofs cover different owners or boundaries rather than duplicating one test path. Later-round metadata will carry this explicit continuation-and-shrink decision without resetting the first-reviewed baseline.

PR #2637, opened 2026-08-31
Durably retire deleted hosted runtimes

rocketman-21:
## Outcome

Makes Temporal workflow termination a durable account-deletion cleanup target. A deletion receipt remains pending until every captured runtime workflow is terminated or confirmed absent, and deadline-limited retries resume from durable progress instead of replaying the same prefix forever.

## Product UX

Account deletion keeps its current immediate response and recovery behavior. Temporary Temporal unavailability no longer lets cleanup appear complete; the existing encrypted receipt retries without blocking the member-facing transaction.

## Direct evidence

- Cleanup, termination, retention-owner, and production-migration suites: 116 focused tests passed on the current candidate.
- The maximum-cardinality test converges all 1,024 captured runtime IDs across bounded attempts without reissuing a confirmed prefix; the gap test proves later concurrent success cannot skip an earlier failure.
- Local PostgreSQL rolling-deploy guard proof: 8 tests passed, including existing-row and old-Web defaults, contract failure on null, final non-null hardening, and both deletion guards.
- Web typecheck, focused ESLint, Prisma validation, and diff check passed.
- Preliminary and final round one found the lost-progress flaw, and final round two resolved it. Required CI then caught the predeploy required-column incompatibility; the current candidate moves that hardening to the existing postdeploy contract lane.

## Architecture and reuse

- Existing systems reused: The encrypted deletion-cleanup receipt, immutable captured runtime-member order, shared attempt deadline, lease-token update, existing cleanup worker, and exact-deployment contract-migration lane.
- New state: One `temporalNextRuntimeIndex` integer on that receipt, defaulting to zero for legacy rows and old-Web inserts.
- New logic: Deterministic batches of four advance only through the contiguous terminated-or-absent prefix. Progress is persisted by the existing receipt update and becomes immediately eligible for the next ordinary attempt; attempts with no progress retain exponential backoff.
- Migration phases: Predeploy adds the cursor nullable with default zero; the exact-deployment postdrain contract lane fails on any null before setting `NOT NULL`.
- New abstractions: None. The existing cleanup receipt, termination helper, and contract-migration owner already own the retry lifecycle, bounded external call, and schema hardening.
- Complexity intentionally avoided: No bitmap, queue, payload rewrite, service, dependency, shared client, transaction-spanning provider call, or duplicate state owner.

The existing KMS-encrypted, foreign-key-free cleanup receipt remains the single retry owner. Temporal calls remain outside database transactions. A crash before the cursor write can only repeat idempotent termination; the lease token prevents a stale owner from regressing durable progress.

- The additive predeploy migration creates a nullable default-zero cursor and a delete guard; the exact-deployment postdrain contract rejects null before requiring the cursor.
- The cursor default makes legacy receipts begin at index zero; a rolling old Web process may repeat work but cannot erase retry ownership.
- One absolute timeout spans Temporal connect plus terminate.
- A failed item stops cursor advancement even when later calls in its batch succeeded; those later calls are safely retried.
- Late signal recreation is contained by the paired private signal-only quiescence patch; termination alone is not treated as a permanent fence.
- Existing runtime-log cleanup guard and the Temporal guard coexist.

## Risks

- Maximum admitted cardinality is 1,024 immutable runtime IDs. Peak Temporal concurrency remains four, with one existing receipt read and one lease-fenced progress update per pending attempt.
- No provider or network call runs inside a database transaction.
- Receipt deletion remains fail-closed until the final cursor reaches the immutable ID count and the completion timestamp is stored.

## Hot reply path impact

None. No webhook, mailbox append, runtime reply, or current-inbound path changes.

## Provider-input impact

None for model or messaging providers. The only external call is idempotent Temporal workflow termination during account-deletion cleanup.

## Deployment concerns

- Deployment: applicable
- Supported skew: The private guard is safe with old Web; predeploy default zero keeps existing receipts and old-Web inserts valid, and postdrain `NOT NULL` still accepts old-Web omissions. New Web must wait for the private guard.
- Safe order: Promote and prove the private Temporal worker first; then let Vercel apply the nullable public expansion, promote the cursor-aware Web build, drain prior functions, and apply the postdeploy `NOT NULL` contract.
- Rollback floor: Keep the private quiescence guard and additive database migration in place if Web rolls back.
- Expected exposure: Existing deleted-runtime loops continue only until the private worker reaches their live frontier; the migration-to-Web window retains incomplete cleanup receipts.
- Reversibility: Roll back Web without dropping the nullable completion column, cursor, or deletion guard; fix forward the replay-patched private worker.
- Convergence proof: Prove the reviewed private Build ID is Temporal Current, then confirm bounded canonical-absent reconciliation and runtime-start aggregates fall while cleanup receipts complete.
- Post-deploy checks: Verify private pollers and binding admission, database migration state, pending cleanup counts, and aggregate reconciliation and runtime errors without selecting identifiers.

## Changelog

- Changelog: not applicable
- Reason: Internal privacy and reliability correction with no new member-visible surface.

## Change-shape breakdown

Classification rule: Runtime and Prisma implementation paths are source; test paths are tests and fixtures; agent and product documentation paths are docs; migration and schema support paths are config and tooling. No binary files changed.

| --- | ---: | ---: |
| Source | 169 | 17 |
| Tests/fixtures | 607 | 23 |
| Docs | 163 | 11 |
| Config/tooling | 46 | 0 |
| Generated/other | 0 | 0 |
| **Total** | **985** | **51** |

## Round 3 retrospective

- Original requirement: Stop deleted canonical accounts from retaining self-retrying Temporal runtimes, and make termination a durable part of the existing account-deletion cleanup without blocking deletion.
- Shape comparison: The first-reviewed head changed 16 files and 623 lines with authored source +144/-15 (159 churn). The current head changes 17 files and 1,036 lines with authored source +169/-17 (186 churn), far below the 2,000-line threshold.
- Growth attribution: The review-driven source growth is the one receipt-owned cursor needed to prevent prefix replay. Round 2 to current changes no runtime source; its 190 changed lines and one added file are primarily the CI-required migration phase split plus direct static and PostgreSQL proof. No underlying mechanism repeated.
- Concepts and owners: The existing encrypted receipt remains the sole retry owner; one integer is its progress. The existing termination helper owns the bounded external call, the private workflow owns late-signal quiescence, and the existing exact-deployment contract runner owns postdrain NOT NULL hardening. There is no new queue, service, state machine, reconciliation loop, bitmap, backfill process, or second state owner.
- Decision: Continue. The cursor is the minimum durable state that closes the proven 1,024-runtime liveness failure, while the nullable expansion plus fail-closed contract is the repository-required deployment shape for that same integer. Splitting either from this PR would reintroduce the failure or an unsafe rollout window.
- Future bound: Freeze runtime architecture at these owners. Further changes are limited to concrete findings on this exact behavior or proof; do not add another lifecycle, owner, compatibility layer, repair pass, or generalized cleanup abstraction.

Reason: Cross-repository privacy and reliability behavior spanning Temporal replay, durable retry progress, and a database migration.

ReviewGPT first-reviewed head: 6963ec81e6f3cceef8f159ddad63ca8f5e8192fa


rocketman-21:
Exact-head CI retry evidence: Release app verification job 99404011805 remained in progress for more than 79 minutes, versus a recent successful p90 of 22m17s and maximum of 22m30s. Live output stopped at 06:49:40Z mid-test; two later instances of the same job passed in roughly 21–22 minutes, and the stalled test/provider source are unchanged from this PR’s prior successful head. Cancelling workflow run 33365086441 and rerunning this exact-head job once; no code or PR-head change.

PR #2679, opened 2026-09-01
Fix stale default wake owner livelock

rocketman-21:
## Why and outcome

A stale durable `default` wake could keep winning Temporal arbitration after live runtime state had already disproved that wake. When the next durable item belonged to a model-free system owner, the default pass handed it off with `progressed: false`; the runner therefore skipped the projection checkpoint, leaving the stale default wake durable and recreating the same pass indefinitely.

This change checkpoints that corrected runtime projection without claiming application progress. The next arbitration selects the actual system owner, while genuine default work keeps its existing priority.

## Product UX

- Effort: Patch.
- Walkthrough: Ready.
- Affected people: members whose background device or maintenance work is blocked behind a stale default timer now converge automatically.
- Material exclusions: fresh conversation input, genuinely due assistant/reminder work, and ordinary foreground replies retain their existing priority and behavior.

## Evidence

- Production incident evidence: the exact alert cohort stayed imported but unhandled while Temporal repeatedly selected `default`; SDK replay of both alert-window and current histories passed, ruling out workflow nondeterminism.
- Production-shaped local replay: a transient, identifier-redacted copy of the exact failing envelope preserved its relative 0/4/4 watermarks, expired-position gap, three live item kinds/order, stale assistant projection, running canonical cron state, and model-free device frontier. It passed on this head with one projection-only checkpoint, unchanged generation/handled state, and an immediate device-owner recheck. No production row or identifier was persisted.
- Untouched `origin/main` red proof: the committed composed regression failed at `checkpointRequests.length` with `0 !== 1`; the assistant phase returned the device-owner wake but emitted no projection checkpoint.
- Head two-cycle proof: the same normalized 0/4/4 fixture restores checkpoint 1's real snapshot in `system_mailbox` mode, consumes live sequence 1, advances handled-through to 2 across the expired sequence gap, preserves live sequences 3/4 byte-for-byte, and keeps generation unchanged. Assistant/Codex preparation remains unused; one empty device snapshot read performs no external wearable-provider execution.
- `pnpm exec vitest run packages/assistant-runtime/test/hosted-runtime-workspace-assistant-phase-foreground.test.ts --no-coverage`: 85 passed.
- `pnpm exec vitest run packages/assistant-runtime/test/hosted-runtime-workspace-runner.test.ts --no-coverage`: 122 passed.
- `pnpm exec vitest run packages/assistant-runtime/test/hosted-runtime-workspace-entrypoint-system-mailbox.test.ts --no-coverage`: 51 passed.
[...]

PR #2680, opened 2026-09-01
Split standby container capacity

rocketman-21:
## Why and outcome

The protected Cloudflare deploy is blocked because one capacity value renders both the ordinary and standby container applications, so adding standby repeats the 748-instance ceiling and exceeds account quota. This adds one optional standby-specific ceiling with backward-compatible inheritance, allowing production to render 648 ordinary, 1 deploy-smoke, and 100 standby instances.

## Product UX

- Effort: Patch
- Result: Ready
- Walkthrough: This is an internal capacity and rollout change. Members see no pool label, new message, allocation, or authority change. Standby remains `off` for the first deploy; `allocate` is excluded.

## Evidence

- Direct: `pnpm exec vitest run --config apps/cloudflare/vitest.node.workspace.ts --project cloudflare-node-deploy --no-coverage apps/cloudflare/test/deploy-automation.test.ts apps/cloudflare/test/container-image-contract.test.ts apps/cloudflare/test/deploy-artifacts.test.ts` passed 85 tests; the rendered config assertion proves 648 normal, 1 smoke, and 100 standby.
- Direct: `pnpm exec vitest run --config apps/cloudflare/vitest.node.workspace.ts --no-coverage apps/cloudflare/test/container-rollout-config.test.ts` passed 3 tests.
- Direct: `pnpm --dir apps/cloudflare typecheck` passed after refreshing the lockfile-pinned dependency install.
- Coverage: Explicit split, omitted-value inheritance, invalid-value rejection, scaffold parity, rollout steps, and prepared-artifact consistency are covered.
- Contract: `CF_STANDBY_CONTAINER_MAX_INSTANCES` is exported through a deploy-only public contract and explicitly excluded from Worker runtime vars, allowing the protected private guard to fail closed on an unsupported public head.

## Non-obvious affected surfaces

- The protected `murph-cloud` workflow must import and forward the deploy-only `CF_STANDBY_CONTAINER_MAX_INSTANCES` contract; its paired patch is required before production variables change.
- Checked-in and hosted-local defaults remain unchanged because an omitted standby value inherits the ordinary ceiling.

## Architecture and reuse

- Existing systems reused: Existing deploy-environment parser, generated Wrangler config owner, Cloudflare container primitive, protected deploy workflow, and current standby coordinator.
- New logic: One positive-integer deploy input with fallback to the already-resolved ordinary ceiling, plus one deploy-only contract export for the protected workflow guard.
[...]

PR #2681, opened 2026-09-01
Fit standby containers within production capacity

rocketman-21:
## Why and outcome

Production deploys declared the full active-runner maximum on both the active and standby container classes. At the current two-vCPU shape that asked Cloudflare for 2,994 vCPU of maximum container capacity, above the account ceiling, so standby application creation failed and prevented the incident fix from shipping.

This change gives the standby declaration an independent capacity input that defaults to one. Current off/shadow deployment fits exactly within the existing ceiling, while any future `allocate` activation fails preflight unless operators first raise and retain standby capacity for the complete bound-target lifecycle.

## Product UX

- Not applicable: this is internal deployment-capacity configuration. It does not change a member journey, UI, copy, or product policy; the member-visible runtime recovery remains in PR #2679.

## Evidence

- Production failure: the only deploy that reached standby application creation rendered active 748, smoke 1, and standby 748 at two vCPU each; Cloudflare rejected the new standby application for exceeding account capacity.
- Current-production safety: standby application count and standby running/allocated instance count are both zero. The deployed mode is `off`, all standby deploy attempts failed, and no runtime/control evidence shows an allocated or retained standby target.
- Base red proof: off/shadow rendered standby `748` with gradual rollout steps instead of `1` with an immediate single step.
- Head arithmetic regression: `748×2 + 1×2 + 1×2 = 1,500` vCPU. Raising standby to 748 produces 2,994 and `allocate` is rejected unless standby capacity is at least active capacity.
- `pnpm test:diff:local`: exit 0 on the stable diff — 152 Node files (2,785 passed, 2 skipped), 6 container-helper tests, and 15 Workers tests.
- Focused independent review: 45/45 tests passed; Cloudflare typecheck, `pnpm complexity:diff`, `git diff --check`, and privacy/secret scan passed.

- Private deploy workflow and GitHub Environments: the new non-secret input requires an explicit fallback mapping and both preview/production values set to `1`; PR #2681 is not deployment-complete without that private half and live parity proof.
- Future standby allocation: `allocate` must raise standby capacity to at least the active maximum before activation and retain it through off/shadow rollback until every bound target retires. Covered by deploy preflight tests and durable reliability/deploy guidance.
- Checked-in Wrangler scaffold: standby defaults to one instance and one immediate rollout step while class, ENAM constraint, binding, image, and migration remain unchanged. Covered by the container contract and rendered-config regressions.

## Architecture and reuse

- Existing systems reused: the existing deploy environment parser, Wrangler renderer, standby mode contract, Durable Object class/binding/migration, and Cloudflare container declaration remain the only owners.
- New logic: one deploy-only positive-integer input independently sizes `StandbyRunnerContainer`; existing preflight rejects `allocate` when it is below active capacity.
- New abstractions: no service, state owner, queue, lifecycle enum, binding, migration, or dependency was added; the new field stays inside the existing deployment environment object.
- Complexity intentionally avoided: capacity is not derived from mode, so rollback needs no target scan or reconciliation machinery; operators retain the raised value until existing owners prove target retirement.

## Complexity impact

- Guard: pass — `pnpm complexity:diff`.
- Hotspots: no changed function exceeds the tracked threshold; complexity debt remains zero and no maximum increased.
- Agent judgment: no further behavior-preserving simplification is justified; a separate input is the smallest shape that preserves both today's one-slot deployment and future bound-target rollback safety.

## Hot reply path impact

- Not applicable: deploy-time Wrangler rendering changes no runtime request path and adds no database, network, provider, or awaited operation to the Foreground Reply Critical Path.

- Murph runtime: Not applicable; no prompt, tool/schema guidance, model request assembly, or provider-visible input changed.
- Codex runtime: Not applicable; no prompt, tool/schema guidance, model request assembly, or provider-visible input changed.

## Change-shape breakdown

Classification is by path and role: deploy automation logic is source, Vitest files are tests/fixtures, durable guidance and the active execution plan are docs, and the checked-in Wrangler scaffold is config/tooling. No generated or binary files changed.

| --- | ---: | ---: |
| Source | 25 | 6 |
| Tests/fixtures | 92 | 2 |
| Docs | 112 | 9 |
| Config/tooling | 2 | 2 |
| Generated/other | 0 | 0 |
| **Total** | **231** | **19** |

## Deployment concerns

- Deployment: applicable
- Supported skew: old/new Worker and runner bundles use the same Durable Object classes, bindings, migrations, and runtime vars; this input affects deploy-time container capacity only.
- Safe order: land the public capacity and runtime fixes, land the private workflow fallback, set preview and production Environment values to `1`, then run the protected production workflow with immediate container rollout.
- Rollback floor: current production has no standby application or bound target, so cap `1` is safe. After any future allocation, retain the raised cap through mode rollback until all bound targets are proven retired.
- Expected exposure: before the coordinated deploy, production continues running the pre-fix Worker and the affected runtime remains stalled; a partial Worker activation without successful container creation is not completion.
- Reversibility: current off-mode capacity can remain at `1`; reverting this public change would only reintroduce the deploy quota failure. Future post-allocation reduction is forbidden until target retirement is proven.
- Convergence proof: the rendered artifact and live Cloudflare declarations must report active `748`, smoke `1`, standby `1`, with the intended Worker version at 100%.
- Post-deploy checks: verify Worker fingerprint/bindings, list and inspect all three container applications, run the hosted smoke, inspect bounded error aggregates, and confirm the exact stalled cohort advances.

## Changelog

- Changelog: not applicable
- Reason: internal deploy-capacity plumbing only; PR #2679 owns the member-visible recovery outcome and changelog entry.

## Risks

- Lowering capacity after a future allocated standby target exists could reject starts. The deploy contract therefore separates capacity from mode and forbids reduction until bound-target retirement is proven.
- The public default is not treated as complete private wiring; workflow mapping, Environment values, and live declaration parity remain explicit deployment gates.

ReviewGPT context sensitivity: sensitive — production Cloudflare container capacity and rollback contract change.

ReviewGPT first-reviewed head: 4eb3786f9d002c3835b9ee8701bae7cdc6ee89ef

### DIFF
--- a/agent-docs/RELIABILITY.md
+++ b/agent-docs/RELIABILITY.md
@@ -373,7 +373,13 @@ Last verified: 2026-08-30
   uses one 250 ms claim/bind deadline. A miss before slot ownership uses the
   ordinary exact-user fallback; an ambiguous bind after the per-member stop
   target is durably reserved retries that exact target instead of risking two
-  live containers.
+  live containers. The standby class, binding, and migration remain deployed in
+  every mode. Its independent platform-capacity input defaults to one for the
+  never-allocated deployment. Before `allocate`, operators raise it to at least
+  the active-runner maximum and retain that raised value through an `off` or
+  `shadow` rollback; they reduce it only after proving every bound standby target
+  has retired. Mode changes never implicitly narrow capacity beneath persisted
+  same-member targets.
   A coordinator transaction admits at most one winner, then replacement
   provisioning runs under `waitUntil`; alarms re-prove readiness, retry failed
   retirement, expire unbound claim tombstones, and drain stale releases. The
--- a/agent-docs/exec-plans/active/2026-09-01-standby-container-capacity.md
+++ b/agent-docs/exec-plans/active/2026-09-01-standby-container-capacity.md
@@ -1,0 +1,64 @@
+# Standby Container Capacity
+
+Status: active
+Created: 2026-09-01
+Updated: 2026-09-01
+
+## Goal
+
+Fit the current never-allocated standby declaration within Cloudflare's
+production container ceiling without removing the standby class, binding, or
+migration, and preserve enough independent standby capacity for any future
+allocation and rollback lifecycle.
+
+## Proven cause
+
+- Production declares 748 active two-vCPU runners and one two-vCPU deploy-smoke
+  runner. Reusing the active maximum for standby declared another 748 runners,
+  taking composed capacity from the 1,500-vCPU account ceiling to 2,994 vCPU.
+- The renderer derived standby capacity from the active maximum even while
+  standby mode was `off` or `shadow`; the platform rejected the composed
+  declaration before the Worker could deploy.
+- Runtime mode is not a safe capacity owner after allocation because persisted
+  bound standby targets can survive a later mode change.
+
+## Contract
+
+- `CF_STANDBY_CONTAINER_MAX_INSTANCES` independently owns standby platform
+  capacity and defaults to `1`. Runtime mode never implicitly changes it.
+- Current `off` or `shadow` production may deploy the one-instance cap only
+  after read-only evidence proves standby allocation has never produced a bound
+  target.
+- Before future `allocate`, raise standby capacity to at least
+  `CF_CONTAINER_MAX_INSTANCES` while still in `shadow` and prove the live value.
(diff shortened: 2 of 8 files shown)


rocketman-21:
Closing as superseded by #2680. The independent standby-capacity contract is already on main with the current 648 active / 100 standby design; this older branch's one-instance default and standby-at-least-active allocation guard are no longer applicable.

PR #2702, opened 2026-09-01
fix(runtime): drain reported metrics in system mailbox mode

rocketman-21:
## Outcome

Classify imported reported-daily-metric work under the deterministic system-mailbox owner so a later metric cannot keep publishing a default assistant wake behind earlier device-sync work. Preserve full active-share refreshes for those metrics while keeping grant-maintenance projection first-materialization-only.

## Product UX

Patch. Delayed connected-health work resumes without a hot no-progress loop; current conversations, reminders, deliveries, and already-materialized shared health views keep their existing behavior. Recovery stays automatic and silent.

## Direct evidence

- Red ownership proof on the unchanged base: the mixed device-plus-metric frontier published a due assistant default wake, the shared classifier returned `default_owned`, and the system-mode entrypoint left the metric pending.
- Red ReviewGPT regression on the first candidate: a reported metric incorrectly called projection with `first-materialization`.
- Green runtime proof on the rebased head: 85 focused runtime and contract tests pass; the entrypoint drains the metric, calls projection without a mode, clears the item, and avoids the assistant lane, while maintenance still calls `first-materialization`.
- Deterministic regression proof: the real in-process entrypoint imports a causal 8,000-step observation, exposes an already-materialized `steps-days.v0` grant only to full refresh, delivers the manual record exactly once, advances the handled frontier, and leaves no default wake. Temporarily restoring the buggy all-projection `first-materialization` selection makes this exact test fail red; all 52 system-mailbox entrypoint tests pass with the fix.
- Hosted Web reconciliation: 58 tests pass and project the metric as `model_free`.
- Hosted execution, assistant runtime, and hosted Web typechecks pass; the Cloudflare runner package build also passes.
- Changelog validation: 39 tests pass.
- Cyclomatic complexity diff passes with debt reduced by one and no maximum increase.
- The production-shaped hosted-local journey creates an active grant and reads back the prior device value plus the manual correction exactly once with no personal reply. A local run with the protected Temporal worker reached the test but its emulated runner container was killed with exit 137 during initial activation, before these assertions; no production invariant was weakened to bypass that host-resource failure.

## Affected surfaces

- Shared hosted mailbox execution-class projection.
- Cloudflare hosted runtime model-free route admission and vault-share projection-mode selection.
- Hosted Web reconciliation frontier classification.
- Public changelog item.
- No schema, wire, Temporal workflow, provider-input, or persisted-state changes.

## Architecture and reuse

- Existing systems reused: the shared mailbox classifier, system-mode route allowlist, deterministic event handler, and existing vault-share projection modes.
- New logic: one event kind and one route action join their existing allowlists; a small predicate selects first-materialization only for projection records created by maintenance wakes.
- New abstractions: one pure local projection-mode resolver; no new state owner, service, queue, retry path, dependency, or compatibility shim.
- Complexity intentionally avoided: no alternate scheduler, mailbox scan, special-case retry, duplicated projection path, or persisted migration.

## Complexity impact

- Guard: pass; `pnpm complexity:diff` reports complexity debt reduced by one and no maximum increase.
- Hotspots: the existing runtime entrypoint remains at 253; the local helper removes projection-mode selection from that hotspot and introduces no new hotspot.
- Agent judgment: the helper is the smallest explicit shape that preserves separate maintenance and reported-metric projection semantics without duplicating the projection path.
- Source: +15/-3
- Tests and fixtures: +165/-11
- Docs and changelog: +53/-4
- Config and tooling: +0/-0
- Generated and other: +0/-0

## Hot reply path

Foreground input priority is unchanged. The corrected item has already completed its canonical import and runs only its deterministic event and projection-record boundary.

## Provider-input impact

None. Provider payloads, parsing, and request behavior are unchanged.

## Deployment concerns

- Deployment: applicable
- Supported skew: Cloudflare must accept the new system owner before hosted Web begins projecting it; schema, state, and Temporal protocol remain compatible throughout the bounded rollout.
- Safe order: deploy the exact candidate to Cloudflare with immediate container replacement and certify its receipt, then allow the hosted Web production deployment for the merged head.
- Rollback floor: rolling either consumer back below this change reintroduces the no-progress loop for this mailbox shape.
- Expected exposure: any old warm container or old Web reconciliation build can retain the previous owner until both surfaces converge.
- Reversibility: no migration or state rewrite; code rollback is mechanically safe but restores the incident.
- Convergence proof: confirm `system_mailbox` execution, device-sync pass progress, handled-frontier advancement, default-wake clearance, and termination of repeated no-progress invocations.
- Post-deploy checks: verify the exact Cloudflare and hosted Web release receipts, then recheck bounded runtime logs and control state for frontier progress with no repeated no-progress attempts.

## Changelog

- Changelog: updated
- Items: 2026-09-01 · background-work-recovers-from-stale-timers
- Presentation: https://www.withmurph.ai/screenshots/ops#changelog-archive

Sensitivity reason: hosted runtime ordering, projection refresh semantics, and cross-surface deployment skew change.
ReviewGPT first-reviewed head: b66fe81a1ed74d84055cf3e3e81ded673a3cd8c5

### DIFF
--- a/agent-docs/exec-plans/active/2026-09-01-reported-daily-metric-model-free.md
+++ b/agent-docs/exec-plans/active/2026-09-01-reported-daily-metric-model-free.md
@@ -1,0 +1,35 @@
+# Reported daily metric mailbox ownership
+
+Status: active
+Created: 2026-09-01
+Updated: 2026-09-01
+
+## Goal
+
+- Stop a later reported daily metric from publishing a default-processing wake while earlier model-free device-sync work owns the durable system-mailbox frontier.
+
+## Product UX
+
+- Effort: Patch.
+- Outcome: delayed connected-health work resumes without a hot no-progress loop.
+- Reaches: members whose runtime imports device-sync work followed by a reported daily metric.
+- Proof: a production-shaped mixed-mailbox regression plus post-deploy system-mode, device-pass, and frontier-advance evidence.
+
+## Root-cause evidence
+
+- `health.daily-metric.reported` is currently classified as `default_owned` even though import already performs its canonical write and execution is a deterministic no-op plus projection record.
+- In a mixed frontier, that later row publishes a due default-processing wake while the earlier device-sync row remains the model-free execution frontier.
+- Default invocations hand the earlier device row back to system mode, then checkpoint the unchanged due default wake, producing repeated no-progress invocations.
+
+## Plan
+
+1. Add a failing contract regression for the reported metric execution class and a failing mixed-frontier regression matching production ordering.
+2. Move the deterministic reported-metric item into the existing model-free ownership set and allow its existing import route in system-mailbox mode.
+3. Add runtime-entrypoint proof that system mode consumes the item without entering the assistant lane.
+4. Run focused tests, package typechecks, complexity, changelog validation, diff/privacy review, exact-head CI, and ReviewGPT.
+5. Merge, deploy with immediate hosted-container rollout, and verify the affected production frontier advances and the loop stops.
+
+## Deployment concerns
+
+- Cloudflare hosted runtime first, with immediate container replacement; then deploy Web only if the merged shared classifier is consumed by its reconciliation build.
+- No schema, wire, Temporal workflow, provider-input, or persisted-state migration is expected.
--- a/agent-docs/exec-plans/completed/2026-09-01-reported-daily-metric-model-free.md
+++ b/agent-docs/exec-plans/completed/2026-09-01-reported-daily-metric-model-free.md
@@ -1,6 +1,6 @@
 # Reported daily metric mailbox ownership
 
-Status: active
+Status: completed
 Created: 2026-09-01
 Updated: 2026-09-01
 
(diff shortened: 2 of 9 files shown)


rocketman-21:
## Round-3 anomaly retrospective

Original requirement: deterministically own `health.daily-metric.reported` work in the system-mailbox path without an assistant wake, preserve an ordinary full refresh for already-active shares, and keep `first-materialization` restricted to grant maintenance.

The first-reviewed implementation added the event classification and route admission with a lightweight entrypoint assertion. The current implementation retains those two ownership changes and adds one local pure predicate at the existing vault-share projection call site. That predicate derives the projection mode from two existing facts: the wake is `runtime.maintenance-requested` and the record is `vault-share.projection`. It introduces no state, service, queue, scheduler, lifecycle, compatibility path, or new runtime owner.

The guarded PR grew from 207 to 317 changed lines across the same eight files. Round 2 accounts for the production growth: ReviewGPT correctly found that applying `first-materialization` to every projection would skip stale already-materialized shares, so the predicate and focused maintenance-versus-metric proof were added. Round 3 accounts only for test growth: the user requested a regression that exercises the canonical import and real system-mailbox entrypoint and cannot pass on a control-flow-only assertion. The ancestry-restoring merge added unrelated cumulative repository changes, but it did not change the eight-file PR implementation.

Concept inventory: production adds one pure local mode resolver and reuses the existing wake kind, event classifier, route action, projection modes, and projection owner. It removes the incorrect implicit concept that every projection is a first materialization. Tests replace the earlier lightweight daily-metric case with persisted canonical import, an already-materialized scope visible only to full refresh, one data-bearing delivery, frontier/drain assertions, and no-default-wake/no-assistant assertions. No owner or durable concept was added by the test.

Decision: explicitly justified continuation. Deleting the predicate would collapse two required semantics back into the reviewed bug; splitting or adding machinery would make the ownership less direct. The current source shape is the smallest explicit derivation at the existing authority boundary.

The expanded in-process regression is also retained as the minimum durable automatic proof. `packages/assistant-runtime/test/hosted-runtime-workspace-entrypoint-system-mailbox.test.ts` runs in the assistant-runtime package suite selected by required CI and owns deterministic canonical import, full-refresh delivery, persisted frontier/drain, and wake/assistant exclusion. The registered `group-sleep-source-sharing` hosted-local E2E is a production-shaped selectable companion rather than an ordinary required-CI gate; it owns full-stack active-grant behavior, the prior device value plus the manual correction exactly once, and no personal reply. Keeping both avoids relying on a resource-heavy external runner for the required deterministic boundary. The observed external-runner exit 137 does not justify weakening either production behavior or the regression.

No additional production machinery or further code change is warranted.

rocketman-21:
Retroactive ReviewGPT round 3 resolved: **PASS**, with no qualifying findings. It explicitly passed determinism, non-fakeability at the changed boundary, required-CI ownership, and sufficiency beside the registered production-shaped E2E companion. Reviewed exact test head: `adcabef900b284161b953fce72f471f37ab3ed08`.

PR #2738, opened 2026-09-02
Fix hosted completion recovery and standby handoffs

rocketman-21:
## Why and outcome

A completed hosted runtime could leave its durable write fence behind when the disposable `RunnerContainer` activation restarted before receiving the result. The next foreground message then repaired that fence, but standby admission rejected every `replaced` action and forced another cold start.

This change makes the container process converge the existing `UserRunner` fence only after releasing all local busy state, admits authenticated foreground replacements to the existing standby pool, and gives allocate mode one shell-prewarm owner. No queue, alarm, poller, persisted promise, new state owner, or dependency is introduced.

## Product UX

- Effort: Patch.
- Walkthrough: Ready.
- Affected people: members who message Murph while background work owns their hosted runner, or immediately after a disposable container activation restarts.
- Result: completed prior work no longer leaves a stale fence, and the current foreground reply may take the ready standby instead of waiting on an exact-user cold start. Background, retention, system-mailbox, non-default, Temporal, and untrusted requests remain excluded from fresh standby claims.

## Direct evidence

- The release-order regression blocks the completion receipt, proves `/health` reports `activeJobCount: 0`, starts a second real HTTP invocation before releasing the first receipt, and proves it receives 200 rather than 409.
- Sender contract tests prove the exact URL, method, headers, and body plus `recorded`/`not_recorded`, unavailable-route, non-2xx, transport-failure, and non-settling one-second timeout behavior.
- Runner-control tests prove valid and stale CAS results, 401 missing authority, 400 malformed input, 413 oversized input, and no extra authority round trip.
- Focused Cloudflare ownership, invocation, standby, alarm, and routing proof passed: 607 tests. The exact foreground controller cases include trusted replacement admission and background/untrusted exclusion.
- Focused hosted-local harness proof passed: 207 tests, including the exact CI billing/stack lane. The exhaustive Journal wake contract passed: 1 test.
- Cloudflare and hosted-local-harness typechecks passed. `pnpm complexity:diff` passed with flat debt, including `runner-outbound.ts`, outbound diagnostics, and the hosted-local stack. `git diff --check` passed.
- Exact-head full-stack proof passed: reset-before-result passed 1/1; foreground standby/preemption passed 9/9 primary cases plus 2/2 checkpoint cases.

- The container entrypoint captures only a successful `{ job, result }`, releases wake/abort/activity/transport ownership, then sends one bounded internal receipt. The old invocation-level and activation-level senders are deleted.
- The extracted runner-control handler bounds and parses the result, requires attempt/generation authority, and invokes the existing atomic UserRunner completion method directly.
- Allocate-mode exact-user prewarming is skipped so the standby coordinator is the only prewarm owner; pending and retained standby targets still reconcile before a fresh claim.
- Hosted-local provisions, inspects, and cleans actual standby containers. The reset E2E loses the activation-owned outer result rather than mocking completion convergence.
- Journal group-fact mailbox parsing remains included because the exhaustive foreground journey exercises every registered system wake.

## Architecture and reuse

- Existing systems reused: `UserRunner` remains the sole durable write-fence authority; the existing bound outbound transport, attempt/generation headers, standby coordinator, and ordinary outer completion path remain authoritative.
- New logic: one post-release container completion sender, one extracted bounded runner-control handler, foreground-intent admission independent of `started` versus `replaced`, and allocate-mode prewarm ownership selection.
- New abstractions: none; the existing bound outbound transport, runner-control route, UserRunner CAS, and standby coordinator remain sufficient.
- Complexity intentionally avoided: no recovery queue, poller, replay manager, scheduler, compatibility state, duplicate fence owner, database table, or new dependency.

## Complexity impact

- Guard: pass — `pnpm complexity:diff` completed with no increased debt.
- Hotspots: `runner-outbound.ts` remains at its base debt and diagnostics decreases maximum complexity by one; the new sender and route are below threshold. Hosted-local stack debt is flat.
- Agent judgment: the remaining modules match existing transport and test seams; collapsing them would put parsing and failure handling back into the large dispatcher.

## Hot reply path impact

The standby change adds no database, provider, or new network operation. A trusted foreground replacement may use the same bounded coordinator claim/bind sequence already used for a trusted fresh start, with the existing exact-user path on a claim miss.

After successful runtime work, the process releases its local busy state and then awaits one internal receipt for at most one second. The receipt has both `AbortSignal.timeout` and an independent `Promise.race`, no retry, and cannot change the completed result. This ordering prevents durable release from racing a container-side 409.

Not applicable for either runtime: no assembled instructions, tool/schema guidance, prompt text, or other provider-visible input changed.

## Design proof

- Design page: https://www.withmurph.ai/screenshots/ops#changelog-archive
- Evidence: only the existing changelog item's source-PR provenance changed; its member-visible copy, production archive component, states, and interactions are unchanged.
- Coverage: the production archive component server-render contract is unchanged.

## Risks

- Fence correctness: only UserRunner's existing atomic runtime-kind user + attempt + generation compare-and-swap can clear the fence; duplicate, late, wrong-user, or wrong-generation receipts do not record.
- Trust/privacy: the internal route requires fence authority, binds the routed user, limits JSON to 256 KiB, and never logs result bodies.
- Recovery: any receipt failure logs `not_recorded` and preserves the completed result; UserRunner retains precise durable failure attribution.
- Rollout: the old process lacks the process-origin sender, so the first deployment must use immediate container rollout after the new Worker route is live.

## Change-shape breakdown

Classification rule: production runtime files are source; `test/**`, `*.test.*`, hosted-local test routes/subclasses, and test Durable Objects are tests/fixtures; Markdown, changelog content, and Frog are docs; hosted-local harness runtime/stack files are config/tooling. Generated output is separate. No binary files changed.

| --- | ---: | ---: |
| Source | 347 | 102 |
| Tests/fixtures | 855 | 164 |
| Docs | 300 | 55 |
| Config/tooling | 66 | 0 |
| Generated/other | 0 | 0 |
| Total | 1,568 | 321 |

ReviewGPT context sensitivity: sensitive — the patch changes a cross-owner runtime protocol, failure recovery, concurrency admission, and Cloudflare deployment behavior.

ReviewGPT first-reviewed head: bcc08ca9458f5b58fee5bbf6b09306b2d414b135

## Deployment concerns

- Deployment: applicable.
- Supported skew: the Worker route is available before the new image. A new process against an older Worker preserves its result after a bounded `not_recorded`; an old process against the new Worker retains only the ordinary outer path and therefore lacks reset convergence during that short window.
- Safe order: use the protected Cloudflare workflow with `container_rollout=immediate`; Vercel requires no tandem deploy.
- Rollback floor: no persisted shape or external API changes. Roll back the full Cloudflare release together; a sender/route mismatch remains fail-open to the completed result for at most one second.
- Expected exposure: immediate rollout may interrupt active containers, which retain the existing checkpoint-safe replacement behavior.
- Convergence proof: protected smoke must report the reviewed Worker release and candidate runner-bundle fingerprint before the canary.
- Post-deploy checks: verify allocate mode, capacity 100, one current-release ready/running standby, an actual `claimed` trusted foreground allocation, clean replies, `recorded` completion receipts, and no new stale-fence repair for the canary.

## Changelog

- Changelog: updated
- Items: 2026-08-30 · replies-wake-sooner-after-idle-time.


rocketman-21:
### Retroactive ReviewGPT closure

Reviewed head: `bcc08ca9458f5b58fee5bbf6b09306b2d414b135`

Two independent exact-head full-patch audits converged on one qualifying finding: **Purpose Drift**. The Journal system-mailbox action/parser additions and their fixture were packaged with the foreground-recovery change even though that lifecycle is adjacent scope.

Disposition:

- The packaging finding is accepted.
- The suggested post-merge deletion is rejected as unsafe. The `journal.group-fact.recorded` producer from `4994d15fcb3893269e7d217bde6e23565fcf997c` predates this PR's base. Removing its consumer now would restore `system_mailbox.unsupported_route` retries and degrade an existing consent-gated private Journal flow.
- Separately owned PR #2747 already carries the focused feature-specific proof. No edits were made to its branch or worktree.
- The independent Mountain round reported no additional qualifying defect. Its completion-warning and best-effort admission-read observations were explicitly non-blocking and did not justify speculative changes.

The safe retroactive remediation is therefore scope/provenance closure, not a production rollback. The merged recovery commit `2f1fc8abb2b961ab7ca4811c65348d970300daeb` was deployed through protected run [33701001804](https://github.com/cobuildwithus/murph-cloud/actions/runs/33701001804), and the guarded three-turn production canary passed in run [33702453259](https://github.com/cobuildwithus/murph/actions/runs/33702453259).

PR #2779, opened 2026-09-03
Fix retained device-sync recovery wake dedupe

rocketman-21:
## Why and outcome
Scheduled device-sync recovery could return 500 after mailbox privacy retention removed payload metadata from an already-imported wake. Rebuilding the stable v3 event produced the same logical wake, but generic exact dedupe compared it with retired null metadata and classified it as a conflict.

This keeps generic mailbox dedupe strict. The scheduled-wake producer recognizes a retired duplicate only when the existing runtime pending-state owner checkpoints that exact mailbox sequence as its first pending system item, the imported watermark covers it, and the canonical lane frontier agrees. The accepted duplicate creates no new mailbox identity, signal, or Temporal handoff. Missing, stale, ambiguous, never-imported, already-handled, or structurally different state fails closed.

## Product UX
- Effort: Product UX Patch
- Walkthrough: Ready
- People: Existing members whose connected-health schedule remains pending across mailbox content retention.
- Exclusions: New wakes, live duplicates, never-imported rows, skipped rows, legacy sequence-less ambiguity, already-handled rows, and malformed or structurally different rows keep the existing fail-closed behavior.
- Plan difference: Review exposed that aggregate watermarks could not prove per-item ownership, so the final design uses the runtime's exact pending item rather than adding more watermark inference.
- Rendered evidence: Not applicable to this background recovery path. The isolated changelog entry is covered by changelog model, page, and route tests.

## Evidence
- PostgreSQL retention proof: 15 passing cases, including exact ownership, a skipped retired row, and the retention-filtered fast-forward with legacy sequence-less pending work.
- Runtime-owner proof: 70 passing mailbox state, import, and checkpoint tests; legacy ambiguity publishes no exact pending sequence.
- Surrounding Web proof: 279 passing device-sync wake, mailbox, due-reconcile, combined recovery, route, and due-connection tests.
- Changelog proof: 59 passing fragment, model, page, and route tests.
- CI fixture follow-up: the five exact runtime status assertions now include the new nullable field (145 focused tests pass), and the current-base OpenAPI example expectation includes its three new defaults (4 tests pass).
- Hosted Web and assistant-runtime typechecks, focused Web ESLint, complexity guard, diff check, and identifier privacy scan passed.

ReviewGPT first-reviewed head: 6dc4d786f4697b894ae90d0c9875ed9b3c2cb455
ReviewGPT final reviewed head: b63d2421e321336d2d3640a04af2bc28bcc20a07
Reason: The patch changes persisted-state idempotency, hosted execution retry ownership, and fail-closed recovery behavior.

## ReviewGPT dispositions
- Round 1 accepted: lane-wide imported and handled watermarks did not prove that every sequence between them was runtime-owned. The first correction narrowed the range to the contiguous handled frontier.
- Round 2 retrospective completed: supported legacy sequence-less work showed that even the contiguous aggregate frontier remained ambiguous. The requirement was restated as exact pending-item ownership and the retrospective was recorded before redesign.
- Final correction: the existing runtime pending-state owner now checkpoints its exact first pending sequenced item; legacy sequence-less work publishes `null`. Web requires an exact sequence match in addition to the existing imported-watermark, frontier, high-water, structure, and privacy checks.
- Complexity disposition: no new state owner, schema, migration, receipt set, queue, scheduler, lifecycle, or repair path.

- Hosted mailbox append internals: producer-specific classification runs only after ordinary exact comparison reports a conflict.
- Assistant runtime checkpoints: the existing redacted status now includes the exact first pending system sequence, or `null` when it cannot be proven.
- Scheduled wake completion: an accepted runtime-owned retired duplicate suppresses the redundant `reconcile_due` signal; inserts and live exact duplicates retain current behavior.
- PostgreSQL load: exact-conflict scheduled retries add one bounded `SELECT EXISTS` inside the existing per-connection transaction. The sweeper remains capped at 25 candidates with five concurrent connection-scoped transactions.

## Architecture and reuse
- Existing systems reused: stable scheduled-wake v3 identity, mailbox retention tombstone, runtime pending-system-mailbox state, workspace checkpoint, imported watermark, handled frontier, and lane counter.
- New logic: the runtime derives one exact pending sequence from state it already owns; Web atomically compares that fact with the retired row.
- New abstractions: no new service or state abstraction; one small progress result keeps the two checkpoint facts derived from a single state read.
- Complexity intentionally avoided: no private payload restoration, generic dedupe relaxation, new durable model, cross-package internal import, or second retry owner.

## Complexity impact
- Guard: pass — `pnpm complexity:diff` completed successfully.
- Hotspots: changed-file debt and maximum complexity are unchanged; no new function exceeds 20.
- Agent judgment: the final design removes unsupported aggregate ownership inference and transports one fact through an existing checkpoint.

## Hot reply path impact
- Not applicable: only scheduled background connected-health recovery changes. Current conversation acceptance and reply handoff are untouched.

- Murph runtime: Not applicable; no prompt, tool schema, generated guidance, or provider-visible request changed.
- Codex runtime: Not applicable; no prompt, tool schema, generated guidance, or provider-visible request changed.

## Change-shape breakdown
Classification rule: authored production TypeScript is source; Vitest files are tests/fixtures; protocol, execution plans, and changelog are docs; no generated output is included.

| --- | ---: | ---: |
| Source | 283 | 38 |
| Tests/fixtures | 644 | 3 |
| Docs | 249 | 0 |
| Config/tooling | 0 | 0 |
| Generated/other | 0 | 0 |
| Total | 1176 | 41 |

## Risks
- Privacy: retired ciphertext, references, byte counts, and hashes remain deleted and are never reconstructed.
- Reliability: absent, mismatched, malformed, overflowing, skipped, already-handled, beyond-high-water, sidecar-bearing, or structurally different state remains a visible conflict.
- Compatibility: older runtime checkpoints omit the exact pending fact, so the Web side fails closed until a new runtime checkpoint is present.

## Deployment concerns
- Deployment: applicable
- Supported skew: older runtime checkpoints may omit the new exact-owner field; Web treats omission as unproven and fails closed without duplicate work.
- Safe order: deploy assistant runtime and Cloudflare first, allow active runtimes to checkpoint the new field, then deploy Web.
- Rollback floor: Web may roll back independently; do not keep the new Web classifier deployed while intentionally rolling all runtimes below the field-producing version.
- Expected exposure: deploying Web first or before a runtime checkpoints can temporarily preserve the existing recovery 500 for that connection.
- Reversibility: the added checkpoint field is optional redacted metadata, so either deployment can be rolled back without data migration.
- Convergence proof: a new runtime checkpoint publishes the exact first pending sequence or null, after which Web evaluates the tuple atomically with canonical lane facts.
- Post-deploy checks: verify checkpoints contain the field and due-reconcile sweeps stop reporting retention-only conflicts for provably pending wakes.

## Changelog
- Changelog: updated
- Items: 2026-09-03 · connected-health-scheduled-recovery-stays-reliable


rocketman-21:
## ReviewGPT round 2 retrospective

- Original requirement: acknowledge a content-retired scheduled wake without another signal only when canonical facts prove that exact mailbox item remains pending in the runtime; otherwise fail closed.
- Shape comparison: the first-reviewed and current heads both contain 253 lines of authored production-source churn. Round 2 added no production concept and only replaced one aggregate-watermark predicate, so review-driven source growth is zero. The repeated finding shows that the aggregate inference itself is the wrong proof.
- Supported-state decision: legacy unsequenced pending system work remains supported. Retired-wake reuse is outside the safe acceptance set while that state makes exact ownership ambiguous; missing or ambiguous proof must fail closed.
- Existing owner: runtime system-mailbox pending state already knows the exact earliest pending sequenced item. The redesign will project that one sequence through the existing redacted checkpoint owner. Web will require the retired row sequence to equal that exact runtime-owned pending sequence, alongside the existing imported/high-water, structural, and privacy checks.
- Complexity decision: replace the inferred `consumed + 1` ownership proof with one derived checkpoint fact. Add no schema, migration, per-item receipt set, queue, lifecycle, retry owner, repair pass, or reconciliation loop.
- Validation decision: add a composed production-path regression for a legacy unsequenced blocker, retention projection fast-forward over retired sequence 1, live sequence 2 import/checkpoint publication, and scheduled retry rejection; retain the safe first-pending acceptance proof.
- Continuation decision: redesign this PR on that existing-owner derivation, rerun focused proof and exact-head CI, then submit substantive ReviewGPT round 3.

